PR #115, opened 2012-04-02
Symlink mojito when installing from GitHub

chrismytton:
Rather than install the full mojito package globally, just symlink the
cloned copy into the global npm path using `npm ln`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ on-line/off-line, multi-device, hosted application platform.
 
     $ git clone git://github.com/yahoo/mojito.git
     $ cd mojito/source
-    $ npm install -g .
-    $ npm install .
+    $ npm install
+    $ npm ln
 
 ### via npm
 


mojit0:
This is a reasonable short-term workaround provided you don't mind having the same version of Mojito for all your applications, or being unsure which version is currently installed globally should you be installing separate local copies. Over the long term our intent is to split the "cli" so bin/mojito would be a separate package you install globally and you'd install whichever "lib" version you need for a particular application. For now I'm going to close this request. But thanks for the quick input and interest! Keep em coming!

chrismytton:
I don't understand your reasoning?

If I want to have a separate version of mojito for each application, then I can just run `npm install mojito@x.x.x` in the root of each application, which will install that version of mojito into the app's `node_modules` directory.

I had thought that the GitHub install instructions were so that you could have a local version to hack on, installed globally so you can test it? Otherwise I don't see the point of the GitHub install instructions.

mojit0:
The reason for the "double install" is currently related more to the fact that the mojito "binary" must be installed for you to use 'mojito create'. Once you've done that you can then cd into the individual applications and install the mojito library. The current package isn't "split" into a "cli" and "lib" pair, but that's our intent. So long-term you'd globally install the command line tools, and locally install the library version your various applications depend on.

triptych:
I don't follow this either - I was able to run the app without the double install...

Issue #58, opened 2012-03-30
Upgrade to latest version of Mustache


add0n:
Upgrade Mojito to use the latest version of Mustache, which includes such niceties as dot-separated object field access.

mridgway:
This is no longer relevant as we are moving towards Handlebars.

PR #116, opened 2012-04-03
Updated mustache.js for client side and mulib for server side

mridgway:
Updated client side adapter to 0.4.2 of mustache.js
Update server side Mulib to use 'mu2' npm dependency version 0.5.13
`mojito test` passes


mridgway:
Related to https://github.com/yahoo/mojito/issues/58

matthewmueller:
+1, currently `null` inputs to the template will blow up the server.

mojit0:
Nice patch. If you'd like this merged quickly we'll need some unit tests for templating which help ensure we're not going to see any regressions, otherwise we'll merge this once we have developed those tests ourselves.

mridgway:
Are there any existing rendering tests that I can use as an example? I was under the assumption that the tests were already there and that running the test suite should identify the regressions.

I'll take a look at creating some new tests then.

mridgway:
Closing this, as even the newest Mu library does not fix the issues. I would prefer to replace Mu, or even Mustache entirely in favor of Handlebars (#174).

PR #117, opened 2012-04-03
Remove unneeded step to create Mojito application

okuryu:
Please review this one.

See also: http://developer.yahoo.com/cocktails/mojito/docs/quickstart/


mojit0:
It's not truly unnecessary, but it is incorrect. It should read:

npm install .

This ensures a local node_modules directory is created and populated with the necessary dependencies.

We'll update and push at the next opportunity.

PR #120, opened 2012-04-06
use public URL instead of yahoo corp

huang47:
now y.ahoo.it/mojito points to developer.corp.yahoo.com and it's not expected


rwaldura:
FYI I fixed the short link.

PR #122, opened 2012-04-06
adding routes in the readme file

selfcontained:
adding clarification surrounding routes in the readme file so developers are aware of it from their initial exposure to the library

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,25 @@ Go to URL:
 
     http://localhost:8666/@HelloMojit/index
 
+Add a route to map "/" to your HelloMojit.index action.  Edit your routes.json so that it looks like:
+
+```
+[{
+    "settings": ["master"],
+    "hello index": {
+        "verbs": ["get"],
+        "path": "/",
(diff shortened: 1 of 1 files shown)


rwaldura:
Thanks for the patch bmharris. 
Is it better to mention routes right upfront, in the README, or should we attempt to delay that complexity till later -- and use a full URL instead in the README? I.e. http://localhost:8666/@myMoijt/index

selfcontained:
I think the concept of routes is pretty well known amongst deveopers, more-so than a magical full URL that mojito supports.  Developers are used to the concept of routing in their mvc, and to not mention it may make some assume that it only has these magical @MojitoName.action urls.  My 2cents.

rwaldura:
bmharris: we're going to try to keep the README super lightweight, and have more solid documentation around routes in the Wiki etc. Thanks again for the patch though! Keep 'em coming!

rwaldura:
E.g. http://developer.yahoo.com/cocktails/mojito/docs/code_exs/route_config.html

PR #121, opened 2012-04-06
Configurable Doctype

gvaish:
Changed HTMLFrameMojit to have doctype configurable

Real code in commit https://github.com/gvaish/mojito/commit/5f1c1a3af5e4b1fd4b0b51b4a8e34ef3f640e499

PR #123, opened 2012-04-06
Configurable Doctype - Based on User-Agent

gvaish:
Pull 121 (https://github.com/yahoo/mojito/pull/121) can be closed.

The updated version allows configurable DOCTYPE declaration based on User-Agent.

Actual code in commit #80bd20c4
(https://github.com/gvaish/mojito/commit/80bd20c4fb3997398991a4b0c0459aed71342175)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+npm-debug.log
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Yahoo! Mojito
+# GVaish mMojito (Gaurav's enhancements to Yahoo! Mojito)
 
 Mojito is the JavaScript library implementing Cocktails, a JavaScript-based
 on-line/off-line, multi-device, hosted application platform.
@@ -66,4 +66,9 @@ Please see the LICENSE.txt file for details.
 Mojito includes the Mulib software available here:
 
 https://github.com/raycmorgan/Mu
+
+## Gaurav's website
+
+http://www.mastergaurav.com
+
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GVaish mMojito (Gaurav's enhancements to Yahoo! Mojito)
+# Yahoo! Mojito
 
 Mojito is the JavaScript library implementing Cocktails, a JavaScript-based
 on-line/off-line, multi-device, hosted application platform.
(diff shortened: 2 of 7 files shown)


gvaish:
Case study added in commit #a8cd0e2bb
(https://github.com/gvaish/mojito/commit/a8cd0e2bb0320657653369a68cc36e799b949543)

epuidokas:
Not sure what policies the Mojito team has, but I'd clean up that commit history and leave the version incrementing out of this pull request.

gvaish:
Erik,

Sure. No need to carry forward the complete commit history! :)

mojit0:
This is a reasonable extension, but our goal for HTMLFrameMojit is that it be more API driven than "automagic" and something you may often write on your own to optimize the needs of your application. For that reason we're suggesting you use this as your own variant rather than bringing this patch in as the baseline version.

gvaish:
Well, an issue... Doctype remains non-configurable in that case.
There is no way during processing (read: controller execution) that the info about "DOCTYPE" can be provided to HTMLFrameMojit except from application.json

I would be eager to know about alternate, API driven mechanism for the same.

add0n:
I think, as a general rule, that we're moving towards having API (i.e. a call into Mojito) for things such as this, since it's much more powerful and flexible.

gvaish:
Gotcha!

drewfish:
I think we should take a different approach.  I think that Mojito should make it easy for people to implement their own equivalent of HTMLFrameMoijt.

If you look at HTMLFrameMojit today, it is doing two things:  putting Mojito client in the page, and putting other assets in the page.  That's it -- those are the only things that HTMLFrameMojit is doing that has anything to do with the Mojito framework.  Any special logic beyond that is specific to each application, and should be done in the application code, not in Mojito.

Of the two Mojito-specific things that HTMLFrameMojit does, deploying the client into the page is already pretty straightforward.  (It's a single call to ac.deploy, guarded by a check on the "deploy" flag in the config.)  That means Mojito just needs to provide an easy way to put the assets in the page.  (I think a clean approach to doing this would be to take that current logic from the HTMLFrameMojit controller and move it into the "assets" AC addon, probably into a method called renderAsHTML(location) or something like that.)

gvaish:
Andrew,

Points noted. Infact, as I write this comment, I have some ideas running at the back of my mind... let me try to "revamp" the HTMLFrameMojit to "externalize" the reusable code into another component that can be used from elsewhere as well not just HTMLFrameMojit

PR #128, opened 2012-04-13
HTMLFrameMojit - Programmatic Configuration

gvaish:
Based on inputs at https://github.com/yahoo/mojito/pull/123#issuecomment-5076878, I've created a patch for configuring the following programmatically:
1. title (in html -> head)
2. doctype

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+npm-debug.log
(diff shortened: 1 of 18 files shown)


rwaldura:
We're going to take a different approach with this.
Thanks for the patch though!

gvaish:
@rwaldura Will await the new HTMLFrameMojit.

Issue #129, opened 2012-04-13
child set to null in children structure should not break the composite execution


caridy:
Sometimes, intentionally, I want to disable a particular child depending on the current context, in which case, I should be able to set the child to null. Here is an example of a dynamic children structure based on the context.

`````` javascript[
    {
        "settings": [ "master" ],
        "config": {
            "children": {
                "header": {
                    "base": "header"
                },
                "footer": {
                    "base": "footer"
                },
                "body": {
                    "base": "body"
                }
            }
        }
    },
    {
        "settings": [ "device:iphone" ],
        "config": {
            "children": {
                "header": null
            }
        }
    }
]```

patch on its way.
``````

PR #130, opened 2012-04-13
child to null should not break the ac.composite.done call

caridy:
disabling a child mojit by setting it to null in the children structure is now supported. Check issue #129 for more details.


drewfish:
Nice and clean.  Thanks for including a unit test too!

isao:
ok, your patches have been applied (couldn't auto merge), thanks Caridy

Issue #23, opened 2012-03-30
Enable node options to be passed into node via "mojito start"


add0n:
Currently, you cannot specify any node options using "mojito start", such as --debug.

add0n:
There is currently a workaround for this issue:

node --debug ./bin/mojito start

PR #131, opened 2012-04-13
Allow Node arguments to be passed via '--node' flag

epuidokas:
https://github.com/yahoo/mojito/issues/23

### DIFF
--- a/source/bin/mojito
+++ b/source/bin/mojito
@@ -1,17 +1,50 @@
-#!/usr/bin/env node
-/*
- * Copyright (c) 2011-2012, Yahoo! Inc.  All rights reserved.
- * Copyrights licensed under the New BSD License.
- * See the accompanying LICENSE file for terms.
- */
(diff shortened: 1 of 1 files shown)


drewfish:
Some folks on the Mojito team think that we should re-design how `mojito start` works.  Currently it's much more convoluted than it perhaps should be.  Once it works a lot simpler, it'll hopefully be much easier to tweak to do what you want on a case-by-case (app-by-app) bases.

epuidokas:
Not sure what other applications typically do for their start commands, but honestly, the workaround didn't seem that bad to me.
https://github.com/yahoo/mojito/issues/23#issuecomment-4848007
`node --debug ./bin/mojito start`

Maybe it's just a matter of documenting and officially supporting usage that way.

isao:
Ok, hopefully you're happy with the workaround. Don't want to have bash in the mojito command code path when the rest of it is js, so closing this for now

PR #136, opened 2012-04-19
Fix for a bug in "mojito build html5app" command.

ricallinson:
Here is small patch for a bug in "mojito build html5app" command which stopped the context from being used when the Mojito server is started as part of the build process. I have also commented out the "manifest" and "index.js" files as they are currently not working as expected.

### DIFF
--- a/source/lib/management/commands/build.js
+++ b/source/lib/management/commands/build.js
@@ -142,13 +142,8 @@ exports.buildhtml5app = function(cmdOptions, store, config, destination,
         // files to fetch through the server
         serverFiles = { css: true, js: true, json: true },
         extension,
-// Removed manifest file from output
-//        manifest = 'CACHE MANIFEST\n',
-//        indexJs = "module.exports = require('express')." +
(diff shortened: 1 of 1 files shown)


drewfish:
I would prefer that the manifest and index.js get fixed, instead of just being commented out.  (I dislike code that is just left commented out, -especially- when no reason is given in the comment for why it's commented out.  This behavior leads to a messy code base, unfortunately.)

ricallinson:
That sounds better. The "index.js" file is redundant and can be removed completely. My view on the "manifest" file is that it should be generated by the Mojito server and not the build step.

drewfish:
If this PR _just_ contained the context bug fix, it would likely be applied fairly quickly.  The changes about removing `index.js` (will this break existing users?) and the "manifest" file (is there other code that makes it?) complicate the situation considerably.

ricallinson:
I will re-submit.

rwaldura:
this issue is superseded by pull #285

PR #137, opened 2012-04-20
Added shaker command

diervo:
We created the shake command in order to run shaker easily


rwaldura:
Sweet, thanks!

isao:
Diego, we're not going to land these patches, as `mojito shaker` is now supported (require'd from the node path)

dmitris:
@isao I think you meant the mojito-shaker - https://github.com/yahoo/mojito-shaker utility (correcting for people landing on the issue through search or browsing)

PR #140, opened 2012-04-24
Update intl addon documentation to show retrieval of all keys

FabianFrank:
I updated the intl addon documentation to ensure people know about the possibility to obtain all keys at once. This is a feature we are using heavily in our team, but took as a while to find. Hope this helps.


isao:
thanks!

PR #142, opened 2012-05-04
[fix bug 4987811] Load app level modules (autoloads and addons) for 'mojito test mojit'

gomezd:
When a mojit requires modules from autoload or addons, 'mojito test mojit' does not load them, thus the test fails.
I've changed the loading so 'mojito test {app,mojit}' both use the store to load app level resources and behave the same way, simplifying a little bit the logic (further enhancements can be done though).


drewfish:
This PR has a bunch of commits that don't appear to be related (probably because it's so old). Could you perhaps create a feature branch for the delta and create a new PR from that?

(Apparently a github PR is locked to the branch, not commit, so further changes to the branch (your "develop" branch in this case) will become part of the PR as well.  You can get around this by creating a separate branch for each pull request.)

caridy:
oh man, everyone is having the same problem with the pull request. We need to document the process better. Btw, you can always use "git cherry-pick" to avoid having another clean pull request.

gomezd:
That was probably my bad. I'll update my repo and if this change is still relevant, will send a clean request. I agree the process should be better documented.
Thanks!

PR #143, opened 2012-05-05
Handlebars support in mojito.

caridy:
Let's get the ball rolling for the handlebars support in mojito. This is just a preliminary job, I plan to add all the missing features into this branch to match the current mustache support. Feel free to reject the pull request for now.

For now, I also added the new yui->3.5.0 to the dependency list although only used by the handlebars engine implementation who requires yui/handlebars. Eventually, when the transition to the new YUI node package happens, should be easy to merge this changes.

Client side is mostly covered by yui 3.5.0, we just need to create the mojito wrapper for it.

TODO: 
- tests
- figure out how to call done and composite.done without specifying the content-path as part of the view object.
- create view-engines/hb.client.js as a wrapper for YUI 3.5.0 handlebars implementation

I don't plan to add support for {{> partial}} since that requires some extra machinery to resolves paths etc.

### DIFF
--- a/source/lib/app/addons/ac/partial.common.js
+++ b/source/lib/app/addons/ac/partial.common.js
@@ -72,6 +72,32 @@ YUI.add('mojito-partial-addon', function(Y, NAME) {
             }, meta);
         },
 
+        /**
+         * This method compiles the "view" specified.
+         * The "view" must be the name of one of the files in the current
+         * Mojits "views" folder.
+         * @param {string} view The view name to be used for rendering.
+         * @param {Object} opts Optional argument depending on the engine.
+         * @return {string} compiled view.
+         */
+        compiler: function(view, opts) {
+            var renderer,
+                mojitView,
+                instance = this.command.instance;
+
+            if (!instance.views[view]) {
+                Y.log('View "' + view + '" not found', 'debug', NAME);
+                return;
+            }
+
+            mojitView = instance.views[view];
+            renderer = new Y.mojito.ViewRenderer(mojitView.engine);
+
+            Y.log('Compiling "' + view + '" view for "' + (instance.id || '@' +
+                instance.type) + '"', 'debug', NAME);
+
+            return renderer.compiler(mojitView['content-path'], opts);
+        },
 
         /**
          * This method calls the current mojit's controller with the "action"
--- a/source/lib/management/commands/compile.js
+++ b/source/lib/management/commands/compile.js
@@ -559,12 +559,32 @@ compile.views = function(context, options, callback) {
                         renderer = new (Y.mojito.addons.viewEngines[engine])();
 
                         if (typeof renderer.compiler === 'function') {
-                            renderedView = JSON.parse(
-                                renderer.compiler(source).toString()
-                            );
+                            // there is not need to stringify this anymore
+                            // since we have a custom method to take care of that
+                            // down the road.
+                            renderedView = renderer.compiler(source);
+                            
                             yuiModuleCacheWriter.createNamespace('compiled.' +
                                 mojitNs + '.views').cache(viewName,
-                                renderedView);
+                                renderedView).toString = function () {
+                                    var chunks = [],
+                                        viewName;
+                                    // to support precompiled views as javascript
+                                    // we need to mess around a little bit. If the compiled version
+                                    // is an object exposing the compiled view through .toString() method,
+                                    // the we should be able to inject functions into the cache structure.
+                                    for (viewName in this) {
+                                        if (typeof this[viewName] !== 'function') {
+                                            chunks.push('"' + viewName + '":' + this[viewName].toString());
+                                        }
+                                    }
+                                    // we are synthetically creating the result string for each group of
+                                    // views. We intentionally call for toString() for every view so we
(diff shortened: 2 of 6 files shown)


FabianFrank:
great job @caridy let me know if you need help :-)

caridy:
If you want to try out the current implementation, check the demo here:

https://github.com/caridy/mojito-handlebars-demo

caridy:
@Schnitz of course I need help jajajaja

source/lib/management/commands/compile.js
```diff
@@ -971,6 +991,12 @@ YuiModuleCacher.prototype.createNamespace = function(ns) {
             _c: {},
             cache: function(k, v) {
                 this._c[k] = v;
+                // exposing the cache object in case we want to do
```
drewfish, line 994: This comment should really be a comment introducing the cache() method (plus add yuidoc tags).

source/lib/app/addons/ac/partial.common.js
```diff
@@ -72,6 +72,32 @@ YUI.add('mojito-partial-addon', function(Y, NAME) {
             }, meta);
         },
 
+        /**
```
drewfish, line 75: Why does this `ac.partial.compiler()` method exist?  Where/how might it be used?

caridy, line 75: Today, with mustache, we can add "mojito-partial-addon" and call for ac.partial.compiler('foo') to get the stringify version of the template, and pass it to the client side if we want to use that there.

With handlebars, we should be able to do the exact same. Here is an example:

https://github.com/caridy/mojito-handlebars-demo/blob/master/mojits/HelloHandlebars/controller.server.js

The difference between mustache and handlebars at this point is that mustache produce a string that has to be processed at the client side by the mustache engine, while handlebars produce a Javascript code that represents a function, and by calling that function, it will return the rendered template.

But from the mojito point of view, they are the same thing, so, ac.partial should behave the same.

source/lib/management/commands/compile.js
caridy, line 994: agreed, I will adjust it.

source/lib/app/addons/ac/partial.common.js
drewfish, line 75: Why would anyone want to call ac.partial.compiler() and send the results to the client side?  `mojito compile views` is meant to take care compiling the views, so that nothing special has to be done in the controller.

isao:
looks like this won't apply cleanly now, sorry. but really looking forward to it.
side note: since views are _.hb._ it doesn't need to be 100% interoperable with the mojito mu implementation

caridy:
@isao, the idea is to have the same support since mustache is a subset of handlebars, so we can drop mustache implementation and still support *.mu.html files as an alias of *.hb.html. I don't see why we need to have both when handlebars can perfectly handle mustache templates.

source/lib/app/addons/ac/partial.common.js
caridy, line 75: @drewfish "mojito compile views" is pretty dummy today (it doesn't select specific views per context, etc). Shaker does a better job in this area, but still I feel that, sometimes, having the ability to serialize a partial and sending it over to the client side to execute a particular process without all the hazard of compiling all of them during build time is a nice to have feature, specially if you have some sort of logic to pick the right view programmatically rather than a fixed value.

drewfish, line 75: Hmm... OK.  I would say that if `ac.partial.compiler()` is sufficiently useful (as you argue) then we should let it drive a change to the view engine API and require that -all- view engines (including `mu`) support it (via the new `compiler` method).

drewfish, line 75: And we should perhaps update `mojito compile views` to be better about contexts, etc.  It seems a little dangerous to invent a new feature when the old one isn't sufficient, when we could instead fix up the old feature.  We don't want to end up with a million slightly-different ways of doing things.

caridy, line 75: I will argue that this type of rollup process (mojito compile views) should be made by shaker rather than mojito itself. But about the specific feature we are discussing here (partial->compiler), I added to facilitate (in the near future) the use of {{> partial}} for handlebars templates. At this point, mojito-partial-addon is the only external addon that knows how to resolve (based on the name of the view) the partial fullpath, etc. So, if we are going to support this feature anytime soon, we need a way to pre-compute partials before executing the composite view. That being said, we can stick to the plan of having no compiler routine at the addon level if you feel strong about this.

drewfish:
Sorry, I didn't mean to let this pull request linger so long.

If you made a pull request that _just_ added handlebars, it would likely be accepted pretty quickly.

The `ac.partial.compiler()` stuff worries me.  I'm worried that it is taking Mojito in the wrong direction.  If/when Mojito supports "partial" (which I think of as "includes") in templating engines, that shouldn't have to involve the controller (or even addons or ActionContext) -- it should just work (from the view of the developer using Mojito).

caridy:
sure, let's close this one then. I will pull another next week!

isao:
Looking forward to it, thanks Caridy

PR #144, opened 2012-05-08
pull 130 part 2 via Caridy@ - allow HTMLFrameMojit child config to specify action as expected

isao:
I did a crude test case here: https://github.com/isao/child-action-testcase

See also https://github.com/caridy/mojito/commit/2bb0d7027be5fa9147985d8a6aa87a8ce6c95ead

Thank you Caridy@


isao:
ah looks like all changes on a branch get included on a pull request. closing this so I can submit separately.

PR #153, opened 2012-05-16
extra dots in model name mean something now, rename to get example working

isao:
minor fix to example app.


isao:
taking advantage of minor commit to test ci hooks

PR #158, opened 2012-05-22
Adding reST documentation

zhouyaoji:
The reST files that are used to render the Mojito docs on YDN should also be in the Mojito project, so developers can contribute to the docs as well.


isao:
Please redo this pull-request off a branch containing only the files you want changed/added

PR #162, opened 2012-05-24
Fix a bug on mustache that prevents page rendering (crashes) when binding data contains null

macmacbr:
Fix a bug on mustache that prevents page rendering (crashes) when binding data contains null.

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -15,10 +15,6 @@ Currently Deprecated
 
 ### Deprecated but Available
 
-* (2012-04-23) The `autoload/` directory is going away in favor of
-`yui_modules/`, which better reflects its contents.  Everthing else about it is
-the same, only the name has changed.  You can start using `yui_modules/` today.
-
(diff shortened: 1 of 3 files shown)


drewfish:
I think we have a preference to update the mustache library outright, pull request #116, rather than try to patch an old version.

mridgway:
Unfortunately even the latest version of the Mu library doesn't fix this issue either. I think the Mu library needs to be replaced, as it doesn't seem up to par with the Mustache spec. Dot notation doesn't work with Mu either.

isao:
Thanks Marco, I cleaned up the pull request a bit and resubmitted it for you. pull #207

closing this one.

rwaldura:
Yeah, we're ditching Mu in favor of Handlebars -- see pull request #179.

isao:
Yup can't wait for pull #179 too. As-is the handlebars pull is opt-in, so this bug fix is nice until the internal mu renderer goes away.

rwaldura:
Unfortunately we're having to delay Handlebar support, so we're going to pull in request #211 which fixes the NULL case.

PR #163, opened 2012-05-25
Upgrade to YUI 3.5

mridgway:
I'm not sure if this is still being worked on by someone else, but I'd like to open this up and see if we can get some traction with this. Maybe @davglass can chime in too.

So far all of the tests run, but they do not pass. Here are my test failures: https://gist.github.com/2785983

This pull request should be considered incomplete until we can get all tests to pass.

PR #165, opened 2012-05-25
Migrating over to YUI 3.5.1 official install

davglass:
This supersedes the request #163 so that it comes from my branch and will be easier to merge in and test.

There is only one failing test now in "precompute" mode and zero in "ondemand" mode.

```
FAILURE DETAILS:
================
mojito-store-server-tests :: Store tests :: appConfig deferAllOptionalAutoloads
breg
Expected: undefined (undefined)
Actual: /static/mojito/autoload/transport/beanregistry/registry.client-optional.js (string)


Total tests: 331    ✔ Passed: 309 ⚑ Deferred: 21    ✖ Failed: 1   93% pass rate
```

### DIFF
--- a/source/lib/management/commands/test.js
+++ b/source/lib/management/commands/test.js
@@ -33,7 +33,7 @@ var pathlib = require('path'),
     coverageDir = pathlib.join(resultsDir, 'coverage'),
     coverageFile = pathlib.join(coverageDir, 'coverage.json'),
 
-    YUI = require('yui3').YUI,
+    YUI = require('yui').YUI,
     YUITest = require(pathlib.join(fwTestsRoot,
         'harness/lib/yuitest/javascript/build/yuitest/yuitest-node')).YUITest,
     TestRunner = YUITest.TestRunner,
@@ -119,12 +119,14 @@ function collectRunResults(results) {
 }
 
 function configureYUI(YUI, store, load) {
-    YUI.GlobalConfig.groups['mojito-fw'] = store.getYuiConfigFw('server', {});
-    YUI.GlobalConfig.groups['mojito-app'] = store.getYuiConfigApp('server', {});
-    YUI.GlobalConfig.groups['mojito-mojits'] = store.getYuiConfigAllMojits(
-        'server',
-        {}
-    );
+    YUI.applyConfig({
+        useSync: true,
+        groups: {
+            'mojito-fw': store.getYuiConfigFw('server', {}),
+            'mojito-app': store.getYuiConfigApp('server', {}),
+            'mojito-mojits': store.getYuiConfigAllMojits('server,', {})
+        }
+    });
 }
 
 
--- a/source/lib/store.server.js
+++ b/source/lib/store.server.js
@@ -3733,21 +3733,14 @@ ServerStore.prototype = {
             return { sorted: Object.keys(required), paths: sortedPaths };
         }
         
-        //TODO davglass USE resolve here!
+        //TODO davglass USE resolve her
         Y = YUI({ useSync: true }).use('loader-base');
-        loader = new Y.Loader({ lang: ctx.lang });
-
-        // We need to clear YUI's cached dependencies, since there's no
-        // guarantee that the previously calculated dependencies have been done
-        // using the same context as this calculation.
-        delete YUI.Env._renderedMods;
-
-        // This approach seems odd, but it's what the YUI Configurator is also
-        // doing.
+        //Use ignoreRegistered here instead of the old `delete YUI.Env._renderedMods` hack
+        loader = new Y.Loader({ lang: ctx.lang, ignoreRegistered: true });
+        
+        //Only override the default if it's required
         if (yuiConfig && yuiConfig.base) {
             loader.base = yuiConfig.base;
-        } else {
-            loader.base = Y.Env.meta.base + Y.Env.meta.root;
         }
         loader.addGroup({modules: modules}, mojitType);
         loader.calculate({required: required});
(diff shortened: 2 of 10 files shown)


source/lib/store.server.js
```diff
@@ -40,11 +40,11 @@ var libfs = require('fs'),
...
     // when using loader for dependency calculations.
-    utilYUI = require('yui3').YUI,
-    serverYUI = require('yui3').YUI,
-    clientYUI = require('yui3').YUI,
+    utilYUI = require('yui').YUI,
+    serverYUI = require('yui').YUI,
+    clientYUI = require('yui').YUI,
 
-    Y = utilYUI().useSync('intl'),
+    Y = utilYUI({ useSync: true }).use('intl'),
```
drewfish, line 47: Does this config ({useSync:true}) only apply to that particular YUI instance?

davglass, line 47: Yes.

source/lib/management/commands/test.js
```diff
@@ -119,12 +119,14 @@ function collectRunResults(results) {
 }
 
 function configureYUI(YUI, store, load) {
-    YUI.GlobalConfig.groups['mojito-fw'] = store.getYuiConfigFw('server', {});
-    YUI.GlobalConfig.groups['mojito-app'] = store.getYuiConfigApp('server', {});
-    YUI.GlobalConfig.groups['mojito-mojits'] = store.getYuiConfigAllMojits(
-        'server',
-        {}
-    );
+    YUI.applyConfig({
```
drewfish, line 122: `useSync` is turned on here, but not turned off elsewhere.  Is that OK?

(Hmmm... I guess we won't know until we do a deeper analysis of where `use()` is being called.  It's OK to do things synchronously while the server is starting up, but after that we should definitely be doing things asynchronously.  That perhaps answers my own question:  set `useSync:false` after the server has started.)

davglass, line 122: Yeah, you could totally set it to false after the server starts, I just changed the things needed to make 3.5.1 run, I know little to nothing about what Mojito is doing under the hood ;)

source/lib/store.server.js
```diff
@@ -3732,21 +3732,15 @@ ServerStore.prototype = {
...
-        // guarantee that the previously calculated dependencies have been done
-        // using the same context as this calculation.
-        delete YUI.Env._renderedMods;
-
-        // This approach seems odd, but it's what the YUI Configurator is also
-        // doing.
+        
+        //TODO davglass USE resolve her
+        Y = YUI({ useSync: true }).use('loader-base');
+        //Use ignoreRegistered here instead of the old `delete YUI.Env._renderedMods` hack
```
drewfish, line 3738: If we're deleting the hack, there's no need to continue talking about it.  Perhaps a comment like this would be better:

``` javascript
// We need to ignoreRegistered so that we get a full list of dependencies (including those that
// happened to be loaded by some other code).
```

source/lib/tests/autoload/app/autoload/loader-tests.common.js
```diff
@@ -116,15 +116,15 @@ YUI.add('mojito-loader-tests', function(Y, NAME) {
...
-                '3.4.1/build/yui/yui-debug.js&' +
-                '3.4.1/build/yui-base/yui-base-debug.js&' + 
-                '3.4.1/build/collection/collection-debug.js&' +
-                '3.4.1/build/oop/oop-debug.js&' +
-                '3.4.1/build/features/features-debug.js&' +
-                '3.4.1/build/dom-core/dom-core-debug.js&' +
-                '3.4.1/build/dom-base/dom-base-debug.js&' +
-                '3.4.1/build/selector-native/selector-native-debug.js&' +
-                '3.4.1/build/selector-css2/selector-css2-debug.js'
+                Y.version + '/build/yui/yui-debug.js&' +
```
drewfish, line 119: nice

source/lib/management/commands/test.js
drewfish, line 122: OK, yeah, we (mojito team) will have to do the rest of the validation for this PR.

davglass, line 122: That was my thoughts, just giving a helping hand with what I could do ;)

source/lib/store.server.js
```diff
@@ -3732,21 +3732,15 @@ ServerStore.prototype = {
...
             return { sorted: Object.keys(required), paths: sortedPaths };
         }
-
-        Y = YUI().useSync('loader-base');
-        loader = new Y.Loader({ lang: ctx.lang });
-
-        // We need to clear YUI's cached dependencies, since there's no
-        // guarantee that the previously calculated dependencies have been done
-        // using the same context as this calculation.
-        delete YUI.Env._renderedMods;
```
drewfish, line 3742: If I add this line back, the test passes again :(

davglass, line 3742: It's probably because your not using `loader.resolve` to resolve the deps. 

You can either put that line back, or switch to calling resolve & getting the modules from the return.

drewfish, line 3742: It seems that `resolve()` is the cleaner API call.  Alas, I couldn't figure out how to change the `loader.calculate({required:required})` into a `loader.resolve()` call.

davglass, line 3742: This should do it:

``` javascript
loader.require(required);
var out = loader.resolve(true); //This forces a calculate
```

drewfish, line 3742: Yeah, I should have dug through the docs more.

Unfortunately, it's still tunneling the requirements from a previous, and I have to resort to the `delete YUI.Env._renderedMods` hack to prevent that.

davglass, line 3742: That's fine, I need to get a repro case of this so I can fix it properly. I thought I covered all the issues with `resolve` in my tests.

PR #157, opened 2012-05-18
Repaired merge for array items. Updated tests.

mojit0:
Repair for https://github.com/yahoo/mojito/issues/12

PR #170, opened 2012-05-29
Replace DaliProxy with simple IO

mridgway:
DaliProxy doesn't seem to be adding anything that is used by Mojito, Ric agreed in a discussion last week. Removing DaliProxy and replacing it with a simple IO call should significantly reduce the complexity of tunnel calls and allow for better error handling to be done (more pull requests coming for this I hope). I also added app level configuration for the tunnelTimeout (this was hardcoded before).

I want to run some more functional tests with this code, but so far I have tested it with the 'examples/sandbox/view_refresh' application.


drewfish:
+1 to the idea.  (Will review later.)

(This indicates the value of this idea:

```
+ 55 additions
- 1,826 deletions
```

)

mridgway:
I ran the code on our current project and there seems to be an issue with metaMerge recursion that I'm trying to work out:

RangeError: Maximum call stack size exceeded
    at Function.APPLY_OVERFLOW (native)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:132:41)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:138:55)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:138:55)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:138:55)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:138:55)
    at Object.metaMerge (…/node_modules/mojito/lib/app/autoload/util.common.js:138:55)
    at Object._updateBuffer (…/node_modules/mojito/lib/app/mojits/TunnelProxy/controller.server.js:34:43)
    at Object.flush (…/node_modules/mojito/lib/app/mojits/TunnelProxy/controller.server.js:50:22)

drewfish:
That might be fixed by pull #157.

mridgway:
Just applied the patch from #157 and it does indeed fix the issue. Everything looks to be running normally. I'll continue testing it though.

caridy:
The other thing that I noticed is that some of the features from DaliProxy are now gone. E.g, performance metrics for every request, something that was built-in as part of the "makeRequest" routine. This has to be included in the new implementation, for those who care about performance (like me!).

I don't see support for "multipart" response headers, even though I have never used it, but I remember the endless discussions about this subject (for the record I was against it). I don't think mojito supports that today, but I'm not 100% sure.

Anyway, my 2 cents.

mridgway:
Caridy, can you point me to where makeRequest had built in performance metrics? I see there is a PerformanceMonitor  class in there, but I don't see a reference to it being used anywhere. I don't think Mojito ever made use of it either. I agree this should be something that is added, perhaps in a future PR though.

caridy:
@mridgway yes, just look for references to "getTime()" which is the way they mark important moments. Now, I'm not so sure this is available or not, in the old Dali implementation for maple, it was available through the custom beans, but I know mojito is kind of obscure in terms of accessing "beans" directly. I will have to investigate more.

isao:
This request is very much appreciated, but unfortunately, this has aged out:

   "This pull request cannot be automatically merged."

Closing for now. Please reopen or make a new one referencing pull 170.

mridgway:
Pull requests shouldn't be closed due to being out of date. They can be updated with another commit to make them easy to merge or they can be pulled and fixed by the developers locally. Closing requests like this just means contributors won't make new ones because they have no idea whether it will be pulled before it can be merged by Github's button.

isao:
By closing it, I didn't imply that something based on this won't ever be accepted. You clearly put a lot of good work into it, and it's something we need. It aged out through no fault of yours.

Anyway, I'll reopen.

mridgway:
I fully understand. I'm just speaking as a matter of process.

I'll update this PR tonight if I get a chance.

isao:
Right, I was thinking it's just a flag and work isn't lost. And any day we can eliminate 1.8k loc without breaking is a good day :)

What is the situation with instrumentation?

mridgway:
From the sound of it, Mojito didn't provide this functionality before, so I don't think it should be a part of this PR. I think this is something we can add as a feature later (and I'm highly interested in having it as well).

isao:
that's cool. just wasn't sure if anything was being taken away (besides
bloat) :)

thanks Michael

-----Original Message-----
From: Michael Ridgway
<reply+i-4789727-903696e1ee2531cecf509e9a329571f6ad4ecc73-46951@reply.githu
b.com>
To: Isao Yagi isao@yahoo-inc.com
Subject: Re: [mojito] Replace DaliProxy with simple IO (#170)

From the sound of it, Mojito didn't provide this functionality before, so
I don't think it should be a part of this PR. I think this is something we
can add as a feature later (and I'm highly interested in having it as
well).

---

redonkulus:
Any update on this pull request? ETA perhaps?

mridgway:
Reopening fresh PR.

PR #172, opened 2012-05-30
upgrade to yui 3.5.1

drewfish:
Here's an update to Dave's PR #165.  I fixed the failing regression, and ran some sample apps (and fixed issues found during that).

### DIFF
--- a/source/lib/app/autoload/loader.common.js
+++ b/source/lib/app/autoload/loader.common.js
@@ -31,7 +31,7 @@ YUI.add('mojito-loader', function(Y, NAME) {
 
             paths = paths || {};
             if (!Y.Object.size(paths)) {
-                cb(null);
+                cb();
                 return;
             }
 
@@ -50,7 +50,7 @@ YUI.add('mojito-loader', function(Y, NAME) {
             }
             scriptsToLoad = Y.Object.keys(scriptsToLoad);
             if (!scriptsToLoad.length) {
-                return cb(null);
+                return cb();
             }
 
             Y.log('loading ' + scriptsToLoad.join(', '), 'mojito', NAME);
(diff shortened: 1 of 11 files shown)


source/lib/app/autoload/loader.common.js
```diff
@@ -59,23 +59,8 @@ YUI.add('mojito-loader', function(Y, NAME) {
...
-                            done = true;
-                        }
-                    }
-
-                    if (done) {
-                        Y.log('SUCCESS', 'mojito', NAME);
-                        cb(null);
-                    }
+                    Y.log('SUCCESS', 'mojito', NAME);
+                    cb(null);
```
mojit0: why the explicit null?

source/lib/store.server.js
```diff
@@ -3733,20 +3731,20 @@ ServerStore.prototype = {
...
 
-        Y = YUI().useSync('loader-base');
-        loader = new Y.Loader({ lang: ctx.lang });
+        Y = YUI({ useSync: true }).use('loader-base');
+        Y.applyConfig({ useSync: false });
 
         // We need to clear YUI's cached dependencies, since there's no
         // guarantee that the previously calculated dependencies have been done
         // using the same context as this calculation.
         delete YUI.Env._renderedMods;
```
mojit0, line 3741: missing semi?

drewfish, line 3741: I just fixed that.  Refresh the PR.

source/lib/app/autoload/loader.common.js
drewfish: It's not strictly necessary.  The first argument of the callback is an error object, and by using `cb(null)` I'm say "really really there's no error".  A matter of style I guess.  I can remove it.

mojit0: As a matter of style I like to keep semantics between null and undefined clear. Passing null here means arguments.length == 1 on the other side. That's different than allowing it to remain 'undefined'.

mridgway:
Nice.

PR #179, opened 2012-06-01
clean pull request for handlebars template engine for mojito server side

caridy:
To use handlebars engine, you just need to name your views like this:

<mojitName>/views/index.hb.html

then in your controller, use this definition:

```
        ac.done({
            foo: 1
        }, {
            view: {
                name: "index",
                engine: "hb"
            }
        });
```

hopefully, since handlebars support mustache syntax, we can have this as the default engine, and pipe *.mu.html as alias of *.hb.html to use the same engine.

### DIFF
--- a/source/lib/app/addons/view-engines/hb.server.js
+++ b/source/lib/app/addons/view-engines/hb.server.js
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2011-2012, Yahoo! Inc.  All rights reserved.
+ * Copyrights licensed under the New BSD License.
+ * See the accompanying LICENSE file for terms.
+ */
+
+
+/*jslint anon:true, sloppy:true, nomen:true, node:true*/
+/*global YUI*/
+
+
+YUI.add('mojito-hb', function(Y, NAME) {
+
+    var fs = require('fs'),
+        HB = require('yui/handlebars').Handlebars,
+        cache = YUI.namespace('Env.Handlebars');
+
+    /**
+     * Class text.
+     * @class HandleBarsAdapterServer
+     * @private
+     */
+    function HandleBarsAdapter(viewId) {
+        this.viewId = viewId;
+    }
+
+    HandleBarsAdapter.prototype = {
+
+        /**
+         * Renders the handlebars template using the data provided.
+         * @param {object} data The data to render.
+         * @param {string} mojitType The name of the mojit type.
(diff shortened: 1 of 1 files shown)


drewfish:
In Mojito there is no default view engine.  The view engine that matches the view (mu for .mu.html views, and handelbars for .hb.html views) is used.  This also means that you don't need to specify `engine:` in `ac.done()` -- as long as there's a view engine that matches the `.{ve}.html` part of the view then that will be used.

isao:
love it. thanks!
(team is going to write some tests first though fyi)

mridgway:
+1

This should become the default and recommended view engine.

isao:
Ok, I'm told we're going to write tests for this before landing it on the mainline.

caridy:
@isao if you provide more details where should we put the tests, I can help as well, but I didn't found any previous tests for "mu".

isao:
Bill, Scott, and Lichun are reworking the tests. Need to make it easier for us to write framework (and app for that matter) unit tests. And making our functional tests public/open too. I'll ask them to keep you in the loop. 

I thought we'd land this as a kind of opt-in semi-supported preview, but it was decided to do more complete doc/test/etc roll out.

(OTOH, you can also release this as a community node/npm module.)

source/lib/app/addons/view-engines/hb.server.js
```diff
@@ -0,0 +1,105 @@
...
+     * @param {boolean} bypassCache Whether or not we should rely on the cached content.
+     * @return {object} literal object with the "raw" and "template" references.
+     */
+    function getTemplateObj (tmpl, bypassCache) {
+        var str;
+        if (!cache[tmpl] || bypassCache) {
+            Y.log ('Loading template from disk: '+tmpl, 'mojito', 'qeperf');
+            // TODO: should we try/catch this? I don't see any try/catch when reading files
+            //       in the rest of the repo, is this the intention?
+            str = fs.readFileSync(tmpl, 'utf8');
```
mridgway, line 33: Is there any reason to make this synchronous? The only thing I can think of is getting multiple hits before the compilation is cached.

caridy, line 33: Yes. Keep in mind that this gets executed once per process per view unless you use bypassCache.

mridgway:
Merged in #356. Thanks @caridy!

PR #181, opened 2012-06-02
View Engine Code Example

zhouyaoji:
Created a code example to show Mojito app developers how to add a view engine to their apps. The example takes users through the steps to install 'handlebars' and then create a view engine addon to use the Handlebars rendering engine.


isao:
this contains all the same commits as your pull request 180, except "215acce Added unit tests for controller."

is that on purpose? should I just pull this one?

PR #182, opened 2012-06-02
support for custom binders when calling ac.done

caridy:
having a structure like this at the mojit level:

```
/mojitName
   /binders/foo.js
   /binders/bar.js
   /views/index.mu.html
   /views/baz.mu.html
```

you can invoke ac.done in your action to customize which binder to use:

```
ac.done(data, { 
    view: {
        name: "index",
        binder: "foo"
    }
});
```

or more customization

```
ac.done(data, { 
    view: {
        name: "baz",
        binder: "bar"
    }
});
```

in general, if there is a file defining a binder, you can now attach it to any view.

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -15,10 +15,6 @@ Currently Deprecated
 
 ### Deprecated but Available
 
-* (2012-04-23) The `autoload/` directory is going away in favor of
-`yui_modules/`, which better reflects its contents.  Everthing else about it is
-the same, only the name has changed.  You can start using `yui_modules/` today.
-
 * (2012-04-23) The `.guid` member of Mojito metadata (such as binder metadata)
 is going away.  Often there's an associated member which more specifically
 expresses the intent of the unique ID (for example `.viewId` or `.instanceId`).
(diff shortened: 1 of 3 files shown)


redonkulus:
+1 to this, as its sorely needed to serve different binders across different devices (mobile vs desktop). What is the status / ETA on this?

isao:
:+1: look great. ran it through the secondary tests (pull-job 23) but it got kicked out for jslint errors.

mridgway:
@drewfish Does the new resource store affect any of the code in this PR? I saw that the notes mentioned that it made this feature possible, so should this be re-implemented to make use of the new functionality in the RS?

drewfish:
I think it should still work with the new resource store.  The new store continues to set 'binder-url', 'binder-path', 'binder-module', 'binder-yui-sorted' in the mojit instance.

FabianFrank:
When looking at @redonkulus use case, I think it would be good that when we have
index.js
index.iphone.js

Mojito automatically serves the index.iphone.js binder. People shouldn't have to implement this logic by themselves. We should have this kind of automatic file picking for devices across the framework, I think.

drewfish:
@FabianFrank That's how Mojito works today.  If you say `ac.done(..., {view: {name: 'index'}});` it will use the `index.iphone.js` binder for iphone users.

What the original request wants is a way to use a binder that doesn't match the name of the action.

PR #189, opened 2012-06-06
change array cp from splice (wrong) to slice. Thank you Vinit Jain [bz 5593789]

isao:
[].splice(0) is wrong, but happens to work except in IE <=8


add0n:
This looks and tests fine.

PR #196, opened 2012-06-08
work around regression in yui@3.5.1 Y.io()

drewfish:


drewfish:
The workaround encorporates into Mojito the fix that will eventually go into YUI:
https://github.com/yui/yui3/commit/5fcddf2efbc9645d24058b40550bc013783f8444

PR #197, opened 2012-06-08
Fix mojito 'test' command for single/standalone(shared) mojit testing.

gomezd:
Fixed 'mojito test' to always use server store to load resources so 'app' and 'mojit' behave the same way.
This allows to test standalone mojits with their onw npm dependencies from node_modules, as well as testing mojits that require autload/yui_modules or addons.

### DIFF
--- a/source/lib/management/commands/test.js
+++ b/source/lib/management/commands/test.js
@@ -665,9 +665,20 @@ runTests = function(opts) {
 
         testModuleNames = ['mojito', 'mojito-test'];
 
+        ResourceStore = require(pathlib.join(targetMojitoPath,
+                'lib/store.server.js'));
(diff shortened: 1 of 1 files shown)


isao:
notes:
- passes regressions mojito-github-commit-pull job 14 (internal)
- David provided test cases separately

mojit0:
Within the context of an application this appears to work. But it breaks 'mojito test' for those working on the framework itself. If you cd into a new checkout of mojito, then descend into the source directory and execute 'mojito test' this patch causes it to fail without executing any tests. I think we'd need a version that works for all scenarios before we could merge this in.

rwaldura:
Fair to close then?

PR #198, opened 2012-06-08
[fix 5496940] Call Y.JSON for platforms that don't have native JSON support (now YUI 3.5 compliant)

add0n:
Another PR to make these calls YUI 3.5.X compliant


isao:
applied. thanks. (cherry-picked first and last to 4b05b29 d4b9c54 on develop.)

PR #200, opened 2012-06-11
[BZ 5518043] Ensure that onBinderComplete gets called the correct number of times

kfay:
There are two cases where onBinderComplete doesn't get called the correct number of times. If called too few times, the bindersComplete counter is off and binders won't get bound.

Additionally, if mojito needs to fetch a binder for first-time use, due to the async nature of Y.use(), there is the potential for 2 binder instances to be created for a single view node.


isao:
looks good, passes regressions, being applied in real apps, soo..

PR #187, opened 2012-06-05
Allow user-provided middlewares with a name starting with 'mojito-' to gain access to the resource store, YUI instance, logger, context, etc.

jlecomte:

PR #204, opened 2012-06-13
Allow user-provided middlewares with a name starting with 'mojito-' to gain access to the resource store, YUI instance, logger, context, etc.

isao:
this pull request is same as Julien's pull #187 applied to develop.


isao:
merged

PR #211, opened 2012-06-16
[BZ 5515521] More null fixes for Mustache

kfay:
Mustache still crashes if a non-enumerable null value is passed to a view or if there's a null in an array. This pull request addresses those cases.

Case 1:
    Controller:
        ac.done( { foo: null } );
    View:
        {{foo}}

Case 2:
    Controller:
        ac.done( { foo: [ 1, 2, null, 4 ] }
    View:
        {{foo}}

### DIFF
--- a/source/lib/app/libs/Mulib/Mu.js
+++ b/source/lib/app/libs/Mulib/Mu.js
@@ -147,7 +147,7 @@ Mu.normalize = function Mu_normalize(context, name) {
         val = val.call(context);
     }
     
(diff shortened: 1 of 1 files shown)


kfay:
I couldn't find the Mu lib tests, but if there is a test suite I can update, please point me to them. Thanks!

add0n:
Karen -

Thanks for this. I'll get it integrated in this sprint!

Cheers,
- Bill

PR #216, opened 2012-06-19
Bug fix for "mojito build html5app" command

ricallinson:
This is a fix for where the command-line context was not being used by the "mojito build html5app" command.


rwaldura:
superseded by #285

Issue #223, opened 2012-06-21
[bz5646042] `mojito start` fails in 0.3.27


isao:
to reproduce:
- upgrade to mojito 0.3.27
- `mojito start`
- error for an app that worked pre 0.3.27, but has YUI3.4.x/Jsdom dependencies:
  
  ```
  ✖ There was an error starting the application:
  ✖ Cannot read property 'documentElement' of null
  TypeError: Cannot read property 'documentElement' of null
  at Object.fn (/Users/isao/Repos/mojito/myfork/source/node_modules/yui/dom-base/dom-base-min.js:7:50)
  ```

mridgway:
Is this relevant any more? Looks like this can be closed since we are recommending upgrading to later versions.

FabianFrank:
I also think it can be closed.

ramses0:
$ mojito version
path.existsSync is now called `fs.existsSync`.
Version of mojito is 0.3.30

...following: http://developer.yahoo.com/cocktails/mojito/docs/getting_started/mojito_getting_started_tutorial.html#make-the-application

$ mojito test app .path.existsSync is now called `fs.existsSync`.
[warn] undefined: overriding application config value: env
[WARN] MojitoServer: your mojit "HelloMojit" has a server affinity and these client-related deps: node-base, node-core, dom-core, dom-base, node-event, node-pluginhost, node-screen, dom-screen, dom-style, dom-style, node-style
[WARN] MojitoServer: Mojito may be unable to start, unless you have provided server-side DOM/host-object suppport

...snip...

timers.js:103
            if (!process.listeners('uncaughtException').length) throw e;
                                                                      ^
TypeError: Cannot read property 'documentElement' of null
    at Object.YUI.add.requires [as fn] (/usr/local/lib/node_modules/mojito/node_modules/yui/dom-base/dom-base-min.js:7:50)

...tutorial or mojito is broken.

rwaldura:
@ramses0 is your code nothing more than the tutorial app? what's in your HelloMojit?
@zhouyaoji

ramses0:
Yep, this is "base case" stupidly following the tutorial.  I tried changing controller.server.js to controller.client.js to maybe adjust the "affinity" (?) but it gave different error messages which I am not familiar enough with mojito / node to figure out off the top of my head.

[...]

PR #224, opened 2012-06-21
log warning if server mojit has dom dependency.

isao:
to address (not fix) issue #223
mojito 0.3.27 upgrades to yui 3.5.1, which removed built-in jsdom support on node.js.

### DIFF
--- a/source/lib/store.server.js
+++ b/source/lib/store.server.js
@@ -3611,6 +3611,19 @@ ServerStore.prototype = {
                 }
             } // foreach mojitType
         } // foreach env
+
+        // log warning if server mojit has dom dependency
+        Y.Object.each(this._mojitYuiRequired.server, function(val, mojit) {
+            var deps = (val['*'] && val['*'].join()) || '',
+                badre = /\b(dom-\w+|node-\w+|io-upload-iframe)/g,
+                isbad = deps.match(badre);
+
+            if (isbad) {
+                logger.log('your mojit "' + mojit + '" has a server affinity and these client-related deps: ' + isbad.join(', '), 'WARN', NAME);
+                logger.log('Mojito may be unable to start, unless you have provided server-side DOM/host-object suppport', 'WARN', NAME);
+            }
+        });
+
     },
 
 


isao:
this change will, where relevant, output log statements during `mojito start` like:

```
[WARN] MojitoServer: your mojit "foo" has a server affinity and these client-related deps: node-base, node-core, dom-core, dom-base, io-upload-iframe
[WARN] MojitoServer: Mojito may be unable to start, unless you have provided server-side DOM/host-object suppport
```

source/lib/store.server.js
```diff
@@ -3611,6 +3611,19 @@ ServerStore.prototype = {
                 }
             } // foreach mojitType
         } // foreach env
+
+        //log warning if server mojit has dom dependency
```
mojit0: Not a big fan of "no space after //". I think we should standardize on "real sentences" for comments. So:

// Log a warning if server mojit has a dom dependency.

source/lib/store.server.js
```diff
@@ -3611,6 +3611,19 @@ ServerStore.prototype = {
...
+        Y.Object.each(this._mojitYuiRequired.server, function(val, mojit) {
```
mojit0: In general I'd like to see us hide code of this type behind a config flag. So we'd see this if we had debug/verbose on but wouldn't run the check otherwise.

source/lib/store.server.js
```diff
@@ -3611,6 +3611,19 @@ ServerStore.prototype = {
...
+            var deps = (val['*'] && val['*'].join()) || '',
+                badre = /\b(dom-\w+|node-\w+|io-upload-iframe)/g,
+                isbad = deps.match(badre);
+
+            if (isbad) {
```
mojit0, line 3621: jslint :)

source/lib/store.server.js
```diff
@@ -3611,6 +3611,19 @@ ServerStore.prototype = {
                 }
             } // foreach mojitType
         } // foreach env
+
+        //log warning if server mojit has dom dependency
+        Y.Object.each(this._mojitYuiRequired.server, function(val, mojit) {
+            var deps = (val['*'] && val['*'].join()) || '',
+                badre = /\b(dom-\w+|node-\w+|io-upload-iframe)/g,
```
mojit0: I think checks here should be explicit module names, contained in a CONSTANT somewhere :)

isao:
Lints fine, put a space in my comment, thanks. This is a stop-gap for people experiencing start failure after upgrading to .3.27. A better fix is a try/catch around the dep instantiation and actual error handling. Is this a no-go as-is?

PR #240, opened 2012-07-01
updating jslint to 2012-05-09

FabianFrank:
updating jslint to 2012-05-09. renaming from fulljslint to jslint, because this is the new name under which jslint is being published.


FabianFrank:
I hate it when my git configs get mixed up, second time today :-( throwing this one away.

PR #241, opened 2012-07-01
updating jslint to 2012-05-09

FabianFrank:
updating jslint to 2012-05-09. renaming from fulljslint to jslint, because this is the new name under which jslint is being published.


mridgway:
I am working with @reid to make node-jslint usable as a node module: https://github.com/reid/node-jslint/pull/42.  This way we won't have to copy/paste any jslint source code into our project, we can just require the npm module.

mridgway:
Superseded by #360

Issue #237, opened 2012-06-28
update `mojito docs` to use yuidoc.js


isao:
yuidoc.js by Dav Glass is more fast and modular, and has no non-node/js deps

current python one has several external dependencies.

see also issue #84

drewfish:
+∞

FabianFrank:
+1 I will submit a pull request shortly.

mridgway:
Fixed in #243.

Issue #84, opened 2012-03-30
[bz4647788] Don't make unnecessary calls to the shell for file system commands


add0n:
The yuidoc script is bad about this.

mridgway:
Are there other places that we need to clean up other than yuidoc? If not, this can be closed since #243 is merged.

drewfish:
We spawn a java process to generate instrumented code for code coverage, in app/commands/test.js.

mridgway:
Ok that makes sense. Once Arrow supports code coverage and is open sourced we will be able drop the java code.

rwaldura:
we'll use Istanbul for code coverage

PR #243, opened 2012-07-02
Yuidocjs integration

FabianFrank:
Migrated from Python yuidoc to YUIDoc.js, also added a --server switch for the new server feature of YUIDoc.js. Also removed all async shell calls from the doc generation process and replaced them with sync syscalls.

Utilized https://github.com/ryanmcgrath/wrench-js to avoid reinventing the wheel when re-implementing recursive fs operations.

This pull request resolves issue #237 and issue #84.


source/package.json
```diff
@@ -20,7 +20,9 @@
         "express": "2.5.2",
         "mime": "1.2.4",
         "yui": "~3.5.1",
-        "yuitest": "~0.7.4"
+        "yuitest": "~0.7.4",
+        "wrench": "~1.3.9",
```
drewfish: I wonder if "wrench" and "yuidocjs" should be in "devDependencies" instead?  (That is, neither of those are needed in production.)

Hmmm... perhaps there should be a separate "mojito-dev" (or "mojito-tools") package that contains development-time things (and let "mojito" just contain the runtime parts).

drewfish: Oh yeah, for sanity please keep the dependencies sorted.

FabianFrank: Those are good points. I did not realize that the dependencies were sorted, sorry. I agree with your assessment and I will move them to devDependencies.

isao:
secondary failed (pull job 30) because wrench isn't whitelisted internally

even so :+1: :+1:

FabianFrank:
I requested wrench to be whitelisted internally.

FabianFrank:
@isao wrench has been whitelisted internally, so the secondary build should also pass now. can you please check again? thanks!

isao:
nice. mridgeway's processing PRs this sprint, assigning..

Fabian, can you give him a heads up on any license/manifest/readme changes we need as well?

FabianFrank:
I don't think we have to make changes to the README or other documentation, because the usage of mojito docs didn't change. wrench is MIT licensed, like all other npm modules that we use. We could mention it at the bottom of the README, like we mention Mustache. On the other hand we also don't mention other dependencies there, so I wasn't sure if we want this. Any thoughts?

FabianFrank:
@isao @mridgway Is there an ETA for when we can merge this one?

mridgway:
@FabianFrank I'll get this fixed up and in by the end of the week.

mridgway:
Merged. Thanks a bunch @FabianFrank.

PR #251, opened 2012-07-07
Add --print option to jslint usage message

redonkulus:
This is just a small enhancement to the mojito jslint --print option that was added recently. Adding it to usage message to make it visible to users.


isao:
:+1:

FabianFrank:
:thumbsup:

Issue #126, opened 2012-04-11
Update or remove Express


chetanankola:
I recently tried deploying my Mojito app to "Nodejitsu" and the folks at Nodejitsu told me that Express needs update since a "race condition bug for listening" (my app apparently wasnt listening right) was fixed in Express 2.5.9

So I was asked to wait till "Mojito" updates its express.

To deploy onto nodejitsu I was asked to make the following changes:

In my App's server.js I was asked to change this line:
module.exports = require('mojito').createServer();
to
module.exports = require('mojito').listen();

and in package.json
    "scripts": {
        "start": "mojito start"
    }

to 

  "scripts": {
    "start": "server.js"
  },

rwaldura:
Thanks for the report, Chetan.
Guess we'll just have to upgrade Express!

mojit0:
There are a couple of issues. Updating express is something we can schedule, but there's also some unfortunate cruft in the startup sequence that means the server.js file won't start mojito properly so changing from 'mojito start' to 'server.js' may not be workable. The good news is this is on our radar for the near-to-mid-term. We'll try to see if we can bump this up given that it's in your way tho!

Marak:
Please let me know if there is anything I can do to help in making sure mojito apps can be required and started programmatically.

A top-level method to start up a server that accepts configuration options would be ideal.

FabianFrank:
@mojit0 @Marak @rwaldura @chetanankola 

I just played with Mojito 0.3.28 and nodejitsu. After small changes (see referenced pull request) I got our weather training app working on nodejitsu:
http://fabianfrank.myweatherapp.jit.su/

You can find the source of the weather app (fixed package.json and server.js) here:
https://github.com/FabianFrank/trainingMojito/tree/master/lesson1/myWeatherApp

rwaldura:
@mojit0 is working on this, via #422.

rwaldura:
See also bz5753621 for an example of how Express is currently used -- i.e. we need to support the upload use-case even though we drop Express.

rwaldura:
We've updated Express to 2.5.10. I'll open another ticket to chuck it out entirely (i.e. drop our dependency on Express).

PR #267, opened 2012-07-14
start app as # node server.js

FabianFrank:
This change allows apps to be started using
$ sudo node server.js

Also they can now be deployed to nodejitsu. I think server.js is a better start script than "mojito start", for example the latter only works if mojito was installed globally, which might not be the case in a lot of production environments. server.js is more portable.

This fixes issue #126.

### DIFF
--- a/source/lib/archetypes/app/default/package.json.mu
+++ b/source/lib/archetypes/app/default/package.json.mu
@@ -16,7 +16,7 @@
         "mojito": ">= 0.1.0"
     },
     "engines": {
-        "node": ">= 0.4.2 < 0.5.0"
+        "node": "0.6.x"
     },
     "scripts": {
         "start": "server.js"
--- a/source/lib/archetypes/app/default/package.json.mu
+++ b/source/lib/archetypes/app/default/package.json.mu
@@ -19,6 +19,6 @@
         "node": ">= 0.4.2 < 0.5.0"
     },
     "scripts": {
-        "start": "mojito start"
+        "start": "server.js"
     }
 }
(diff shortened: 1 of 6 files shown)


FabianFrank:
I also made it work with heroku, just some minor changes:
https://github.com/FabianFrank/trainingMojito/compare/FabianFrank%3Amaster...heroku

http://mojito-demo.herokuapp.com/

caridy:
@FabianFrank this is challenging because of the variable port. In manhattan, it is manhattan the one responsible for defining the port and calling for listen on the exported server object. In dev, we let mojito to pick up the port from application.json. In nujitsu and heroku they will provide a port as well. I like your patch  to listen(process.env.PORT), maybe that's the right thing to do. I think we will need to work with manhattan to get them to use the same principle used by nujtsu and heroku.

About the appPort in application.json, I think that's just useless since you can't leverage the dimensions to have different port per env or whatever, I think we should just remove that thing and stick to the server.js definition for the listen on dev mode when process.env.PORT is not defined.

FabianFrank:
Thanks @caridy, let's follow up internally and I will provide updated code after we have a conclusion.

FabianFrank:
Since Denis and therefore Manhattan agreed to accept a server.js that calls .listen() we can change Mojito to align with what the community seems to have accepted as standard.

mojit0:
Closing in favor of a more comprehensive update to Mojito startup logic: https://github.com/yahoo/mojito/pull/422

rwaldura:
See #422.

Issue #230, opened 2012-06-25
Support for dynamic title in HTMLFrameMojit


devsathish:
Currently HTMLFrameMojit supports static title only (using application.json). Its a very basic use case and it has to supported at the platform level.
There is a workaround to set title using JS., But its not the right way since its not SEO compatible. Also, customizing HTMLFrameMojit / ShakerHTMLFrameMojit is not a good idea.

The fix is already available., Pls refer to this pull request: https://github.com/yahoo/mojito/pull/123

rwaldura:
See pull request #123.

devsathish:
Its just a change of two lines.
Replace HTMLFrameMojit:controller.server.js:76 with the following.

//Set the title (with fallback to default title)
var title = (meta.common && meta.common.title)?meta.common.title:'';
data.title = title || ac.config.get('title') || 'Powered by Mojito ' + Y.mojito.version;

I don't have the diff file. Hence giving the change in this comment.

rwaldura:
Thanks for the patch. It does appear very simple indeed. We've been wanting to address this area (frame configuration) comprehensively, but maybe it's time for a quick fix.

We'll look at this in Sprint 38 starting July 16 and ending 2 weeks hence, thanks!

haisuby:
is this patch available in the current version? I couldn't see any changes for this title issue. rwaldura@, could you please confirm the inclusion of this patch in the recent version? Thanks!

caridy:
Closing this now that we have the infrastructure to facilitate the creating of new frame mojits, which is what we want people to do.

PR #268, opened 2012-07-15
Adding meta config field for HTMLFrameMojit to create meta tags

markandey:
Adding a config attribute meta for HTMLFrameMojit 
if application.json contains

<pre>        "specs": {
            "frame": {
                "type": "HTMLFrameMojit",
                "config": {
                    "title": "Hello App",
        ------->     "meta": [{
                        "name": "hello",
                        "content": "hi"
                    }],
         ...
</pre>

<pre>
&lt;!DOCTYPE HTML&gt;
&lt;html&gt;
&lt;head&gt;
&lt;script type=&quot;text/javascript&quot;&gt;var MOJITO_INIT=new Date().getTime();&lt;/script&gt;
&lt;meta &quot;name&quot;=&quot;hello&quot; &quot;content&quot;=&quot;hi&quot; &gt;

&lt;title&gt;Hello App&lt;/title&gt;
...
</pre>

### DIFF
--- a/source/lib/app/mojits/HTMLFrameMojit/controller.server.js
+++ b/source/lib/app/mojits/HTMLFrameMojit/controller.server.js
@@ -47,7 +47,64 @@ YUI.add('HTMLFrameMojit', function(Y, NAME) {
 
         return data;
     };
+    var renderMetaTagAttributes = function(arrayMeta) {
+        //if application.json contains
+        // "specs": {
+        //     "frame": {
+        //         "type": "HTMLFrameMojit",
+        //         "config": {
+        //             "title": "Hello App",
+        //------->     "meta": [{
+        //                 "name": "hello",
+        //                 "content": "hi"
+        //             }],
+        //             ...
+        //this function returns 
+        // [
+        //     {
+        //         "attributes": [
+        //             {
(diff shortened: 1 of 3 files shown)


drewfish:
I actually prefer a different approach, where Mojito makes it easy for anyone to write their own replacement to HTMLFrameMojit.  (HTMLFrameMojit doesn't really do that much that is special.)  In Yahoo, we already have a couple of teams who've written their own custom versions.

markandey:
It's your call..
Since HTMLFrameMojit is part of framework and documented, people will use it and then they might find it hard to add a meta tags. 
I saw the HTMLFrameMojit's **template(view)  was having a way to add**, but somehow it was not implemented in **Controller**. SO I thought this could be one of the TBD task.

rwaldura:
See also issue #230.

rwaldura:
As Drew mentioned, the current HTMLFrameMojit has some limitations, which Mojito users like yourselves run into rather quickly. 

We can patch the existing mojit, but it seems like a real chance for something new and better. Would you be interested in offering your own for adoption? 

@renatoi may have some comments here, re. e.g. the MicroFrame?

PR #274, opened 2012-07-17
attempt to fix the "app" build test broken by PR 252

drewfish:


rwaldura:
See pull request #252.

PR #276, opened 2012-07-18
better version of fix for the "app" build test broken by PR 252

drewfish:
Ugg... I didn't go far enough last time :(


rwaldura:
see pull request #252

PR #285, opened 2012-07-18
Bug Fix: Use command-line context in "mojito build html5app" command

ricallinson:
This is a fix for where the command-line context was not being used by the "mojito build html5app" command.


rwaldura:
Thanks Ric. We'll try to merge in the next Sprint, starting Monday 7/23.

Issue #278, opened 2012-07-18
Affinity bug in i13n.common.js


ricallinson:
This file has "common" affinity yet depends on a "server" affinity addon. As such it fails "dependsOn" checks done in the "action-context" once deployed to a client.

The problem can be found at line 220:

I13nAddon.dependsOn = ['config', 'http', 'url'];

rwaldura:
What is the fix? What needs to be removed from the dependsOn array?

ricallinson:
Option 1:

There could be two versions of the Addon, one for the server and one for the client.

Option 2:

Remove "http" from the "dependOn" array and add it as a YUI require. However, this would always throw a warning as the http Addon will never be found by the client.

```
I13nAddon.dependsOn = ['config', 'url'];

Y.mojito.addons.ac.i13n = I13nAddon;

}, '0.1.0', {requires: [
    'mojito',
    'mojito-addon-http'
]});
```

I can submit a patch for Option 2 if required?

In either case the i13n Addon should not be part of the public facing Mojito library as it cannot be used by anyone other than Yahoo!

ricallinson:
Pull request submitted https://github.com/yahoo/mojito/pull/287

PR #287, opened 2012-07-18
Bug Fix: i13n.common addon depended on http.server addon which broke client only deployments

ricallinson:
I removed the hard dependency a replaced it with a YUI require. This will still throw a warning but will not generate an error that stops the client from executing.


rwaldura:
see #278

mridgway:
We opted to split the addon into server and client. The i13n addon has also been moved into the internal Mojito which will be released today.

Issue #273, opened 2012-07-17
Custom HTTP status message


alganet:
RFC2616 says:

```
The individual values of the numeric status codes defined for
HTTP/1.1, and an example set of corresponding Reason-Phrase's, are
presented below. The reason phrases listed here are only
recommendations -- they MAY be replaced by local equivalents without
affecting the protocol.
```

It should be possible to return something like this:

```
return ac.done({}, {http: {code: 400, status: "Malformed recipient"}, serialize: 'json'});
```

Similar functionality could be available on `ac.error()` as well.

I'm willing to send a pull request if needed. What do you think?

mojit0:
I can't speak to how we'd prioritize merging this feature, but if you have a use case requirement and would like to contribute a solution we'll definitely consider it.

rwaldura:
I'd like to understand better, it feels a little strange to me: why issue a raw HTTP message from the framework?

rwaldura:
@mojit0 Could this be done from the middleware? Should it be?

mojit0:
@rwaldura It will depend on the nature of the error/status code in question. Certain codes might be detectable as early as middleware in the pipeline, but most are far more likely to occur within the engine and well after the middleware has handed off to the Mojito core and the server-side controller logic.

alganet:
@rwaldura Our team is developing an application that also behaves like an API when needed. There is a mojitProxy-less mojit that responds to raw RESTful requests that must follow a strict contract.

One of our major requirements is to have a really small payload and fast network performance, and reusing the status message seems like a good approach, even compatible with AJAX clients via `xhr.statusText`.

Our current implementation uses the default "Warning" HTTP header for this, but it isn't as canonical as customizing the status message, since the "Warning" header has been used for expressing errors in the entity body, not related to the request itself. 

Tunneling the errors through the entity body itself and sending a generic HTTP status is one of our possible choices, but this generates overhead in both clients and proxies trying to calculate the Content-Length of the response.

FabianFrank:
@alganet I think this absolutely makes sense, when providing APIs.

I have one more use case and I think it answers the why question of @rwaldura :
It is a huge help to not only see a generic error code but a specific description of the real error from the perspective of an API customer. A lot of good Yahoo APIs even give you a URL in the description where you can read more. I think Mojito should support this feature, because it can make APIs implemented in Mojito a lot better.

rwaldura:
[...]

PR #300, opened 2012-07-22
Error http reasonphrase

FabianFrank:
Providing the ability to provide a reasonPhrase property when calling ac.error. Mentioning this in the API doc and also adding a UT. This addresses issue #273.

### DIFF
--- a/source/lib/app/autoload/action-context.common.js
+++ b/source/lib/app/autoload/action-context.common.js
@@ -47,7 +47,11 @@ YUI.add('mojito-action-context', function(Y, NAME) {
      * @param {Error} err A normal JavaScript Error object is expected, but you
      *     may add a "code" property to the error if you want the framework to
      *     report a certain HTTP status code for the error. For example, if the
-     *     status code is 404, Mojito will generate a 404 page.
+     *     status code is 404, Mojito will generate a 404 page. Additionally you
+     *     might provide a reasonPhrase property, to override the default humran
+     *     readable description for this status code with one specific to your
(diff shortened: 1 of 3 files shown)


isao:
Thanks Fabian. I'd love to accept this if only for the unit tests, and it passes secondary (pull job 29) :clap:

Since ac.error is a [standard Error object](https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error) is there a reason you use "reasonPhrase" instead of "message"? Just wondering.

FabianFrank:
I took the term reasonPhrase from the HTTP documentation: http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html. Node itself also decided to re-use this term.

isao:
sounds good to me, thanks again Fabian.

:+1:

alganet:
Thank you very much Fabian! That will help my team a lot =D

ooskapenaar:
SEO and website marketing requires not only a message as in this PR, but fully template styled HTML 404 and 500 pages.

Should this functionality be implemented in the middleware (express) or in the mojito layer, e.g. output-handler.server.js?

ooskapenaar:
I am patching the error function in _ac-context.common.js_ and _output-handler.server.js_ to add an html parameter in a backward compatible manner:

``` javascript
-     error: function(err) {
+     error: function(err,html) { 
```

``` javascript
-         var out = '<html><body><h1>Error: ' + err.code + '</h1>' +
+        var out = html || '<html>' +
+                '<body><h1>Error: ' + err.code + '</h1>' +
```

PR #306, opened 2012-07-24
add run script for CI

lzhan:


FabianFrank:
Hi Lichun,

I don't think we should be adding shell scripts to the Mojito codebase. Something like
$ mojito arrow -type [unit|functional] [-seleniumHost HOSTNAME] ...
makes more sense, because it will scale much better than a collection of scripts. The config should not always have to be on the command line, but also be storable in a config file, such as arrow.json. It should be similar to how mojito docs and mojito test works, so that you can run arrow tests using the mojito command for the framework, a full app or a single mojit. In order for more Cocktails projects to adopt arrow, we should integrate it with mojito. Otherwise it is less appealing than mojito test, which showed th way to go with respect to how tests are an integral part of the framework and the apps we develop.

Let me know if you need help implementing a new command mojito arrow, I have already worked on that part of he codebase.

Fabian

isao:
Note, this PR as it's non-functional right now.. Background: it's a helper script for jenkins only. 

Fabian, we're not sure about making arrow/selenium/phantomjs a mojito devDependency or not, but I like your implication of adding it as an optional module mojito-cli-cmd-arrow style. Comments/input welcome

PR #309, opened 2012-07-26
rm local copy of mime, use npm mime module instead

isao:
tested locally: unit tests, newsboxes


drewfish:
yay!

isao:
inspired by mojit0's yuitest/javascript deletion... next up yuidocjs hopefully

isao:
actually, it'd be nice to npm module-ize much of the internals... nice comment by Isaac Schlueter in his [npm-www Design Philosophy](https://github.com/isaacs/npm-www)

```
No lib folder
If you would put it in lib/, then it belongs in a separate module.
```

mojit0:
Yeah, I like the kind of defacto standard that' arisen around 'deps' for JS files that aren't npm modules but dependencies, lib for only your own stuff, and everything else as a node module.

drewfish:
Does the NPM community have a standard on how to ship NPM modules to the browser?  The special sauce of Mojito is that much of the framework and application code runs in both the server and client.  We currently do that using YUI modules, but yeah, if there's an NPM standard for that, that would be a more broadly used/understood standard and would be a better choice just because of that.

isao:
not a standard afaik, but there's @substack's [browserify](https://github.com/substack/node-browserify) which looks really neat, and some related projects. prob more. and reportedly some folks actually use npm for hosting browser-only code/assets. 

but we have YUI, Loader, and Gallery, so I don't think we can cross the streams efficiently, so to speak. browserify is for node.js code in the browser, and yui/mojito is kind of.. the other way around...

PR #311, opened 2012-07-26
enhance README for new users

FabianFrank:
remove application.json step that's not needed anymore
raise awareness of UT and docs feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,19 +28,6 @@ Create a mojit:
 
     $ mojito create mojit HelloMojit
 
-Edit application.json to configure mojit so that it looks like:
-
-    [
-        {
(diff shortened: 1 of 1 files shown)


isao:
:+1: zhouyaoji?

zhouyaoji:
Great: the simpler, the better. The README will also contrast more with the quickstart tutorial, so the latter, which has a step for editing 'application.json', becomes more important.

zhouyaoji:
Correction, the quickstart has a step for editing the controller. The getting started tutorial has a step where the user creates an instance of a mojit in application.json. In any case, the progression from README to quickstart to getting started is nicely paced.

PR #317, opened 2012-07-27
[Fix 5712477] Make sure to callback when there is an error

add0n:
Fixed to make callback fire when there is an error (kudos to Scott Shattuck for his assistance).


drewfish:
+1

PR #318, opened 2012-07-27
Ported composite addon tests

mridgway:


tests/base/mojito-test.js
```diff
@@ -28,6 +28,20 @@ YUI.add('mojito-meta-addon', function(Y, NAME) {
 
 /*
  */
+YUI.add('mojito-composite-addon', function(Y, NAME) {
+    Y.namespace('mojito.addons.ac');
+    return function() {};
```
drewfish, line 33: I've never seen a YUI.add() that had a return statement before.  What does that do?

mridgway, line 33: Not sure, it was copy/pasted from @mojit0's examples.  I don't think it does anything to be honest.

FabianFrank:
How can those tests be run? Using mojito test?

mridgway:
Eventually yes, but not right now. We're gearing up for the new testing framework.

mojit0:
New tests, and existing tests, are being moved to Arrow in anticipation of open sourcing that framework. At present all that work is being done in a separate branch with anticipation of merging it into develop when the tests are ported, Arrow is open sourced, and we've completed work to augment mojito test with a way to run both the tests existing customers have written and the new framework tests being written for Arrow.

mojit0:
New tests, and existing tests, are being moved to Arrow in anticipation of open sourcing that framework. At present all that work is being done in a separate branch with anticipation of merging it into develop when the tests are ported, Arrow is open sourced, and we've completed work to augment mojito test with a way to run both the tests existing customers have written and the new framework tests being written for Arrow.

ss

PR #325, opened 2012-07-29
wip assets addon unit test port

isao:


mojit0:
Gotta work on those syncs :)

isao:
this is off `git checkout -b arrowd17c56aa up/arrow` and cherry-picking 17c56aa... I'm surprised. But I think we can just squash this entire branch onto develop.

mridgway:
This should be merged into the arrow branch instead of develop right?

isao:
@mridgway ah. you're right. I think I did this on my phone. closing!

Issue #109, opened 2012-03-30
Refactor certain complicated parts of the store


add0n:
For readability.

Issue #95, opened 2012-03-30
Proper Mojito version number comparison


add0n:
It could be one of two things:

have resource store return the correct Mojito version, which was not just hard-coded in a source file somewhere
or

Within a mojit package.json, there should be a dependency declared on the version of Mojito it can run within like ">0.1.0.96"

Issue #86, opened 2012-03-30
[bz4647471] Real device detection


add0n:
As a part of this, allow a configuration to skip device detection altogether for performance reasons.

Issue #59, opened 2012-03-30
test files under addons/ac/ are loaded when mojito starts


add0n:
Files using the suffix "-tests" must not be loaded when you start your mojito application. That's not the case for files under addons/ac/

drewfish:
This looks to be a duplicate of issue #50.

PR #326, opened 2012-07-30
redesigned resource store

drewfish:
Herein lies the new resource store.
## backward-compatibility changes
- Users will now need to configure the iphone selector by hand (`{"settings": ["device:iphone"], "selector":"iphone"}`) in their `application.json` files.  (This resource store no longer assumes the selector is just the device.)
- In a mojit's `package.json`: the version check is now spelled `engines.mojito` instead of `yahoo.mojito.version`.
- In a mojit's `package.json`: `config.mojito` is now spelled `yahoo.mojito`.
- The second argument to `serializeClientStore()` is gone.  (This was pretty much internal to mojito anyway, so likely no one will miss it.)
## issues fixed
- issue #10: Ensure Mojito can handle YUI module names with underscores
- issue #32: [mojito compile rollups] using dot notation in the name of the css files will make compile to discard some of them
- issue #50: mojito start loads tests from addons
- issue #54: "mojito compile inlinecss" has unexpected behavior when there is a css file name that contains the string "iphone"
- issue #59, bz 5380714: test files under addons/ac/ are loaded when mojito starts
- issue #95, bz 4647440: Proper Mojito version number comparison
- issue #109: Refactor certain complicated parts of the store
- issue #110: Detect duplicate resource ids on cookdown
- bz 4606483: Mojit specs can be within a spec.json file so they can be YCBed
- bz 4647169: considering making a utility YUI object in ResourceStore
- bz 5362206: Need ability to select controller, view, model and binder based on context
- bz 5495519: mojito is not loading - stuck retrieving all the datetype-date-format for each locale
- bz 5530096: Mojito tries to parse .json~ files in [mojit]/specs/
- bz 5541500: mojito not doing fallback
- bz 5609501: L10n: Root strings file is not used by default when a non-root strings file is present 
## issues unblocked
- issue #37: i18n support for html5apps
- issue #78: Complete client resource store implementation
- issue #86: Real device detection
- bz 4805839: Custom archetypes
- bz 5450431: [json files granularity] how to include or synthetically define values for dimensions
- bz 5605173: {selector} is not defined or explained in the documentation 

This passes yahoo-internal fork build 54, with the exception of one test.  That test is broken, and should be updated once this is merged to develop.  (The test is using `mojito start --context lang=de` which is used to set the default context.  However, every request gets the lang set in the context, so that will override the default context specified on the command line.)


drewfish:
Can't cleanly merge.  I'll cut a new PR when that's available.

PR #269, opened 2012-07-16
Added compatibility with Y.Model

focuzz8:
Related bug https://github.com/yahoo/mojito/issues/259

Issue #259, opened 2012-07-10
[v1 B-55829] Mojito "model" should converge to Y.Model


drewfish:
Mojito's idea of a model should align closely with YAF's idea of a model .  This might even go so far as to throw away Mojito's models and use Y.Model instead.

(The goal is that an app built using YAF can migrate easily/naturally into a mojito app.)

trungpham:
Yay. I never understood the idea of the current Mojito Model.

It cannot be instantiated. It does not hold any data. It looks like a group of static functions?
Why was it called a model to begin with? :)

focuzz8:
As I see, models as they exist, can be usefulfor example as gateways for external api (xml or any other - not just rest) etc. So they are useful, but are they "data models", probably "service models"?
Just my thought...

trungpham:
Exactly. They are just service model. They hold no data.

The trick here to somehow migrate the data from the node to the browser
after node populated the data.
On Jul 14, 2012 2:40 PM, "Alexey Shafranovich" <
reply@reply.github.com>
wrote:

focuzz8:
Hope this helps https://github.com/yahoo/mojito/pull/269
This just enables `Y.Model` to be passed into `ac.models[NAME]` this way:
`Y.mojito.models[NAME] = Y.Base.create('pieModel', Y.Model, [], { }, { });`

rwaldura:
See also forum discussion at 
http://developer.yahoo.com/forum/Yahoo-Mojito/Integrating-with-YUI-3-5-MVC-App/1339551100881-219004e1-417c-4bd9-bf46-4c5d7ee7c1cc

focuzz8:
Well, I can say that transparent routing support on server and client can be right thing. It can be implemented manually using Y.Router (with manual client-side route-to-action mapping) but it can be perfect to share routes.json config on client and server, but it requires to spend some time to be done.
About dynamic mojit loading - I've created "placeholder" - generic mojit based on code of LazyLoad, that reacts on messages(events) addressed to him, and reloads it's content (recreates inner mojit). Can share some thought about it.
And I think, good idea is to give ability (I don't tried yet) for binders, to access models. It can be useful for form validation for example.
I think there is no need of Y.View on client - controller already does everything.
Just my thought and my vision.

Satyam:
Please, don't add Y.Model to Mojito as it is.   It has several architectural shortcomings and it would be a bad thing if those propagated to Mojito.  
[...]

PR #337, opened 2012-08-01
PR #269 Added compatibility with Y.Model

drewfish:
This is PR #269 updated with a test and delinted.


drewfish:
As mentioned in the original PR, this is related to bug #259.

mridgway:
From what I understand, this patch makes a Y.Model available in a controller so that you could instantiate one or more instances of it. These models would then be passed to the view to be rendered. But the view does not have a way of access the attributes of the model because it should be using model.get('foo').

I think we need to discuss more in depth what integrating YAF components means for Mojito. We should come up with some functional use cases of how we would use the YAF components and work from there, instead of throwing the components in and sorting them out afterwards.

caridy:
@mridgway until we get a better integration, the idea is to do something like this:

```
index: function(ac) {
    var foo = new ac.models.ModelFoo({/* model config */});
    foo.load(function() {
        ac.done(foo.toJSON());
    });
}
```

Although, I don't see the point of having the models collection, I rather prefer to do something like:

```
index: function(ac) {
    var foo = new Y.ModelFoo({/* model config */});
    foo.load(function() {
        ac.done(foo.toJSON());
    });
}
```

and defining the proper requirement as part of the controller's requires config, since that's more natural for YUI users, and ultimately, the optimization of what needs to be loaded (in the client side) in case you have more than one model under a mojit.

and eventually, once we get better integration with Y.View/Y.Renderer we should be able to do something more advanced, in this case passing the reference of the model into the view/renderer, just like YAF.

/cc @drewfish

drewfish:
@mridgway I think you have a very good point.  We should have a complete answer for the integration before we take any steps (so that we take the right steps).

The assumption is that Y.Models are also useful for the controller.  However, that's very much from the "classic mojito" point of view, and doesn't take into account things like you've mentioned.

mridgway:
I agree having the `ac.models` collection doesn't seem necessary.

This implementation at least gets users started with Y.Model, so I suppose that's a good thing. In the future if it's just a matter of changing `ac.done(model.toJSON())` to `ac.done(model)` then we're making progress towards our goal (which I think we should still have a solid plan for).

drewfish:
Closing this, since the plan is to merge this work into the yaf-develop branch, which is PR #370.

PR #341, opened 2012-08-02
Fixed descriptors to work for server side tests using nodejs driver

mridgway:
Also, cookie.server.js tests ported.

This commit will allow us to run both server and client tests (previously running server tests didn't include the mojito-test.js file, so it failed).

**Important**: This can only be pulled once arrow includes commonlib files before test lib files. I have submitted a patch so I'm hoping it can be pulled in soon. Otherwise we are blocked on getting server side tests running with arrow.


mridgway:
The patch for arrow has been applied and will be available in the next release in the next couple of days.

mridgway:
Remaking to pull into develop.

PR #346, opened 2012-08-04
arrow tests wip

isao:


source/lib/app/autoload/util.common.js
```diff
@@ -50,6 +50,7 @@ YUI.add('mojito-util', function(Y) {
 
         array: {
 
+            // wtf: this appears to be unused and nonfunctional
```
mridgway: Can we make this a TODO line instead of a WTF line? =P

tests/unit/lib/app/autoload/test_descriptor.json
```diff
@@ -1,48 +1,64 @@
...
         }
     },
     {
-        "settings": [ "environment:development" ]
+        "settings": [ "environment:development" ],
+        "name": "mojito-nodejs-unit-tests--is-not-hooked-up-yet",
+        "dataprovider": {
+            "deploy.server": {
+                "params": {
+                    "lib": "$$config.lib$$/app/addons/ac/deploy.server.js,../../../../../../source/lib/app/autoload/loader.common.js,../../../../../../source/lib/app/addons/ac/http.server.js",
```
mridgway, line 57: The second library here should be able to use $$config.lib$$

mridgway:
Simple comments on this, though nothing blocking. Looks good: :shipit:

tests/unit/lib/app/autoload/test_descriptor.json
isao, line 57: you'd think. didn't work for me.. does it work for you?

source/lib/app/autoload/util.common.js
isao: oh god how did that get there I am not good with computer
:) will fix

tests/unit/lib/app/autoload/test_descriptor.json
drewfish, line 57: Long-standing bug in mojito's javascript-only YCB where it'll only make one substitution in a value.

PR #348, opened 2012-08-06
Develop

jintao:
Add proxy server before Mojito(express), then proxy launch Mojito server and also start monitor File System code change. Detect code change will restart mojito. 
Client side (current only HTMLFrameMojit) will start web socket connection with Proxy, if Mojito server restart , will notice client to reload page.

This change introduced the live edit feature of Meteor(Most code are from Meteor). Can on going to integrate Meteor DB sync through web socket connection, and start local DB server behind Proxy server.

### DIFF
--- a/source/lib/app/mojits/HTMLFrameMojit/views/index.iphone.mu.html
+++ b/source/lib/app/mojits/HTMLFrameMojit/views/index.iphone.mu.html
@@ -2,6 +2,7 @@
 <html>
 <head>
 <script type="text/javascript">var MOJITO_INIT=new Date().getTime();</script>
+<script src="http://cdn.sockjs.org/sockjs-0.3.1.min.js"></script>
 {{#meta}}
 <meta name="{{name}}" content="{{content}}">
 {{/meta}}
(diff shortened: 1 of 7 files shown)


caridy:
hey @jintao, what's the intent of this new feature? It looks like you're targeting development, to be able to see live changes whilte developing on top of mojito, but I see a lot of production pieces changed as well (like the default views for HTMLFrameMojit).

jintao:
Yes, Just want to support live changes while developing mojito (the same as meteor), Actually I have not understand  resource management in mojito , So I do not know how to auto add client side support scripts during launch mojito, So I put client side support scripts in to HTMLFrameMojit defult view... 

I will continue to understand how to more auto add client side scripts, But if you can help, that will be good.

rwaldura:
@jintao Maybe your PR isn't bit-perfect as-is, but I think it's awesome!
Thanks very much for this. I'm going to sit on it for a bit, and schedule for Q4, okay? There are other things in the same vein that we want to do.
Thanks again!

PR #350, opened 2012-08-06
Loader to support custom yui configs

caridy:
when using the loader at the server side to generate combo urls we should pipe the yui config into the loader settings otherwise some settings will not be honored.


isao:
:+1:

mridgway:
This looks good to me.

drewfish:
+1

PR #354, opened 2012-08-06
yui-lang modules error when pkg is bundle

diervo:
When you define your package as a "bundle", 
the mojitType value is "shared" for the yui-lang modules, 
which makes mojito raise a invalid YUI lang error since basename !== mojitType

I just add a check so when the mojitType is shared, 
we accept as well the lang module as valid.

I think it shouldnt have any imlpications further, and al the tests pass perfectly.

### DIFF
--- a/source/lib/app/addons/rs/yui.server.js
+++ b/source/lib/app/addons/rs/yui.server.js
@@ -188,7 +188,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                 if (!res.yui) {
                     res.yui = {};
                 }
-                if (fs.basename === mojitType) {
+                if (fs.basename === mojitType || mojitType === 'shared') {
                     res.yui.lang = '';
                 } else if (mojitType === fs.basename.substr(0, mojitType.length)) {
                     res.yui.lang = fs.basename.substr(mojitType.length + 1);


drewfish:
-1.  The diff confuses me, since it seems to say that `mojits/Foo/lang/Foo_en.js` **isn't** for language `en`.  That doesn't seem right.

drewfish:
I think the heart of the problem is that Mojito doesn't (currently) allow shared mojits to define language bundles.  Perhaps that is what should change.

isao:
@drewfish @dferreiroval what should we do here? it's a one-liner

my opinion- shared mojits as they exist now should be deprecated, but I might be convinced until more sensible scoping changes that are planned land

diervo:
Agree, doesnt make too much sense as far as I understand...It was a conception mistake for our part. Deprecate it.

isao:
Actually I think it's a good/thoughtful patch. I should have not opined, especially if it solved a real problem for you. But if not let's close.. Thanks @dferreiroval

source/lib/app/addons/rs/yui.server.js
```diff
@@ -188,7 +188,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                 if (!res.yui) {
                     res.yui = {};
                 }
-                if (fs.basename === mojitType) {
+                if (fs.basename === mojitType || mojitType === 'shared') {
```
drewfish, line 191: Note for posterity:  this probably doesn't do what was intended.  This is saying that _all_ lang bundles found in a shared mojit are really the "default" language and not the language they otherwise specify.  (For example, it's saying that lang/Foo_de.js in a shared mojit is really english, not german.)

PR #356, opened 2012-08-06
Add Handlebars as an optional view engine

mridgway:
This set of commits makes Handlebars available for use on both the server side and client side.

A subsequent pull request will make Handlebars the default. Another subsequent PR will make the Handlebars view engine render "mu.html" templates.


source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,104 @@
...
+                if (obj) {
+                    adapter.flush(obj.template(data), meta);
+                    Y.log('render complete for view "' +
+                        tmpl + '"',
+                        'mojito', 'qeperf');
+                }
+                // TODO: what should we do when the template fails to load?
+                //       should we just flush an empty string? the error message
+                //       was already reported by the getTemplateObj method for sure.
+                adapter.done('', meta);
```
drewfish, line 48: It might be better if the handler is `function(err, obj)` and then it can call `adapter.error(err)` instead.

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,104 @@
...
+
+/*jslint anon:true, sloppy:true, nomen:true*/
+/*global YUI*/
+
+YUI.add('mojito-hb', function(Y, NAME) {
+
+    'use strict';
+
+    var HB = Y.Handlebars,
+        cache = Y.namespace('Env.Handlebars');
```
drewfish: Why this namespace?  Are we using it in a similar way that others might use it?  Might it be better to use a "mojito" namespace to avoid collision?

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,104 @@
...
+            this._getTemplateObj(tmpl, !meta.view.cacheTemplates, handler);
+        },
+
+        /**
+         * Cache the reference to a compiled handlebar template, plus
+         * a raw string representation of the template.
+         * @private
+         * @param {string} tmpl The name of the template to render.
+         * @param {boolean} bypassCache Whether or not we should rely on the cached content.
+         * @param {function} callback The function that is called with the compiled template
```
drewfish, line 60: Might be nice to document the callback parameters, such as `@param {function(obj)} callback`.

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,104 @@
...
+         * @return {object} literal object with the "raw" and "template" references.
+         */
+        _getTemplateObj: function (tmpl, bypassCache, callback) {
+            var handler = function (templateStr) {
+                if (templateStr) {
+                    // applying a very simple local cache to avoid reading from disk over and over again.
```
drewfish: Hmmm.... this file is .client.js, so talk about "reading from disk" is a little confusing... :)

source/lib/app/addons/view-engines/hb.server.js
```diff
@@ -0,0 +1,116 @@
...
+        },
+
+        /**
+         * Loads a template from the file system
+         * @param tmpl The location of the template file
+         * @return {string} The template string
+         * @private
+         */
+        _loadTemplate: function (tmpl) {
+            return fs.readFileSync(tmpl, 'utf8');
```
drewfish: If possible, we should avoid the readFileSync().  Since view engines are run during request time, blocking here means that we block handling of other requests while we're waiting for the template to load from disk.

source/lib/app/addons/view-engines/hb.server.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+    };
+
+    Y.namespace('mojito.addons.viewEngines').hb = HandleBarsAdapter;
+
+}, '0.1.0', {requires: []});
```
drewfish: Why not just requires:['handlebars'] and then pull var HB from Y.Handlebars?

drewfish:
I can imagine both `.client.` and `.server.` versions being exactly the same, except for the implementation of the `_loadTemplate()` method.  (That should have the same signature in both versions.)

(Hmm... and come to think of it, ideally `_loadTemplate()` would be some implemented in a utility lib somewhere, since I can image that _every_ view engine would need to have something like `_loadTemplate()` with separate implementations for client and server.)

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,104 @@
...
+         * Loads a template from a remote location
+         * @param tmpl The location of the template
+         * @param cb The callback to call with the template response
+         * @private
+         */
+        _loadTemplate: function (tmpl, cb) {
+            Y.log('Loading template from server: ' + tmpl, 'mojito', 'qeperf');
+            Y.io(tmpl, {
+                on: {
+                    complete: function (id, resp) {
```
drewfish: I think some kind of error handling is needed here, if the server times out or such.  (Ideally an error should propagate up to `cb`, so that it can figure out how to respond.)

tests/unit/lib/app/addons/view-engines/test-hb.client.js
```diff
@@ -0,0 +1,86 @@
...
+        var suite = new Y.Test.Suite("mojito-hb client tests"),
+            TEMPLATES = {
+                'oldObjNotation.hb.html': '<div>{{#tester}}{{test}}{{/tester}}</div>',
+                'dotNotation.hb.html': '<div>{{tester.test}}</div>'
+            };
+
+        suite.add(new Y.Test.Case({
+            setUp: function () {
+                this.viewEngine = new Y.mojito.addons.viewEngines.hb();
+                this.viewEngine._loadTemplate = function (tmpl, cb) {
```
drewfish, line 29: nice

drewfish:
Oh, and if we can make is so they're the same except for the implementation of `_loadTemplate()`, then the two test files can just be a single one with `.common.` affinity.

source/lib/app/addons/view-engines/hb.client.js
mridgway, line 48: Changed this. adapter.error will now be called with a non-200 status code on IO call.

source/lib/app/addons/view-engines/hb.client.js
mridgway: Yeah, this was from @caridy. I'll defer to him.

source/lib/app/addons/view-engines/hb.client.js
mridgway: Removed this comment. Was a remnant form a copy/paste =P

source/lib/app/addons/view-engines/hb.server.js
mridgway: So I discussed this with @caridy as well. It's a double-edged sword: if you use asynch then you may have multiple hits that have to load from the filesystem while the first is still reading but if you use synch you block other requests. I would prefer asynch as well, but this will then require an API change for `compiler` and `precompile` functions as they'll need a callback.

source/lib/app/addons/view-engines/hb.server.js
mridgway: Fixed, good catch.

source/lib/app/addons/view-engines/hb.client.js
mridgway: Fixed as per previous comment.

source/lib/app/addons/view-engines/hb.client.js
mridgway, line 60: Not sure if there is a standard way of documenting this. I'll look into this.

mridgway:
Indeed. It almost seems like an abstraction for a template-loader would be good to have, and then the view-engines are just dealing with how to render and compile strings. Perhaps a future refactor?

source/lib/app/addons/view-engines/hb.server.js
drewfish: Yes, that's the problem with async, it's a virus :smile: -- if a low-level function is async that very often causes higher-level functions to need to be async as well.

I think the use of `readFileSync()` in a codepath that happens during runtime might have significant performance implications, especially with higher levels of concurrency.  Yeah, I do think that means that `compiler()` and `precompile()` should by async.  (And yeah, that's outside the scope of this pull request.)

I'm not especially worried about multiple reads of the same file happening at the same time.  Yeah, we should probably optimize this, but also very often the operating system filesystem has caching built in, and might also be doing collapsing of multiple calls.  (Similarly, we should be doing same-request collapsing in .client. so that we're not making multiple HTTP requests for the same template at the same time.  It's more important here in fact, because the Y.io library is very likely _not_ doing same-request collapsing.)

drewfish: (As an aside, why do we have both `compiler()` and `precompile()`?  They appear to be doing the same thing!)

drewfish:
Yeah, I think the template-loader should be a future refactor.

I'm going back-and-forth whether `render()` and `compiler()` (and `precompile()`) should just be passed the template contents, of whether it should use a "utils" library for loading the template.

At this point, I favor the latter, because:  (a) it's a generic "loadFile()" method that does the right thing for each environment, whether passed a filesystem path or a URL, and (b) future rendering engines might support partials/includes, which means that they'll have the file paths to load besides the root template.

source/lib/app/addons/view-engines/hb.client.js
caridy: Old habits are hard to change, let's use mojito namespace instead of YUI namespace if that's more convenient.

The main purpose of the cache is to share compiled views across multiple Y instances (mojit instances). I wasn't sure if Y.mojito.\* was shared across all Y instances, I'm still not sure!

What the best way to share data across Y instances under mojito namespace then? something equivalent to Y.Env === YUI.Env

source/lib/app/addons/view-engines/hb.client.js
caridy, line 60: Here is an example of how to document callbacks for the new yuidocs:

https://github.com/yui/yui3/blob/master/src/app/js/model.js#L375

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,115 @@
...
+                    return;
+                }
+
+                if (obj) {
+                    adapter.flush(obj.template(data), meta);
+                    Y.log('render complete for view "' +
+                        tmpl + '"',
+                        'mojito', 'qeperf');
+                }
+                // TODO: what should we do when the template fails to load?
```
drewfish: I think this comment can go away now that we have the `adapter.error()` call.

source/lib/app/addons/view-engines/hb.client.js
drewfish: Hmm... I guess I'm suggesting something like `Y.namespace('Env.Mojito.Handlebars')` then.  The "Env.Handlebars" seems very generic, and somebody else might be using that as well, causing us to trip each other up.

source/lib/app/addons/view-engines/hb.server.js
caridy: Let's just keep in mind that the number of views is a finite number, and they will be cache after the first read per view. Also, we can easily change all this to async since the "render" method is async by nature (adapter.done or adapter.error are the callbacks). 

Compile and precompile are meant to be used as part of the build process IMHO.

source/lib/app/addons/view-engines/hb.server.js
caridy: In my original code I was using:

HB = require('yui/handlebars').Handlebars

instead of the regular "requires" statement. This is the way @ericf is using it in photonearme project. I wasn't sure about the implications. I think this approach will isolate HB into its own Y instance, but I remember having some problems with the way HB.compile works when using the "requires". But it if works fine, then we can keep using the require :)

caridy:
@mridgway @drewfish I think that makes a lot of sense. There are a couple of things though:
1. a template engine can pre-compile views as part of the build process, in which case those views will be available in the client side as a YUI module (probably shaker can take care of this)
2. a template engine can go a get a *.html from the server on demand, and compile it on the fly.

I think _loadTemplate abstraction can take care of the hazard of resolving the template on the client side from a YUI module or by loading it over the network, and calling back with the content or object representation in case of a precompiled version as a YUI module.

This will close the loop IMHO, since people will not need to deal with the hazard of resolving the templates and partials, just the manipulation of the string or precompiled-representation of the string.

source/lib/app/addons/view-engines/hb.client.js
mridgway: Changed it to `Y.namespace('Env.Mojito.Handlebars')`

source/lib/app/addons/view-engines/hb.client.js
mridgway, line 60: Updated.

source/lib/app/addons/view-engines/hb.server.js
mridgway: Now that I think about it, I think the `require('yui/handlebars').Handlebars` is the more standard way of using YUI server side according to the docs. I'm not sure what the difference is though.

source/lib/app/addons/view-engines/hb.client.js
mridgway: Removed

source/lib/app/addons/view-engines/hb.server.js
ericf: When you `require('yui/[something]')` in moduleA.js and in moduleB.js, it uses the same YUI instance under the hood, and therefore you get back the same `Y.Handlebars` object.

These two ways use a shared YUI instance:

``` javascript
var Y1 = require('yui/handlebars'),
    Y2 = require('yui').use('handlebars');

Y1 === Y2 // => true
```

If you want to create a separate YUI instance, you can do the following:

``` javascript
var YUI = require('yui').YUI,
    Y   = YUI({useSync: true}).use('handlebars');

// Notice the `useSync`, that's important. The way of using a shared instance 
// already does that.
```

There's some more details on our [YUI on Node.js](http://yuilibrary.com/yui/docs/yui/nodejs.html) user guide.

source/lib/app/addons/view-engines/hb.client.js
```diff
@@ -0,0 +1,117 @@
...
+        /**
+         * Renders the handlebars template using the data provided.
+         * @param {object} data The data to render.
+         * @param {string} mojitType The name of the mojit type.
+         * @param {string} tmpl The name of the template to render.
+         * @param {object} adapter The output adapter to use.
+         * @param {object} meta Optional metadata.
+         * @param {boolean} more Whether there will be more content later.
+         */
+        render: function (data, mojitType, tmpl, adapter, meta, more) {
```
ericf: The `more` really smells like API creep on this adapter. Are you guys locked into this `render()` interface?

ericf:
I feel strongly that we should avoid compiling templates on the client, simply because the Handlebars compiler is so large:
- `handlebars-base`: **3KB**
- `handlebars`: **32KB**

It still seems worth being able to load the templates on demand, but instead of shipping the raw templates over the wire, we could instead ship dynamically generated YUI modules which contain the pre-compiled Handlebars templates. This way templates can become module dependencies, be loaded via YUI's Loader, and combo-loaded when more than one template is being requested.

This is the approach I've been using when sharing templates between the server and client.

source/lib/app/addons/view-engines/hb.client.js
drewfish: It might well be.  (The `more` relates to streaming... indicating whether there is more data that needs to be rendered in the template (true) or whether this is the last time data will be pushed through the template (false).)

Regardless, changing the `render()` API is outside the scope of this pull request.

source/lib/app/addons/view-engines/hb.server.js
drewfish: @ericf Do you see a downside to just using requires:['handlebars'] ?  I'd like to keep things as simple as possible.

drewfish:
Yeah, that's what the `mojito compile views` command does.  It's up to the user to decide which approach they wish to take.

caridy:
Hmm, I agree with both of you, but this is conflicting:

a) if we use the same .common.js script for HB engine, it will have to require "handlebars", and it will load that in the client side (32k). Obviously this is pretty bad. This defeats the original purpose of reusing the same implementation. I believe keeping them separate is not a big deal, and we will have more control for optimizations.

b) if a template is required on-demand, and we load it over the wire as a regular html (raw template) we will have to require "handlebars", which again introduce the overhead of loading 32k (even if we require it on demand when needed). This kinda push us to my initial proposal: templates should "always" travel as compiled in a form of a yui module, and in dev-mode mojito (and _loadTemplate routine for the matter) should take care of this process, and mojito compile views (or shaker for the matter) should take care of the optimization based on "requires" statements.

c) when do we really need the raw version to travel over the network (or from local filesystem)? maybe when generating a html5app, but in that case we can also compile them as YUI modules as part of the build process. I'm not so sure we should allow the raw to travel over the network. (they are smaller in a raw format though).

mridgway:
So this pull request has spiralled into a very large set of features and refactors. All of this is definitely the direction we want to head, but it is made more difficult by the fact that we have to still support Mu (not my choice unfortunately). I propose that we get we get Handlebars into the framework so that it is available to use and is the default. Once we are able to remove the Mu engine, we will be able to refactor components that deal with fetching, compilation, and rendering of templates, such that we can make the above optimizations.

drewfish:
I agree.  This PR has uncovered a lot of good ideas, but they shouldn't/can't be implemented as part of this pull request.

caridy:
:+1:

source/lib/app/addons/view-engines/hb.server.js
ericf: @drewfish The things to consider are how where templates, partials, and helpers will be cached, and what type of sandboxing you'll want with these things.

If you use a standard YUI module dependency `require: ['handlebars']`, then the `Y.Handlebars` object and any templates, partials, or helpers that have been registered to it, will only be available in that sandbox (I think, this should be tested because Handlebars is an odd duck since it's an embedded project in YUI.) This is what I'd suggest doing as a starting point. If you need more flexibility and control you might like to use Node.js-style: `require('yui/handlebars')`.

caridy: The rational that I used in the initial code was: "whenever a template is cache, it uses the fullpath as a hash, which means it there is not need to have a sandbox per mojit.", and that works well with require('yui/handlebars').

Now, I don't know if this works nicely with the new RS (I guess it behaves the same), and how are we going to support partials. I think we will have to put more thoughts on the partial resolution in the short term, in which case, having a sandbox per mojit might work better to register partials per mojit. This doesn't mean we can't have a global hash for cache entries, but we will need a way to register those partials somehow.

Another question for @drewfish, are we going to support global views? and global partials? If yes, then how they work in the context of a mojit?

PR #358, opened 2012-08-07
Store client URL prefix fix

ricallinson:
The store.client.js code builds a URL with which to call the server for either a "Spec" or "Type" JSON file. The URL it generates expects the value of "this.staticPrefix" to start with a slash ("/"), however it doesn't always have the slash. The result is an invalid URL which crashes the app in browsers that do not fix the URL on the fly.

This fix checks for the existence of the slash and if it's not found it will add one.


drewfish:
@capecodehq Can you post here the value of `staticHandling.prefix` that you are using?

PR #359, opened 2012-08-07
Added tunnelTimeout property to the app configuration object allowing users to set the timeout for the communication between the client and server through the tunnel.

zhouyaoji:


docs/dev_guide/intro/mojito_configuring.rst
```diff
@@ -131,6 +131,9 @@ configuration Object
 | ``tunnelPrefix``                                       | string               | "/tunnel/"        | The URL prefix for the communication tunnel            |
 |                                                        |                      |                   | from the client back to the server.                    |
 +--------------------------------------------------------+----------------------+-------------------+--------------------------------------------------------+
+| ``tunnelTimeout``                                      | number               | 30000             | The timeout in milliseconds for the communication      |
```
mridgway, line 134: The default value is actually 10000 as per the original hard coded value.

PR #362, opened 2012-08-08
Moved yuitest and jslint to devDependencies

mridgway:
If this is approved, go ahead and merge.

### DIFF
--- a/source/package.json
+++ b/source/package.json
@@ -16,11 +16,9 @@
         "connect": "1.8.2",
         "express": "2.5.2",
         "glob": "~3.1.11",
-        "jslint": "~0.1.9",
         "mime": "1.2.4",
         "semver": "1.0.14",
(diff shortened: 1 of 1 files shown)


isao:
+1

drewfish:
I mostly agree.  However, there is a user-training/documentation component (since users will need to install mojito with the `-dev` flag).

mridgway:
This shouldn't actually change anything for the user. By default, `npm install` will also install devDependencies. In order to not install them, one would pass the production flag: `npm install --production`

drewfish:
Ah, great.

+1

isao:
@mridgway `npm i mojito` won't install mojito's devDependencies - that will require `cd path/to/mojito && npm i` at least for recent npms

mridgway:
Yeah, I would argue that's a bug with npm.

PR #363, opened 2012-08-08
Bug fix for the "build html5app" commands use of URLs

ricallinson:
The build html5app command uses the URLs from the resource store to generate files. The URLs that it gets do not have the value of "appConfig.staticHandling.prefix" prepend if it was defined. This patch simply adds the prefix if it was found to the front of the URLs.


drewfish:
The res.url should definitely have the staticHandler.prefix attached.  If not, that's a bug elsewhere, and not a bug in commands/build.

drewfish:
Look in `app/addons/rs/url.server.js`, that's the _only_ place that each resource's URL is set.  It's possible that there's a bug elsewhere that is preventing addon-rs-url from getting the right appConfig.

drewfish:
Oh, addon-rs-url is just using the _static_ appConfig, so the URLs aren't context-sensitive.  That is definitely a bug (and be the bug you're running into).

source/lib/app/commands/build.js
```diff
@@ -161,6 +161,7 @@ exports.buildhtml5app = function(cmdOptions, store, config, destination,
         contextObj = libqs.parse(cmdOptions.context),
```
drewfish, line 161: (BTW, all other users of `contextObj` should probably be using `config.context` instead too.)

ricallinson:
Thanks Drewfish. I'll close this pull and open a new one once I've gone down the rabbit hole.

PR #364, opened 2012-08-08
restore warnings about DOM YUI modules on the server side

drewfish:


isao:
+1

ericf:
We have began discussing how we can surface which modules are not suitable for the server on the YUI Library website.

ericf:
I wanted to point out that I just created this page on the YUI website which allows people to easily see the set of modules which are natively supported in Node.js and those which require a DOM:
http://yuilibrary.com/yui/environments/

The current script we use to generate this is here:
https://gist.github.com/53ff69e7cc6de31c9543

PR #365, opened 2012-08-09
Made handlebars the default renderer for new projects and all examples

mridgway:
Rename of all .mu and .mu.html files to .hb and .hb.html respectively. CLI commands also now use handlebars.


drewfish:
:metal: ship it

caridy:
:+1: awesome work @mridgway

PR #370, opened 2012-08-09
PR #269 Added compatibility with Y.Model 

drewfish:
This is PR #269 updated with a test and delinted.


ericf:
@drewfish Will there ever be any invoking of the constructor-based models by Mojito, or is that left to the developer?

drewfish:
No, Mojito doesn't really do anything with the models, just manages them a little.  It is actually up to each controller to use the model as they see fit.

I'm strongly tempted to close this PR, since I think [Caridy's proposal](https://github.com/yahoo/mojito/pull/337#issuecomment-7497339) is cleaner/better.

ericf:
+1 for leveraging the `Y` object like in @caridy's comment.

drewfish:
I'm actually +1 for Caridy's suggestion too (despite authoring this PR).  What he suggested is the simple, natural way to use YUI modules (not matter what type).

caridy:
yay! if we are all agreed, let's do that then. Please, close this PR.

drewfish:
Closing this PR as "won't implement".  The natural YUI module should be used instead.

PR #209, opened 2012-06-15
Pull request for feature request: https://github.com/yahoo/mojito/issues/208

imalberto:
Changes with unit tests:
- source/lib/app/addons/ac/output-adapter.common.js
- source/lib/app/addons/view-engines/mu.server.js
- source/lib/app/autoload/view-renderer.common.js
- source/lib/tests/autoload/app/addons/view-engines/mu-tests.server.js 

Feedback welcome.

PR #371, opened 2012-08-10
Fix buffer mojito040

imalberto:
This pull request supercedes https://github.com/yahoo/mojito/pull/209 to merge with develop. Original pull request is outdated.


mridgway:
Hi Alberto, we recently added the Handlebars view engine. Do the changes to the mu engine need to be applied there as well? I can't tell whether the Handlebars engine would need any changing to support this.

mridgway:
Looks good to me. :shipit:

rwaldura:
Supersedes PR #209.

drewfish:
+1

mridgway:
So the options are now passed to the mustache engine but not the Handlebars. So we need to decide whether options should be passed to Handlebars or the options go away once we remove Mustache.

drewfish:
I like this feature, since it allows users to customize how the view engines run.  So I propose we pass the options to Handlebars.

caridy:
:+1: it will not hurt to have that in place in case someone wants to hack into the _template_ engine.

PR #372, opened 2012-08-10
Merge binder with mojito040

imalberto:
Pull request that addresses binder sizes being dispatch to user agents.
See [bugzilla 5749628] for more info.


source/lib/app/addons/ac/deploy.server.js
```diff
@@ -37,6 +40,309 @@ YUI.add('mojito-deploy-addon', function(Y, NAME) {
...
+                binder = binderMap[viewId];
+                // From Zed: Prune binderMap prior to deploying to client
+                if (!binder.name) {
+                    // Delete mojit entries that have no "binder"
+                    delete binderMap[viewId];
+                    binder = null;
+                } else if (binder.children) {
+                    // Don't have MVC running locally, so don't need
+                    // the "children" mojits either
+                    delete binder.children;
```
redonkulus: We do this in our app now and was introduced before we moved all affinities from server to common. This could cause issues with mojitProxy.destroyChild and possibly introduced memory leaks when refresh view is used. Its worth looking into the affects this has on the client before its baked into Mojito.

@kfay may have more info here.

source/lib/app/addons/ac/deploy.server.js
```diff
@@ -16,7 +16,10 @@ YUI.add('mojito-deploy-addon', function(Y, NAME) {
 
     var fs = require('fs'),
         yuiFilter = 'min',
-        minify;
+        minify,
+        DEPLOY_YUI_PREFIX = '$x_',
+        DEPLOY_MOJITO_PREFIX = '$y_',
```
drewfish: Not sure how these are used (yet), but it might help avoid confusion of the YUI_PREFIX is `$y_` :smile:  (The MOJITO_PREFIX could be `$m_`.)

source/lib/app/addons/ac/deploy.server.js
drewfish: Yeah, deleting `binder.children` is quite dangerous, since I think it might break some of the advanced features of client-side mojito.

source/lib/app/addons/ac/deploy.server.js
```diff
@@ -37,6 +40,309 @@ YUI.add('mojito-deploy-addon', function(Y, NAME) {
...
+                    delete binder.children;
+                }
+                if (binder !== null) {
+                    for (module in binder.needs) {
+                        if (binder.needs.hasOwnProperty(module)) {
+                            if (binder.needs[module] &&
+                                    Y.Lang.isString(binder.needs[module])) {
+                                path = binder.needs[module];
+                                if (yuiBase && -1 !== path.indexOf(yuiBase)) {
+                                    path = path.replace(yuiBase, yuiPrefix);
```
caridy: Why do we need a yuiPrefix?

source/lib/app/addons/ac/deploy.server.js
imalberto: Thanks for the feedback. In retrospect, line185-193 should probably be removed completely. Agreed ?

kfay: I think there's still some validity in lines 185-188. If there's no binder for a mojit, we don't need the binderMap information.

source/lib/app/addons/ac/deploy.server.js
imalberto: All YUI modules share a common base URL. In this case, yuiPrefix replaces that base URL. Once on user agent, the YUI base URL can be recreated.

caridy: @imalberto so this feature is targeting "mojito build" right? If that's the case, how is this different from having a custom yui->config setting in application.json for a particular dimension with these custom values: comboBase, base, combine, filter?

imalberto: @caridy Just to clarify, this is not a "mojito build" feature -- that code is triggered by constructMojitoClientRuntime(). Others can comment, but by the time ac.deploy addon is run, all the yui modules urls have already been set (with correct base, comboBase) -- all that is done in mojito-loader. If those custom values are set in application.json, that would work. See line 640 - it would definitely be better if this was retrieve from config.

source/lib/app/addons/ac/deploy.server.js
imalberto: Agreed :)

imalberto:
Rejecting PR after discussion with Ric.

PR #373, opened 2012-08-10
fix cp/paste error

isao:


mridgway:
:shipit: For things like this I don't think a PR is necessary.

PR #376, opened 2012-08-10
Fixed test descriptors to work on server side

mridgway:
Now that the new Arrow has been released, the commonlibs are now included before the test libs. This allows us to specify mojito-test.js in the commonlibs so that it is loaded on both the client and server side. I removed the script tag for mojito-test.js in the mojito-test.html page since it is no longer needed.

This also includes a port of the cookie.server.js tests as well.

Note: Update Arrow to get these tests to run.


isao:
shipit :)

PR #377, opened 2012-08-11
Allow jslint configuration to be overridden by cli.json config file

redonkulus:
This allows applications to override the default jslint configurations via a new config file called cli.json. This file could also be used in the future to augment other mojito commands like "test".

The {app}/cli.json will look like this:

``` json
{
    "jslint": {
        "white": true,
        "sloppy": true
    }
}
```

### DIFF
--- a/source/lib/management/commands/jslint.js
+++ b/source/lib/management/commands/jslint.js
@@ -31,7 +31,7 @@ OPTS = (function () {
 
     // Load options from cli.json
     try {
-        options = JSON.parse(fs.readFileSync(path.join(process.env.PWD, 'cli.json'), 'utf-8')).jslint.rules;
+        options = utils.getCliConfig().jslint.rules;
     } catch (e) {
         options = {};
     }
--- a/source/lib/management/commands/jslint.js
+++ b/source/lib/management/commands/jslint.js
@@ -11,6 +11,7 @@
 var fs = require('fs'),
     path = require('path'),
     utils = require('../utils'),
+    OPTS = {},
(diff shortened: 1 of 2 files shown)


mridgway:
Can you move the rules down a level to something like:

```
{
    "jslint": {
        "rules": {
            "white": true,
            "sloppy": true
        }
    }
}
```

I could see value of having configuration for excluded directories as well. I want to make sure that we can limit changes to this configuration in the future.

Otherwise, +1 from me.

redonkulus:
Good call, moved under "rules".

isao:
+1

source/lib/management/commands/jslint.js
```diff
@@ -23,6 +24,46 @@ var fs = require('fs'),
...
     ];
 
+// Optionally, allow application to override default set of configurations from cli.json file
+OPTS = (function () {
+    var options,
+        defaultOptions;
+
+    // Load options from cli.json
+    try {
+        options = JSON.parse(fs.readFileSync(path.join(process.env.PWD, 'cli.json'), 'utf-8')).jslint.rules;
```
drewfish: Reading the cli.json file should go into management/utils.js.  We'll want to add support for cli.json in the other commands in the future.

redonkulus:
@drewfish this was fixed last week, forgot to update PR.

jenny:
We are proposing to move to JS Hint. Will close this for now, but let's revisit after that lands.

redonkulus:
@jenny will we be able to customize the jS Hint config settings?

PR #378, opened 2012-08-13
Added documentation covering the Mojito architecture and mojits

zhouyaoji:


drewfish:
The "model factory" confuses me.  As far as I know, there is no such thing in Mojito.

PR #381, opened 2012-08-13
add node 0.8 compatibility 

isao:
test for node 0.8's fs.existsSync() or older path.existsSync() and
use the appropriate one. (async form not used in mojito). 

mojito test passes on node vers 0.8.6 and 0.6.19


rwaldura:
does this PR supersede #250?

isao:
@rwaldura yes, that pr couldn't merge automatically, and assumed `var path` etc where mojito code used `var pathlib`, var `libfs` etc... Also didn't incl 0.6 support

isao:
updated to be merge-able, omitting the dep version edits

mridgway:
:shipit:

Issue #375, opened 2012-08-10
shared assets have wrong URL, especially when staticHandling.appName is "/"


drewfish:
Consider an app with shared assets and staticHandling.appName = "/".  In that case, the shared assets should have URL prefix "/static/" or perhaps "/static/shared/" (it's debatable).  However, it looks like they have prefix "/static///".

redonkulus:
@rwaldura this would be a blocker from upgrading to 0.4.

Even with changing appName to something other than "/", I still do not get the shared assets to load.

rwaldura:
is this related to #358?

drewfish:
@rwaldura That's unknown, since the root cause of #358 is unknown.

PR #383, opened 2012-08-13
fix issue #375, unified handling of frameworkName, appName, and prefix

drewfish:
fix for issue #375


mridgway:
Looks good to me.

PR #388, opened 2012-08-14
first version of performance harness

drewfish:
probably over-engineered, but hopefully flexible enough to accommodate future needs


drewfish:
will port to javascript

PR #390, opened 2012-08-15
Moved npm configuration to the repository root

mridgway:


mridgway:
I'll move the {bin,lib} folders up now then.

That's an interesting point. Do we need the tests/ directory to be in the npm package? Do our functional tests rely on this? Right now, I've made it so that only the bin, docs, and lib folders get packaged.

mridgway:
I think I'd prefer to add a special case for node_modules directories that are used for tests. We may have cases (and I have had this before) where I've installed node_modules into an example app to try things out.

mridgway:
I removed the other PR since this one included those commits.

You can see Travis CI working at http://travis-ci.org/#!/mridgway/mojito

mridgway:
Oops, meant to post these comments in the other PR.

PR #391, opened 2012-08-15
Travis CI Configuration

mridgway:
This will get Travis CI running our `mojito test` unit tests for Pull Requests. Once Arrow is open sourced I would like to explore using Travis to run Arrow functional and unit tests as well.

Note: This includes commits from PR #390

PR #392, opened 2012-08-15
Functional test port

mridgway:
Note: This relies on the PR #391 (Travis and package.json changes) so that the ./tests/func directory gets access to node_modules.

This is a port of the Arrow functional tests from the qetest repo. I also fixed the run.js script such that it should run correctly now. Eventually this script could be moved to a `mojito arrow` command and then add unit tests to it.

I now run it using `./tests/func/run.js test -s`

Options are the following:

```
    .option('-b, --no-build', 'Don\'t build the apps')
    .option('-d, --no-deploy', 'Don\'t deploy the apps')
    .option('-s, --no-selenium', 'Don\'t start selenium')
    .option('-a, --no-arrow', 'Don\'t start arrow server')
```

I use `-s` because I have my selenium server running at startup.

It looks like there are some failing tests, but I think Lichun can help out with those. I'd like to get functional tests running from Arrow as soon as we can.


drewfish:
+1

The new `run.js` looks real nice.

PR #394, opened 2012-08-15
fix broken UTs in mojit skeletons

FabianFrank:
the renamed tests require the DOM and therefore need to have set the client archetype until mojito provides them a DOM on the server. today mojito creates apps that crash during mojito test. a lot of people reported this during the open hack day in bangalore, after they started playing with mojito.


FabianFrank:
closing duplicate, I seem to experience github hickups right now

Issue #261, opened 2012-07-12
Ability to use custom yui cdn in mojito


focuzz8:
Mojito generates sandboxes for mojits using yahoo cdn, but this will not work, if you need for example https. So yui can work with custom yui cdn, but in mojito it seems like links are hardcoded.
Solution that works for me: http://developer.yahoo.com/forum/Yahoo-Mojito/Configuration-to-load-YUI-locally/1340702863093-e3799041-c2a5-4b8d-9690-bbb41effa133

mridgway:
It looks like this could be fixed by passing "base" or "comboBase" into the configuration for YUI. The ability to configure YUI was added in https://github.com/yahoo/mojito/pull/350. @focuzz can you confirm that this will work for you?

focuzz8:
yes, but I've just found one more place to patch this way (and it seems like there is bug there since this patch and new rs)
lib/app/addons/rs/yuiserver.js at line 482 -> loader.base = this.yuiConfig.base;

focuzz8:
PS you can see result of wrong url generation (with yahoo cdn) in mojito log. Something like

```
[MOJITO] (1345200114416) qeperf: mojit instance expansion complete: yui_3_5_1_2_4_1345200105025_10123
[MOJITO] (1345200114416) qeperf: loading YUI modules for YUI instantiation: yui_3_5_1_2_4_1345200105025_10123
[MOJITO] (1345200114416) mojito-loader: loading http://yui.yahooapis.com/3.5.1-2/build/datatype-date-format/lang/datatype-date-format_en.js, ...
```

when using internationalization

rwaldura:
Thanks @focuzz ! Let's try and fix this.

caridy:
I don't think the internationalization is related with the "base" or "comboBase", but I will investigate. I can ensure you that my patch for SSL support works just fine since we use it in Axis ;)

caridy:
@focuzz can you try to add an empty:

   "yui": {
         "config": {}
   }

in your application.json to see if that solves the problem with the language property in the combo url?

focuzz8:
yes, mojito with empty yui section works

caridy:
I think we can close this issue since yui.base or yui.comboBase should not be used to customize the client configuration, it only affect the way mojito populates YUI at the server side (by default from node_modules/yui/.

For the client, you can use yui.config.comboBase in application.json.

@rwaldura probably we should clarify that in the documentation.

rwaldura:
[...]

PR #406, opened 2012-08-17
honoring context.lang in the loader

caridy:
It should ALWAYS be as part of the loader configuration, even when application.json >> yui.config is undefined

This solves the issue reported by @focuzz on issue #261.


isao:
+1

PR #408, opened 2012-08-18
Fix for handlebar renderer when ac.flush() is called more than once

imalberto:
PR that fixes the issue (in mojit):

ac.flush('hello ');
ac.flush('world!');
ac.done();


mridgway:
:shipit:

rwaldura:
See also #452 which appears related.

PR #418, opened 2012-08-21
Bug Fix: Fixes how "mojito compile inlinecss" generates its output

carystanley:
Bug Fix: Fixes how "mojito compile inlinecss" generates its output, so that each instance doesn't clobber the namespace of _mojito._cache.compiled.css.inline

### DIFF
--- a/lib/app/commands/compile.js
+++ b/lib/app/commands/compile.js
@@ -30,6 +30,7 @@ var libpath = require('path'),
 
     // private class
     YuiModuleCacheWriter,
+    YuiInlineCssCacheWriter,
 
     // public exports
     usage,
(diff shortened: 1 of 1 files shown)


rwaldura:
@dferreiroval does Shaker make this PR obsolete?

diervo:
@rwaldura Right now Shaker doesn't support inlining throw fies, but I have open a issue on Shaker for that. AFAIK the compiler version is kind of dummy, so I would like to implement something more robust. Anyway this compiler option it's interesting in the meanwhike. I let you know.

diervo:
@rwaldura @jenny This feature will be available as part of the new set of features in Shaker3.0, which will be out in January. This means that Shaker will take of that. Other thing is if you want to keep the cssinline thing on Mojito.
Close accordingly.

drewfish:
@dferreiroval that would be awesome!  I'm inclined to rip out this feature once shaker has a better version.  I'm not sure if we need to fix this in the mean time, since I'm not sure how many users are using `mojito compile inlinecss`.

carystanley:
I agree, if Shaker is to replace this functionality then it should be taken out of Mojito.
(It only worked for the "default" context anyways)

caridy:
This is the realm of mojito-shaker, in 3.x shaker supports this.

PR #420, opened 2012-08-22
Moved fixtures; Migrated resource store addon tests

mridgway:
Check out the path to fixtures and make sure this is where we want them.


mridgway:
Travis lies. Looks like it could be an environment error on their 0.4 box.

PR #424, opened 2012-08-23
change example & fixture mojito dependency to >0.3

isao:


examples/newsboxes/package.json
```diff
@@ -14,6 +14,6 @@
         "npm": "> 1.0"
     },
     "dependencies": {
-        "mojito": "0.3.x"
+        "mojito": "> 0.3"
```
mridgway, line 17: Why a space between > and 0.3? npm documentation states that ">version" is the format.

isao, line 17: lines 13 & 14 use "> ", as does mojito's package.json, so I tried to be consistent.

mridgway:
+1

PR #425, opened 2012-08-23
Store client url fix

mridgway:
Clean up and tests for #358


isao:
+1 (I was actually doing this one as well..)

PR #426, opened 2012-08-23
Remove colors.js and depend on colors npm pacakge

mridgway:


isao:
+1

we should also turn off colors if `!process.stdout.isTTY`

Issue #431, opened 2012-08-24
.npmignore excludes lib/tests and test coverage jars


isao:
this breaks `mojito test -c app .`

isao:
## steps to reproduce

with HEAD of develop d3df9426a13823d766dee519df02bf6e1f4f1c98

```
% npm rm -g mojito
% cd some/mojito app
% npm i path/to/mojito#d3df9426a13823d766dee519df02bf6e1f4f1c98
% ./node_modules/mojito/bin/mojito test app -c .
Instrumenting "/Users/isao/Desktop/newsboxes" for test coverage
    (this will take a while).

fs.js:420
  return binding.lstat(pathModule._makeLong(path));
                 ^
Error: ENOENT, no such file or directory '/Users/isao/Desktop/newsboxes/node_modules/mojito/lib/tests'
    at Object.lstatSync (fs.js:420:18)
    at Object.realpathSync (fs.js:878:23)
    at instrumentDirectory (/Users/isao/Desktop/newsboxes/node_modules/mojito/lib/app/commands/test.js:498:24)
    at /Users/isao/Desktop/newsboxes/node_modules/mojito/lib/app/commands/test.js:803:13
    at Object.run (/Users/isao/Desktop/newsboxes/node_modules/mojito/lib/app/commands/test.js:651:9)
    at main (/Users/isao/Desktop/newsboxes/node_modules/mojito/lib/management/cli.js:125:13)
    at Object.<anonymous> (/Users/isao/Desktop/newsboxes/node_modules/mojito/lib/management/cli.js:137:1)
    at Module._compile (module.js:446:26)
    at Object..js (module.js:464:10)
    at Module.load (module.js:353:31)
```

PR #432, opened 2012-08-24
fix issue #431 where test coverage deps were excluded

isao:
change "tests" from .npmignore to "/tests" as these
former case excluded things required for mojito test,
specifically lib/tests/harness/lib/yuitest/java

TODO use coverage jars in yuitest npm pkgs or similar
(though the don't appear in npm yui or yuitest yet)?

also ignore "node_modules" in top level or in examples/*
only (test fixtures need to contain node_modules dirs


mridgway:
I think we should exclude all `tests` as it was before and then add back in the jars that are needed (!lib/tests/harness). The tests themselves are not needed by a downstream application, so we should exclude them.

isao:
top level (arrow) tests are excluded

but we need fw tests to support `mojito test` command per the docs

.gitignore
```diff
@@ -1,12 +1,12 @@
 .DS_Store
-node_modules
-!tests/fixtures/packages/node_modules
-!tests/fixtures/packages/node_modules/*/node_modules
```
drewfish, line 4: So `tests/fixtures/packages/node_modules` (and children) are definitely needed in order to do proper testing in the resource store.  Does the new .gitignore support that (it's hard for me to tell from looking at the rules).

isao:
BTW, we should get rid of lib/tests/harness/lib/yuitest/ in favor of npm packages with https://npmjs.org/package/yui-coverage

mridgway:
@isao Agreed. I think we should evaluate that as we make `mojito test` work with Arrow.

.gitignore
isao, line 4: yes, I know, I said as much this morning.. :)

`./node_modules` and `./examples/**/node_modules` are _excluded_ and everything else is allowed

mridgway, line 4: Yeah, I'd prefer to be explicit about what we want to keep as it was before.

isao, line 4: yes, sorry for the funny diff.. just look at the full file.

I sorted to compare with `.npmignore`

PR #433, opened 2012-08-24
Apply the "pathToRoot" if the "contentPath" is *not* relative

isao:
this is a cleanup of pull #284
thanks @capecodehq


mridgway:
Where did Ric's commit go? I feel like we should preserve the original PR's commits to give them credit.

isao:
Agreed, but git apply/am didn't work after the source dir change, even though patch -p2 did..

drewfish:
Is there a unit (or functional) test for this?

drewfish:
I think there's already a setting to fix this.  The problem is more complicated than just adding a ".", since the HTML file could be built at foo/bar/baz.html, in which case the static URLs should be prefixed with "../..".

See `forceRelativePaths` here:
http://developer.yahoo.com/cocktails/mojito/docs/intro/mojito_configuring.html#html5app-obj

PR #439, opened 2012-08-27
fix (and un-ignore) the resource store unit test for optional YUI modules

drewfish:


drewfish:
travis build fails on only one node version (0.8), due to errors we've seen (transiently) in other builds

@travisbot please try again

mridgway:
+1

PR #443, opened 2012-08-28
[docs fix 5779384] Removed errant backticks.

zhouyaoji:


mridgway:
Typos and small fixes like this don't need to be in pull requests. They can just be merged into develop manually.

Issue #438, opened 2012-08-27
AdapterResourceStore: expandInstance


diervo:
The adapterResourceStore acts as a proxy for the real resourceStore (where the pluggins are binded).
The AdapterResourceStore removes the action from the spec when passing it to the real store. (Lines: 65-72 resource-store-adapter)
In consecuence, from the plugins I cannot know which action was called.
Adter all the calling the adapterStore merge back the original specs with the ones expanded (Line:88)

So the store.expandInstance would need the full instance object to be able to distinguish the action.

@drewfish P1 for Shaker please!! :)

diervo:
@rwaldura Could you assing proprely priority? Shaker2.0 should not ship until you guys can fix me this issue.
I would like to fix it but I think it may brake other things since is a change on the store... So drew to the rescue! ;)

PR #447, opened 2012-08-28
Update lib/app/mojits/TunnelProxy/controller.server.js

diervo:
We need to pass the right action when we call dispatch so the store "expandSpec" get executed with the right action.

### DIFF
--- a/lib/app/mojits/TunnelProxy/controller.server.js
+++ b/lib/app/mojits/TunnelProxy/controller.server.js
@@ -91,7 +91,7 @@ YUI.add('TunnelProxy', function(Y, NAME) {
                 );
                 return;
             }
-
+            proxyCommand.instance.action = proxyCommand.action;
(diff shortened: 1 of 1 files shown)


drewfish:
@dferreiroval is this to fix issue #438 ?

diervo:
@drewfish nop, but is the same thing in other place, the command has the wrong action again, and the action you need is within command.instance. I dont know exactly why. Maybe the fix should be more deeper in the code... this fix it so far...

drewfish:
Yeah, this PR doesn't look right.  The action should be attached to the `command`, and that should be sufficient.  (We shouldn't have to add the action to the `instance`.)

diervo:
the action is there on "command" object,
but for some reason command.action = 'index' always
and then the real one is in command.instance.action

dont know why...

diervo:
and again we will remove this since we have another way to do it.

PR #448, opened 2012-08-28
[YAF] view to template

drewfish:


mridgway:
Yowza! +1 conceptually. Not sure if I can read through all of the changes though =P

drewfish:
Yeah, 646 files changed :smile:, which makes me wonder how many files left _haven't_ changed.

drewfish:
The unit tests (both kinds) pass when I run them by hand.  Haven't figured out how to run the functional tests yet though.

rwaldura:
let's complete this when we really have some level of integration with YAF

Issue #429, opened 2012-08-24
PUT requests not handled correctly by the REST library


gotwarlost:
from rest.common.js:

``` javascript
                if ('GET' === method) {
                    if (-1 === url.indexOf('?')) {
                        url += '?' + params;
                    } else {
                        url += '&' + params;
                    }
                } else if ('POST' === method) { /* WHAT ABOUT PUT? */
                    ioConfig.data = params;
                }
```

This bug is not present in 0.3.26 and was introduced in 0.3.27. It is affecting our capability to move forward since we need PUT requests to a backend webservice to work.

rwaldura:
Ouch.
https://github.com/yahoo/mojito/blob/develop/lib/app/autoload/rest.common.js#L75

rwaldura:
Next Sprint candidate. Sprint's already pretty full though -- might get bumped. How critical is this? What's it blocking?

mojit0:
A proper REST implementation should support GET, POST, PUT, and DELETE. For performance reasons it's sometimes useful to also support HEAD requests. As part of that support HTTP status codes other than 200 can indicate success and should be properly handled. We may want to open a separate issue for full REST support so we can focus this bug on only what's being blocked at the moment tho.

gotwarlost:
agree with @mojit0 - currently we are unable to move forward with our mojito dependency and have to pin ourselves to v0.3.26

This particular fix should be a one-line change (i.e. add a clause for PUT in addition to POST) - the effort is more on writing a unit test for this.

rwaldura:
Next Sprint, S41.

mridgway:
Looks like this regression was added by #196 which was a fix for a YUI 3.5.1 regression. If we're upgrading to YUI 3.6 we can roll back #196 because it was fixed with IO.

mridgway:
Upgrade to YUI 3.6 is blocked by a regression, so we're waiting until 3.6.1 or 3.7.0 before we upgrade. Therefore, I am fixing this in place as it is now. PR incoming.

mridgway:
Fixed in #450.

PR #450, opened 2012-08-29
Fixed Issue #429: Added params to body for PUT requests

mridgway:
We should probably look at what's going on with other request methods too, but if we're going to upgrade YUI in the next couple of weeks, these fixes would be rolled back anyway. For now, just fixing up the PUT.

Fully abstracted out the Y.io call to make for better mockability. Fixed up the functional and unit tests as well.

### DIFF
--- a/lib/app/autoload/rest.common.js
+++ b/lib/app/autoload/rest.common.js
@@ -53,9 +53,29 @@ YUI.add('mojito-rest-lib', function(Y, NAME) {
     Y.namespace('mojito.lib').REST = {
 
         /**
+         * Calls IO. Provides mockability
+         * @param url
+         * @param config the IO configuration
+         * @param callback
          * @private
          */
-        _doRequest: function(method, url, params, config, callback) {
+        _doRequest: function (url, config) {
+            Y.io(url, config);
+        },
+
+        /**
+         * Creates the configuration to be passed to the IO request
+         *
+         * @param {String} method HTTP method
+         * @param {String} url HTTP location
+         * @param {Object} params Params that are passed in the request
+         * @param {Object} config Additional configuration
+         * @param {Object} config.headers Headers to be sent with the request
+         * @param {Number} config.timeout Timeout for the IO request
+         * @param {Function} callback The handler for the response
+         * @private
+         */
+        _makeRequest: function(method, url, params, config, callback) {
             // TODO: [Issue 72] Figure out why 'params' values are attaching
             // themselves to headers!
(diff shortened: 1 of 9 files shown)


mridgway:
Ouch, didn't change old unit tests...

caridy:
:shipit:

lib/app/autoload/rest.common.js
```diff
@@ -53,9 +53,29 @@ YUI.add('mojito-rest-lib', function(Y, NAME) {
     Y.namespace('mojito.lib').REST = {
 
         /**
+         * Calls IO. Provides mockability
+         * @param url
+         * @param config the IO configuration
+         * @param callback
          * @private
```
drewfish, line 61: I think  you also have to specify `@method` because yuidoc doesn't parse it out automatically.

lib/app/autoload/rest.common.js
```diff
@@ -53,9 +53,29 @@ YUI.add('mojito-rest-lib', function(Y, NAME) {
...
+         * Creates the configuration to be passed to the IO request
+         *
+         * @param {String} method HTTP method
+         * @param {String} url HTTP location
+         * @param {Object} params Params that are passed in the request
+         * @param {Object} config Additional configuration
+         * @param {Object} config.headers Headers to be sent with the request
+         * @param {Number} config.timeout Timeout for the IO request
+         * @param {Function} callback The handler for the response
+         * @private
```
drewfish, line 77: I think  you also have to specify `@method` because yuidoc doesn't parse it out automatically.

lib/app/autoload/rest.common.js
```diff
@@ -53,9 +53,29 @@ YUI.add('mojito-rest-lib', function(Y, NAME) {
...
         /**
+         * Calls IO. Provides mockability
+         * @param url
+         * @param config the IO configuration
+         * @param callback
          * @private
          */
-        _doRequest: function(method, url, params, config, callback) {
+        _doRequest: function (url, config) {
+            Y.io(url, config);
```
drewfish, line 64: This is a private one-line method, that is called in only one place. Do we really need it?

mridgway, line 64: This is intentional so that we can mock out _doRequest. Before the unit tests were mocking out what is now _makeRequeset, which skips 90% of the logic in this library.

drewfish, line 64: OK, makes sense.

drewfish:
+1

PR #451, opened 2012-08-29
mojito yaf-develop to use the bleeding version of yui to keep them in sync

caridy:
trying to get travis to build yui npm module using after_install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ node_js:
   - 0.8
   - 0.6
   - 0.4
-after_install:
-    - ./travis/yui_after_install.sh
+before_install:
+    - ./travis/before.sh
+install:
+    - ./travis/install.sh
 notifications:
   email:
     recipients:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ node_js:
   - 0.8
   - 0.6
   - 0.4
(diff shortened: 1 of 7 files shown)


drewfish:
Is this the approach that YUI uses for their travis builds?

drewfish:
From the travis logs:  `make: *** No rule to make target 'npm'.  Stop.`

caridy:
@drewfish yes, this is the approach they use. I had some issues with the path, and the workflow, should be solved with the last  commit.

caridy:
@drewfish it seems to be working now. Can you try it locally, by running `./travis/travis.sh`?

Things that are pending:
- how to facilitate the use of this travis config for bleeding branches only but not on develop and master?
- how to speed up the process since it takes a long time now due the YUI builder?

mridgway:
+1, looks good.

drewfish:
I got this error:

```
shifter --config build-npm.json
make: shifter: No such file or directory
make: *** [npm] Error 1
```

drewfish:
Also, after cloning the YUI repo into `yui-src`, it doesn't appear to checkout the `3.x` branch.

caridy:
@drewfish I added the -b 3.x option on the clone command as part of #459

PR #452, opened 2012-08-29
Remove call to flush for client side hb to prevent multiple done calls

mridgway:
During some testing I discovered an issue with the client side HB view engine that made the output handler call `done` twice (this is arguably an issue with the client side output handler, but that's an issue for another day).

I also cleaned up the cacheTemplates configuration which could have thrown a "Cannot read property of undefined" error.

### DIFF
--- a/lib/app/addons/view-engines/hb.client.js
+++ b/lib/app/addons/view-engines/hb.client.js
@@ -31,22 +31,27 @@ YUI.add('mojito-hb', function(Y, NAME) {
          * @param {string} tmpl The name of the template to render.
          * @param {object} adapter The output adapter to use.
          * @param {object} meta Optional metadata.
+         * @param {boolean} more Whether there is more to be rendered
          */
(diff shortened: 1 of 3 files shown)


drewfish:
Should this have a unit test as well?

mridgway:
Yep, was already on its way =D

rwaldura:
Did @capecodehq report this? He was going to.

rwaldura:
Acutally, his issue  may or may not be the same. He reported (verbally) seeing some duplicate output when using HB.
I'll let @capecodehq elaborate further, and maybe open a separate ticket.

mridgway:
It may in fact be the same issue. I was seeing the invoke handler being called twice, because adapter.flush called ac.done and then adapter.done also called ac.done.

caridy:
@mridgway that solution is more elegant, thanks!

:shipit:

PR #456, opened 2012-08-30
fix #438 more lightweight resource store adapter

### LINKED ISSUE #438: AdapterResourceStore: expandInstance

drewfish:
fixes #438 #76 #77 #78

### DIFF
--- a/lib/app/autoload/resource-store-adapter.common.js
+++ b/lib/app/autoload/resource-store-adapter.common.js
@@ -48,66 +48,37 @@ YUI.add('mojito-resource-store-adapter', function(Y, NAME) {
         },
 
 
-        expandInstanceForEnv: function(env, instance, context, callback) {
-            var base = {},
-                source = {},
-                my = this;
+        expandInstanceForEnv: function(env, inInstance, context, callback) {
+            var me = this;
 
-            if (!instance.instanceId) {
-                instance.instanceId = Y.guid();
-                //DEBUGGING:  instance.instanceId += '-instance-common-' +
-                //    [instance.base||'', instance.type||''].join('-');
-            }
-            // DEPRECATED, but kept in case a user is using.
-            instance.guid = instance.instanceId;
+            return this.store.expandInstanceForEnv(env, inInstance, context, function(err, outInstance) {
+                if (err) {
+                    callback(err, {});
+                    return;
+                }
 
-            // What are being asked to expand?
-            if (instance.base) {
-                source.name = instance.base;
-                source.func = this.getSpec;
-            } else if (instance.type) {
-                source.name = instance.type;
-                source.func = this.getType;
-            } else {
-                // We don't have any inputs so fail
-                throw new Error('There was no info in the "instance" object');
-            }
+                // Ensure the "instance" has been properly resolved. If
+                // there are no specs in the application.json file, there is
+                // an error below because the instance is invalid. We should
+                // check here for a valid instance object and throw an error
+                // if it is not. This happens because someone could create a
+                // routes.json file with routes that don't route to mojit
+                // instances, and the URI router creates invalid commands,
+                // which are passed into the dispatch.
+                if (!me.validate(outInstance)) {
+                    callback(new Error('Instance was not valid.'));
+                    return;
+                }
 
-            // This contains the app "definition" and app config
-            my.getApp(env, context, function(app) {
+                if (!outInstance.instanceId) {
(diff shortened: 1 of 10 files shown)


lib/app/autoload/store.client.js
```diff
@@ -132,12 +132,61 @@ YUI.add('mojito-client-store', function(Y, NAME) {
...
+            // What are being asked to expand?
+            if (inInstance.base) {
+                source.name = inInstance.base;
+                source.func = this._getSpec;
+            } else if (inInstance.type) {
+                source.name = inInstance.type;
+                source.func = this._getType;
+            } else {
+                // We don't have any inputs so fail
+                throw new Error('There was no "base" or "type" in the "instance" object');
```
mojit0: return here?

lib/app/autoload/resource-store-adapter.common.js
```diff
@@ -48,66 +48,37 @@ YUI.add('mojito-resource-store-adapter', function(Y, NAME) {
...
+                    return;
+                }
+
+                if (!outInstance.instanceId) {
+                    outInstance.instanceId = Y.guid();
+                    //DEBUGGING:  outInstance.instanceId += '-data-common-' +
+                    //    [outInstance.base||'', outInstance.type||''].join('-');
+                }
+                // DEPRECATED, but kept in case a user is using.
+                outInstance.guid = outInstance.instanceId;
```
mojit0, line 79: Should we start adding warn level logging around deprecated features?

lib/app/autoload/store.client.js
```diff
@@ -132,12 +132,61 @@ YUI.add('mojito-client-store', function(Y, NAME) {
 
     ClientStore.prototype = {
 
-        /*
-         * TODO: REVIEW RE [Issue 76].
+
+        /**
+         * Expands the instance into all details necessary to dispatch the mojit.
+         * @method expandInstanceForEnv
+         * @param {string} env the runtime environment (either `client` or `server`)
```
mojit0: Comment text should be a full sentence form. Initial caps and end with a period.

lib/app/autoload/store.client.js
```diff
@@ -132,12 +132,61 @@ YUI.add('mojito-client-store', function(Y, NAME) {
...
+                source.func = this._getSpec;
+            } else if (inInstance.type) {
+                source.name = inInstance.type;
+                source.func = this._getType;
+            } else {
+                // We don't have any inputs so fail
+                throw new Error('There was no "base" or "type" in the "instance" object');
+            }
+
+            // Here we get either the a spec or a type
```
mojit0: Grammar

lib/app/autoload/store.client.js
drewfish: After a `throw`?  I've never seen that before.

drewfish: Hmm... that should `callback(new Error(...))` instead of `throw`.  I'll make that change.

lib/app/autoload/resource-store-adapter.common.js
drewfish, line 79: This is already in the "deprecation" notes.  It's been deprecated for a while, and is ripe for removal in my opinion.

I'm not sure how we'd put a warning on a property we always populate (seems noisy) unless we re-implement this as a property getter.  Not worth the effort in my opinion.

lib/app/autoload/store.client.js
drewfish: I'll make that change.  However, according to the grammar rules I learned (or at least, the way I remember it), a sentence can be fully grammatical even without a subject or verb if those can be easily inferred.  As such, don't expect to see "This parameter is" at the beginning of the sentence :wink:

mojit0: I'm working more from the "imagine the comment in YDN docs perspective. So:

The runtime environment (either 'client' or 'server').

drewfish:
Travis failure is our recurring intermittent `mojito-mu-tests` error.

PR #459, opened 2012-08-31
support for route events in controllers

caridy:
This is the preliminar work to support events in the controller.

Traditionally, you define a controller like this:

```
Y.namespace('mojito.controllers')[NAME] = {
    init: function(config) {
        this.config = config;
    },
    index: function(ac) {
        ac.done('');
    }
};
```

we want to also support:

```
Y.namespace('mojito.controllers')[NAME] = {
    init: function(config) {
        this.config = config;
        this.on('index', this.onIndex, this);
    },
    onIndex: function(e) {
        e.ac.done(''); // this API is subject to change
    }
};
```

or even something more advanced like this (prototypal definition):

```
Y.namespace('mojito.controllers')[NAME] = function (config) {
    this.config = config;
    this.on('index', this.onIndex, this);
};
Y.namespace('mojito.controllers')[NAME].prototype = {
    onIndex: function (e) {
        e.ac.done(''); // this API is subject to change
    }
};
```

or even something super advanced like this:

```
Y.namespace('mojito.controllers')[NAME] = Y.Base.create(NAME, Y.BaseCore, [], {
    initializer: function (config) {
        this.config = config;
        this.on('index', this.onIndex, this);
    },
    onIndex: function (e) {
        e.ac.done(''); // this API is subject to change
    }
}, {
    ATTRS: {
        photos: {} // potentially a global model instance
    }
});
```

obviously, we can do Y.App instead of Y.BaseCore on this last example, and it should provide a fully functional App fragment in a form of a mojit.

/cc @ericf @drewfish

### DIFF
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -303,9 +303,12 @@ YUI.add('mojito-action-context', function(Y, NAME) {
             // If the action is not found try the '__call' function
             if (Y.Lang.isFunction(controller.__call)) {
                 actionFunction = '__call';
+                Y.log('__call is deprecated, please use a router event instead.', 'warn', NAME);
+            } else if (controller.fire && controller.getEvent && controller.getEvent(actionFunction)) {
+                Y.log('Using a custom event to trigger the action into the controller', 'info', NAME);
             } else {
-                // If there is still no joy then die
-                error = new Error("No method '" + command.action +
+                 // If there is still no joy then die
+                error = new Error("No method or event '" + command.action +
                     "' on controller type '" + instance.type + "'");
                 error.code = 404;
                 throw error;
@@ -320,7 +323,14 @@ YUI.add('mojito-action-context', function(Y, NAME) {
         Y.log('action context created, executing action "' + actionFunction +
             '"', 'mojito', 'qeperf');
 
-        controller[actionFunction](this);
+        if (controller[actionFunction]) {
+            Y.log('Action as a method is deprecated, please use a route event instead for action: ' + actionFunction, 'warn', NAME);
+            controller[actionFunction](this);
+        } else {
+            controller.fire(actionFunction, {
+                ac: this
+            });
+        }
     }
 
     Y.namespace('mojito').ActionContext = ActionContext;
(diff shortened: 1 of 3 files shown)


ericf:
I currently have YUI Pull Request [#168](https://github.com/yui/yui3/pull/168) open which allows `Y.BaseCore` to be used with `Y.Base.create()`. I've asked Satyen to review it so we can merge it into YUI's 3.x branch.

caridy:
With the last commit, we can support something like:

```
[
    {
        "settings": [ "master" ],
        "specs": {
            "foo": {
                "type": "Foo",
                "config": {
                     "models": {
                          "m1": "ModelXYZ"
                     }
                }
            }
        }
    }
]
```

where you can define regular mojito models under /models folder or mojits/MojitName/models folder, and those models will be instantiated, and they will be pushed into the controller as an attribute based on the mapping configuration.

In other words, a model instance can be pushed into a mojit, shared between mojits, serialized, and rehydrated by the framework before instantiating a mojit instance.

Next stop: how can the framework create/instantiate those models before dispatching any mojit?

drewfish:
The `models` config should be a top-level part of the spec (not inside `config`), as in:

``` javascript
[
    {
        "settings": [ "master" ],
        "specs": {
            "foo": {
                "type": "Foo",
                "models": {
                     "m1": "ModelXYZ"
                },
                "config": {
                }
            }
        }
    }
]

```

This is because the `config` is meant only for the controller. Anything that is for the framework (such as `type` or `deploy`) is a sister to `config`.

drewfish:
Besides that `spec.config.models` to `spec.models` change, +1.

caridy:
@drewfish the problem with spec.models is the store. As today, the instance.models will hold the mapping from the RS. Something like:

```
{ bar: true, foo: true, m1: 'ModelXYZ' }
```

where models/bar.server.js and mojits/Foo/models/foo.common.js are the models. And m1: 'ModelXYZ' is what we pass in the spec.

We can always do some conditional statements, but I rather prefer to use a different nomenclature, maybe "data", something like:

```
        "specs": {
            "foo": {
                "type": "Foo",
                "config": {},
                "data": {
                       "m1": "ModelXYZ"
                }
            }
```

or we can change the store :)

drewfish:
We should change the store to do the right thing. `specs` is user-facing configuration, and we should be focused on giving a good API to the users, and fix up our implementation to support that good API.

Issue #264, opened 2012-07-12
Compatibility with node.js on windows


focuzz8:
Since there is no more dependency on *nix (yui test migrated to 0.7.4) is there possible to make run of Mojito on windows? I spend some time, but found that there is problem with loader - in mojito/source/lib/index.js - YUI instance doesn't contain any standard modules, that should be loaded after init. Probably problem is in nodejs-yui, but I'm not sure.

rwaldura:
Hi @focuzz:
Indeed I don't think we ever tested that, seeing how we use mostly Macs and Linux boxen. 
Would you be able to narrow it down further? What seems to be the problem in nodejs-yui?

focuzz8:
Yes, I will try to find root of problem, today some hours later. It seems (as I remember) like yui loader can't find modules that are required to start mojito... I'll dig it a bit and will write or more deep description, or solution (if i'll find any)

rwaldura:
Yeah might just be a silly path issue.

focuzz8:
I've made some fixes, so mojito starts and renders mojits (there is problem with binders). So situation is better and it don't passes only 2 tests. You can see changes here https://github.com/focuzz/mojito/tree/win-compat
I'm not doing pull request, because multi os support slightly broke cody prettiness, because
- path.sep appeared only in node 0.8 and I'm using handmade function to get separator
- there is a lot of places where path.combine have used to generates url, what brakes on windows
- robably I've found not all places where fixes needed
  So can anybody review it and tell me whether I'm doing it in right way, or I should correct something?
  Thank you.

rwaldura:
Thanks Alexey. This is starting to sound like a bigger item than anticipated, does it not?

We're not currently setup to run our automated tests on Windows, so it's going hard for us to validate your fixes (Of course, we could validate they don't break MacOS/Linux, but that's only a negative test...)

Few people have asked for Windows support, it hasn't been a priority for us at this point.

focuzz8:
Well, I'm using them all (win + lin + mac) for different tasks, so it's not a critical problem for me right now. I understand about priority, so, I'll try to fix remaining issues in future(probably week or two). And after I can try to maintain this windows compatible branch, and merge your releases - and probably community will grow by time. If it will be interesting to anyone - probably it can be unofficial windows compatible branch or something like this in future (hope).

rwaldura:
Sounds like a good compromise. Thanks!
Let's keep this issue open to track this feature.

[...]

PR #460, opened 2012-09-04
Issue #264: Fixes for windows paths

mridgway:
This PR gets mojito up and running on Windows (tested on Windows 7 with Node 0.8.6). I tested against the "newsboxes" example app and all seems to be working.

Known issues:
- Unit tests need to be fixed so that they pass on a Windows box, although they still pass on our dev and build boxes
- `mojito jslint`'s exclusion matcher uses regular expressions that contain "/". It may be better to use glob to get a list of exclusions instead.
- Other command line tools haven't been tested

I'm sure there's more, but this should at least get the ball rolling for anyone out there that is interested in Windows support.

### DIFF
--- a/lib/app/addons/rs/url.server.js
+++ b/lib/app/addons/rs/url.server.js
@@ -176,6 +176,11 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 rollupParts = [],
                 rollupFsPath;
 
+            // Fix windows paths
+            if ('win32' === process.platform) {
+                relativePath = relativePath.replace(/\\/g, '/');
+            }
+
             // Don't clobber a URL calculated by another RS addon.
             if (res.hasOwnProperty('url')) {
                 return;
@@ -202,7 +207,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                         rollupParts.push(this.config.appName);
                     }
                     rollupParts.push('rollup.client.js');
-                    rollupFsPath = libpath.join(this.appRoot, 'rollup.client.js');
+                    rollupFsPath = [this.appRoot, 'rollup.client.js'].join('/');
                 }
             } else {
                 if ('mojit' === res.type) {
@@ -213,7 +218,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 if (res.yui && res.yui.name) {
                     rollupParts.push(res.mojit);
                     rollupParts.push('rollup.client.js');
-                    rollupFsPath = libpath.join(mojitRes.source.fs.fullPath, 'rollup.client.js');
+                    rollupFsPath = [mojitRes.source.fs.fullPath, 'rollup.client.js'].join('/');
                 }
             }
 
--- a/lib/app/addons/rs/url.server.js
+++ b/lib/app/addons/rs/url.server.js
@@ -176,7 +176,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 rollupParts = [],
                 rollupFsPath;
 
-            // Fix windows paths
+            // Translate from filesystem path to URL
             if ('win32' === process.platform) {
                 relativePath = relativePath.replace(/\\/g, '/');
             }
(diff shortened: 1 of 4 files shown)


lib/store.server.js
```diff
@@ -991,6 +992,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             if (!fs.isFile && fs.subDirArray.length < 2 && 'addons' === fs.subDirArray[0]) {
                 return true;
             }
+
```
mridgway: This new line made it in because of a console.log testing. I'll remove this.

lib/app/addons/rs/url.server.js
```diff
@@ -202,7 +207,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                         rollupParts.push(this.config.appName);
                     }
                     rollupParts.push('rollup.client.js');
-                    rollupFsPath = libpath.join(this.appRoot, 'rollup.client.js');
```
drewfish, line 205: This seems very strange to me.  My understanding was that the `path` library was supposed to do filesystem path construction in a way appropriate for each platform.

lib/app/addons/rs/url.server.js
```diff
@@ -176,6 +176,11 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 rollupParts = [],
                 rollupFsPath;
 
+            // Fix windows paths
+            if ('win32' === process.platform) {
```
drewfish, line 180: Is there a way that the right thing can happen without the `process.platform` check?  In this case, can the source of `relativePath` be fixed to not have the wrong path separator?

lib/app/addons/rs/url.server.js
mridgway, line 205: My understanding is that the URL addon is in charge of generating URLs, not file system paths. As such, we should not be using libpath here at all, or else we get URLs with backslashes.

lib/app/addons/rs/url.server.js
mridgway, line 180: On Windows, relative path contains backslashes, so in order to generate an appropriate URL, those backslashes need to be converted to forward slashes. I don't know of another way to do this. Theoretically you don't need the Windows check, so we could probably just remove the conditional.

lib/app/addons/rs/url.server.js
drewfish, line 205: That's mostly true. However, the `rollupFsPath` variable contains a path on the filesystem, and so should probably be using libpath.

lib/app/addons/rs/url.server.js
```diff
@@ -213,7 +218,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 if (res.yui && res.yui.name) {
                     rollupParts.push(res.mojit);
                     rollupParts.push('rollup.client.js');
-                    rollupFsPath = libpath.join(mojitRes.source.fs.fullPath, 'rollup.client.js');
+                    rollupFsPath = [mojitRes.source.fs.fullPath, 'rollup.client.js'].join('/');
```
drewfish: Ditto here, this is also a filesystem path.

lib/app/addons/rs/url.server.js
drewfish, line 180: Yeah, it took me a bit to realize it was translating from FS path to URL. Perhaps the comment should say that instead:  `// translate from filesystem path to URL`.

lib/app/addons/rs/url.server.js
mridgway, line 205: Got it. I'll change these fsPaths back to how they were before.

rwaldura:
this PR is for issue #264: Windows support

lzhan:
I will find a window machine to test and merge this pr.

caridy:
@jenny @lzhan can we get rid of this? :)

caridy:
Closing after merging #1103

PR #462, opened 2012-09-05
performance improvements

drewfish:
Three key things:
- replaced `Y.clone()`, it is expensive
- caching in `store.selector.getPOSLFromContext()`
- restored same caching of YCB file reads as is in `0.3.30`


mojit0:
+1

FabianFrank:
+1

PR #464, opened 2012-09-05
ack... lost context... newbie js error :(

drewfish:


mridgway:
+1. Travis failure can be ignored.

PR #467, opened 2012-09-06
Arrow Unit and Functional Tests Running on Travis

mridgway:
Go go Travis CI!

Full test suite takes about 15-20 minutes, but runs against 0.4, 0.6, and 0.8 in parallel.


.travis.yml
```diff
@@ -3,6 +3,14 @@ node_js:
   - 0.8
   - 0.6
   - 0.4
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - "wget http://selenium.googlecode.com/files/selenium-server-standalone-2.25.0.jar"
+  - "java -jar selenium-server-standalone-2.25.0.jar -p 4444 > /dev/null 2>&1 &"
+  - "cd node_modules/yahoo-arrow; npm install yui; cd ../..;"
```
mridgway: This is due to a nasty YUI dependency collision since Mojito already has it installed. I'd like to work with @proverma on this to figure out what is going on. Please note that developers that want to run the tests locally may have to do this same step.

package.json
```diff
@@ -57,7 +59,7 @@
     "bugs": "https://github.com/yahoo/mojito/issues",
     "scripts": {
         "start": "./bin/mojito start",
-        "test": "./bin/mojito jslint -p && ./bin/mojito test"
+        "test": "./tests/run.js test -u --group server --driver nodejs && ./tests/run.js test -u --group client --driver selenium"
```
mridgway, line 62: `npm test` will now just run the unit tests. We can debate whether it should also run `mojito test` as well, but I feel that it shouldn't run the functional tests since they take around 20 minutes to run.

.travis.yml
caridy: @mridgway we should move these statements into its own .sh file (arrow.sh maybe), just like we did with yaf-develop branch under the ./travis/ folder. In my mind, .travis.yml should be simple enough, just having references to scripts we want to execute.

We can also create a ./travis/travis.sh script for anyone who want to run these tests locally.

I wonder if we should have the .jar file as part of our repo.

package.json
drewfish, line 62: I agree, so long as travis is running the full test suite (func as well as unit).

drewfish:
False-positive?  I'm seeing the following in job 157.1:

```
1217TypeError: Cannot call method 'createElement' of null
1218 at /home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/node_modules/yui/dom-core/dom-core-min.js:7:262
...
1228[2012-09-06 07:25:37.570] [ERROR] NodeDriver - Failed to get the test report
1229[2012-09-06 07:25:37.572] [ERROR] Arrow - Failed to get the test report, Controller: /home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/lib/controller/default
```

package.json
caridy, line 62: Yes, I think unit tests is just fine, and whoever wants to run functional tests, they should be able to do it manually.

.travis.yml
drewfish: It might be nice to also include a link to the docs (http://about.travis-ci.org/docs/user/gui-and-headless-browsers/). That might make maintenance easier in the future should the docs change.

drewfish:
In the func tests of job 157.1, there are a bunch of descriptors that don't seem to be running any tests. For example, line 1601: `0 Passed, 0 Failed , 0 skipped`. I don't know if this is OK or not.

.travis.yml
mridgway: @caridy I can refactor the travis.yml into scripts as you requested. I'm not a fan of having the .jar as part of the repo.

mridgway:
@isao is looking into the createElement errors. These are created by the new middleware tests.

The `0 Passed, 0 Failed , 0 skipped` is a result of having empty test files. During the transition we created placeholder tests that instantiate the components so that we can easily add tests. This way, if we start doing code coverage, the files still have some coverage and are included in the report.

.travis.yml
mridgway: Thinking about this some more, I don't know if it is necessary to refactor this out. Some of these steps like `xvfb` and the display port I would not expect anyone to run outside of the Travis environment. I think the other three steps would just be a documentation issue. Given that, I don't know what advantage we get by moving these into separate files.

isao:
@mridgway pull #468 should fix the `createElement` error-- caused by mixing nodejs code with browserland

.travis.yml
caridy: @mridgway the main reason for refactoring this is because we could have a dummy before_install script, and a custom one for our bleeding branches (yaf-develop for example) to use the bleeding YUI version (branch 3.x). Otherwise we will have to have a different .travis.yml file in those branches.

mridgway:
All tests are now passing. I consider this ready to be merged. There is still some clean up work to be done separately and we'll have to monitor stability as we go.

isao:
shipit

thanks @mridgway

caridy:
@mridgway thanks for the refactor on travis.yml :)

PR #473, opened 2012-09-10
fix #470, revert #456, reopen #438

### LINKED ISSUE #470: instanceId is not populated for child mojits in client side composites
This is a regression identified between 0.4.3 and the current develop branch. This breaks client-side broadcasting when trying to target a mojit's child.

drewfish:
fix #470, revert #456, reopen #438

Revert "Merge pull request #456 from drewfish/gh438-rs-adapter"

This reverts commit 82ab5af974067d010f4eea2e723a8780d4cbd7d9, reversing
changes made to 755806560eb66d8515544e1fb053c7d692dd950d.

Conflicts:

```
lib/app/autoload/store.client.js
```

### DIFF
--- a/lib/app/autoload/mojito-client.client.js
+++ b/lib/app/autoload/mojito-client.client.js
@@ -727,7 +727,7 @@ YUI.add('mojito-client', function(Y, NAME) {
 
         doRender: function(mp, data, view, cb) {
             if (!mp._views || !mp._assetsRoot) {
-                this.resourceStore.expandInstanceForEnv('client', { type: mp.type }, mp.context,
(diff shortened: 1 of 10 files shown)


mridgway:
+1 for doing this so we can cut a release.

drewfish:
Note that merging this PR will mean that Shaker2.0 (still) doesn't work with Mojito 0.4.x.

mridgway:
My understanding is that a fix for this won't be ready for when we want to do a release. If we can fix this soon after that release, I don't see a reason not to do another follow up release.

PR #474, opened 2012-09-10
Performance metrics in mojito

caridy:
Getting perf.server.js as expected. Adding new configuration settings to enable and disable some of the metrics:

```
[{
    "settings": [ "master" ],
    "specs": {},
    "perf": {
        "timeline": true,
        "mark": true
    }
}]
```

With these new settings, we can get some useful metrics through the Node console. We can also instrument any mojit or any mojito-core related process by using `mark` or `timeline` methods. Here is an example:

```
YUI.add('something', function(Y, NAME) {

    // mark an important moment in the request life-cycle
    Y.mojito.perf.mark(NAME, 'metric-name',
            'metric description', 'uniqueId if needed');

    // open a timeline metric if you want to measure a process
    var perf = Y.mojito.perf.timeline(NAME, 'metric-name',
            'metric description', 'uniqueId if needed');

    // at some point in the future, you can call "done"
    // to mark the end of the timeline.
    perf.done();

}, '0.1.0', {requires: [
    'mojito-perf',
]});
```

As a result, we will get something like this in the console:

```
[MOJITO] (11) mojito-perf: something:metricName[id] mark=120 (metric description)
[MOJITO] (12) mojito-perf: something:metricName[id] timeline=300 (metric description)
[MOJITO] (14) mojito-perf: mojito:mojito-perf:express:timeline[1] timeline=420 (time to respond)
```

### DIFF
--- a/lib/app/addons/ac/composite.common.js
+++ b/lib/app/addons/ac/composite.common.js
@@ -19,12 +19,14 @@ YUI.add('mojito-composite-addon', function(Y, NAME) {
         this.id = id;
         this.callback = callback;
         this.__meta = [];
+        this.__perf = Y.mojito.perf.timeline(NAME, 'child:done', 'time to render a child mojit', id);
     }
 
 
     AdapterBuffer.prototype = {
 
         done: function(data, meta) {
+            this.__perf.done();
             this.buffer[this.id].meta = meta;
             this.buffer[this.id].data += data;
             this.buffer.__counter__ -= 1;
@@ -202,7 +204,8 @@ callback({
             var ac = this.ac,
                 buffer = {},
                 content = {},
-                meta = {};
+                meta = {},
+                perf;
 
             cfg.children = cfg.children || {};
 
@@ -211,6 +214,8 @@ callback({
                 throw new Error('Cannot process children in the format of an' +
                                 ' array. \'children\' must be an object.');
             }
+
+            perf = Y.mojito.perf.timeline(NAME, 'execute', Y.Object.keys(cfg.children), cfg.id);
 
             // check to ensure children doesn't have a null child
             // in which case it will be automatically discarded to
@@ -249,6 +254,8 @@ callback({
                         }
                         ac.assets.mixAssets(meta.assets, cfg.assets);
                     }
+
+                    perf.done();
 
                     cb(content, meta);
                 });
--- a/lib/app/addons/ac/output-adapter.common.js
+++ b/lib/app/addons/ac/output-adapter.common.js
@@ -214,8 +214,7 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
 
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_end[' + instance.type +
-                ':' + action + ']', 'ac.done() completed for Mojit "' +
-                instance.type + '" with action "' + action + '"');
+                ':' + action + ']', 'ac.done() completed', instance.id);
     };
 
 
--- a/lib/app/addons/ac/output-adapter.common.js
+++ b/lib/app/addons/ac/output-adapter.common.js
@@ -49,7 +49,12 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
             renderer = null,
             contentType,
             contentPath,
-            viewEngineOptions = (instance.appConfig && instance.appConfig.viewEngine) || {};
+            viewEngineOptions = (instance.appConfig && instance.appConfig.viewEngine) || {},
+            perf;
+
+        // Perf metric for time to done
+        perf = Y.mojito.perf.timeline(NAME, 'mojit:done[' + instance.type +
+                ':' + action + ']', 'time to execute ac.done process', instance.instanceId);
 
         if (Y.Lang.isString(meta)) {
             // If the meta string is a serializer set it
@@ -212,9 +217,12 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
             adapter[callbackFunc](data, meta);
         }
 
+        // closing mojit:done
+        perf.done();
+
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_end[' + instance.type +
-                ':' + action + ']', 'ac.done() completed', instance.id);
+                ':' + action + ']', 'ac.done() completed', instance.instanceId);
     };
 
 
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -251,7 +251,8 @@ YUI.add('mojito-action-context', function(Y, NAME) {
             adapter = opts.adapter,
             store = opts.store,
             actionFunction,
-            error;
+            error,
+            perf;
 
         // "init" is not an action
         if (command.action === 'init') {
@@ -268,6 +269,9 @@ YUI.add('mojito-action-context', function(Y, NAME) {
         this.type = instance.type;
         this.context = command.context;
         this.models = models;
+
+        perf = Y.mojito.perf.timeline(NAME, 'ac:init',
+                'time to set up AC object', instance.instanceId);
 
         // identify this as internal... users probably won't want to use it, but
         // addons might need
@@ -312,14 +316,23 @@ YUI.add('mojito-action-context', function(Y, NAME) {
             }
         }
 
+        // closing the perf metric for ac:init
+        perf.done();
+
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_start[' + instance.type +
-            ':' + command.action + ']', 'Calling Mojit', instance.id);
+            ':' + command.action + ']', 'Calling Mojit', instance.instanceId);
 
         Y.log('action context created, executing action "' + actionFunction +
             '"', 'mojito', 'qeperf');
 
+        perf = Y.mojito.perf.timeline(NAME, 'ac:action:invoke',
+                'time to call an action from AC', instance.instanceId);
+
         controller[actionFunction](this);
+
+        // closing the perf metric for ac:action:invoke
+        perf.done();
     }
 
     Y.namespace('mojito').ActionContext = ActionContext;
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -314,8 +314,7 @@ YUI.add('mojito-action-context', function(Y, NAME) {
 
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_start[' + instance.type +
-            ':' + command.action + ']', 'Calling the Mojit "' + instance.type +
-            '" with action "' + command.action + '"');
+            ':' + command.action + ']', 'Calling Mojit', instance.id);
 
         Y.log('action context created, executing action "' + actionFunction +
             '"', 'mojito', 'qeperf');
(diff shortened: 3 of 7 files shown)


caridy:
Client side metrics are coming next, plus more mojito-core marks and timelines.

lib/app/addons/ac/output-adapter.common.js
```diff
@@ -214,8 +214,7 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
 
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_end[' + instance.type +
-                ':' + action + ']', 'ac.done() completed for Mojit "' +
-                instance.type + '" with action "' + action + '"');
+                ':' + action + ']', 'ac.done() completed', instance.id);
```
drewfish: Note that not all instances have IDs -- only those that come from the top-level "spec" in application.json will. It might be better to use instance.instanceID instead.

lib/app/autoload/action-context.common.js
```diff
@@ -314,8 +314,7 @@ YUI.add('mojito-action-context', function(Y, NAME) {
 
         // Time marker
         Y.mojito.perf.mark('mojito', 'core_action_start[' + instance.type +
```
drewfish, line 323: Shouldn't the "core_action_start" and "core_action_end" marks be a timeline instead?

lib/app/autoload/dispatch.common.js
```diff
@@ -148,7 +148,9 @@ YUI.add('mojito-dispatcher', function(Y, NAME) {
 
                 function runMojit() {
                     var moduleList,
-                        mojitYuiModules;
+                        mojitYuiModules,
+                        perf = Y.mojito.perf.timeline(NAME, 'mojit:use',
```
drewfish, line 159: This should probably be `mojit:dispatch` instead.  Inventing new terminology will only add confusion, in my opinion.

lib/app/autoload/perf.server.js
```diff
@@ -7,119 +7,212 @@
...
+    if (!YUI._mojito) {
+        YUI._mojito = {};
+    }
+
+
+    var store = YUI._mojito._perf,
+        perfConfig = Y.config.perf || {},
+        requestId = 0,
+        colorRed   = '\u001b[31m',
+        colorReset = '\u001b[0m',
```
drewfish, line 27: Having VT100 terminal colors worries me.  There are times when we _shouldn't_ be showing colors (such as logging in Manhattan). I'd prefer no colors, to naive color implementation that doesn't take into consideration the execution environment.

Also, calling this out using colors assumes that that is what the user wishes to look at.  Right now, that makes sense since we're focusing on performance. However, later when our mindset is different, the colors won't be appropriate.

lib/app/autoload/perf.server.js
```diff
@@ -7,119 +7,212 @@
...
+        getgo,
+        microtime;
```
drewfish, line 29: We sorta have a tradition in mojito codebase to prefix variables that hold libraries with `lib`, as in `libpath = require('path')`. This is to prevent collision between variables names that make sense for the current algorithm and the libraries.

That idea is definitely not applied consistently throughout mojito thought :disappointed: though I wish it was.

lib/app/autoload/perf.server.js
```diff
@@ -7,119 +7,212 @@
...
+     * This method is responsible for calling "dump".
+     *
+     * @method instrumentMojitoRequest
+     * @param {object} req the request object from express.
+     * @param {object} res the response object from express.
+     **/
+    function instrumentMojitoRequest(req, res) {
+        var id = (requestId += 1),
+            perf,
+            end = res.end;
```
drewfish, line 192: Is monkeypatching res.end the best way to accomplish this? Does the "req" object have an event we can hook into?

lib/app/autoload/perf.server.js
```diff
@@ -7,119 +7,212 @@
...
+                    'friend, use it.', 'warn', NAME);
         }
 
-
-        function dump() {
-            var len = buffer.length,
-                i;
-            for (i = 0; i < len; i += 1) {
-                Y.log(buffer[i], 'mojito', NAME);
+        perf = timeline('mojito', NAME + ':express:timeline', 'time to respond', id);
```
drewfish, line 201: The "express" confuses me? If this is the timeline for the entire request, I would prefer to see "request" or something like that.  Also, since this _is_ a timeline, why the ":timeline"?

lib/index.js
```diff
@@ -188,9 +188,19 @@ MojitoServer.prototype = {
 
         configureYUI(Y, store, CORE_MOJITO_MODULES);
 
+        // in case we want to collect some performance metrics,
+        // we can do that by setting:
+        // application.json (master) perf->metricName = true
+        YUI.applyConfig({
+            perf: appConfig.perf
```
drewfish, line 195: The YUI config object doesn't have a "perf" key, so why are we setting this here?  http://yuilibrary.com/yui/docs/api/classes/config.html

If we're going to pollute the YUI config object with our own configuration details, we should at least namespace it to our library, as in: `YUI.applyConfig({ mojito: { perf: ... }});`.

lib/index.js
```diff
@@ -265,6 +273,11 @@ MojitoServer.prototype = {
 
             logger.log('START', 'mojito', 'server');
 
+            // if perf metrics are on, we should hook into
+            // the mojito request to flush metrics when
+            // the connection is closed.
+            Y.mojito.perf.instrumentMojitoRequest(req, res);
```
drewfish, line 279: Instrumenting each request does look a lot nicer than instrumenting the whole server.  By doing this, we can be sure that performance metrics for each request are kept separate from other requests.

drewfish, line 279: Oh... my expectation is for this to look something like this:  `Y.mojito.perf.instrumentMojitoRequest(appConfig.perf, req, res);`

lib/app/addons/ac/composite.common.js
```diff
@@ -19,12 +19,14 @@ YUI.add('mojito-composite-addon', function(Y, NAME) {
         this.id = id;
         this.callback = callback;
         this.__meta = [];
+        this.__perf = Y.mojito.perf.timeline(NAME, 'child:done', 'time to render a child mojit', id);
```
drewfish, line 22: I think "child:done" is the wrong name for this.  The description leads me think the name should just be "child".

Oh, I'm not sure that "render" is the right verb in the description. "Render" already means something very particular in Mojito.  Perhaps "time to execute a child mojit" or "time to dispatch a child mojit".

lib/app/addons/ac/composite.common.js
```diff
@@ -212,6 +215,8 @@ callback({
                                 ' array. \'children\' must be an object.');
             }
 
+            perf = Y.mojito.perf.timeline(NAME, 'execute', Y.Object.keys(cfg.children), cfg.id);
```
drewfish, line 218: I don't thinks `cfg.id` contains what you think it does.

lib/app/autoload/perf.server.js
```diff
@@ -7,119 +7,212 @@
...
+                done: function () {}
+            };
+        },
 
+        mark: perfConfig.mark ? mark : function () {},
 
-            mark: perfEnabled ?
-                    mark :
-                    function() {},
+        dump: perfConfig ? dump : function () {}
```
drewfish, line 234: Why are we exporting `dump`?  This appears to be an internal method.

drewfish:
I think the perf implementation could be cleaned up a lot.  For example, that `store` variable is worrying. I'd prefer to see a single object, created for each request, that contains the details for just that request.

lib/app/autoload/action-context.common.js
caridy, line 323: I think so. There are few "mark" actions as part of the legacy that need to be reviewed. Marks should be really important moments rather than start/end, you're right. I will start changing some of them.

lib/app/autoload/dispatch.common.js
caridy, line 159: Hmm, I wanted to measure the "use" routine, and I want to measure the whole "dispatch" process separately.

caridy:
Yes, more clean up is definitely needed, documentation, and client side, will work on that today. "store" for example is a legacy stuff, and a bunch of other things. And eventually adding hierarchical perf metrics.

lib/app/autoload/dispatch.common.js
drewfish, line 159: OK, it should maybe be called "Y.use" instead then.

lib/app/autoload/perf.server.js
caridy, line 192: The previous approach was trying to get in the way by doing app.on('connection'), then adding the "end" event. Apparently, it didn't play nicely with mojito. In many cases, specially when using a browser to hit the page, the connection never reach the close or end point, but using curl does it. I guess it was the way mojito is ending that connection.

After moving it into this low level hooking, everything works perfectly.

lib/index.js
caridy, line 195: I was thinking to use the same at the server and client (eventually consolidate perf module since it does exactly the same), and at the client there is not just thing as YUI_config.mojito.<something>, maybe this is a good time to add it ;)

lib/index.js
caridy, line 279: There are few problems I guess:
- mojito-perf might be included at a mojit level to do perf, in which case that configuration needs to be available, in which case we don't call instrumentMojitoRequest, but other settings are there.
- at some point, we might want to instrument few things but not the MojitoRequest, in which case we need the config anyone.
- In the client side, there is not such thing as instrumentMojitoRequest, and the config has to be available to be used.

In my mind, Y.config.perf or Y.config.mojito.perf is the right thing to use to unify the implementation.

lib/app/addons/ac/composite.common.js
caridy, line 22: agreed.

lib/app/addons/ac/composite.common.js
caridy, line 218: You're right, it doesn't add any ID into the log. I'm getting this:

mojito-composite-addon:execute timeline=634 (slot-1,slot-2,slot-3,slot-4,slot-5,slot-6,slot-7,slot-8,slot-9,slot-10,slot-11,slot-12,slot-13,slot-14,slot-15,slot-16,slot-17,slot-18,slot-19,slot-20,slot-21,slot-22,slot-23,slot-24,slot-25,slot-26,slot-27,slot-28,slot-29,slot-30,slot-31,slot-32,slot-33,slot-34,slot-35,slot-36,slot-37,slot-38,slot-39,slot-40,slot-41,slot-42,slot-43,slot-44,slot-45,slot-46,slot-47,slot-48,slot-49,slot-50)

and I should be getting: mojito-composite-addon:execute[ID-HERE]. What will be the right way to collect the instanceId in this case?

lib/app/autoload/perf.server.js
caridy, line 234: Few reasons:
- in the short-term, we want "dump" to produce a json for the SVG.
- at some point, someone might want to dump data programmatically to collect the state of the execution. Imagine a mojit that runs last, and print the perf info in the page.
- it was exposed before, not so sure why :)

caridy:
closing this one in favor of #478

PR #475, opened 2012-09-10
Fixed tunnel handler arrow test

mridgway:
@folta Looks like a unit test that was missed in #473


mridgway:
Travis failure is mu test again.

PR #479, opened 2012-09-11
Cleaned up functional tests and in some cases made them more stable

mridgway:
- Cleaned up tabs/spaces in descriptors
- Added test identifiers to suite names so that we can differentiate successful tests
- Removed duplicate yui-tests files and fixed references in descriptors
- Made a couple tests more stable


mridgway:
Looks like there is a bug in Arrow with 0.6. I'm currently debugging this.

mridgway:
I submitted a PR to Arrow: https://github.com/yahoo/arrow/pull/14

This build should fix the internal builds and the issues some of our local test runs were having.

drewfish:
huge changeset, only did a quick review.  +1 to changes I saw.

+1 to four bullet points above.

PR #480, opened 2012-09-11
adding cache for getAppConfig function (resourceStore)

diervo:


lib/store.server.js
```diff
@@ -315,10 +320,11 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             ycb = this.config.readConfigYCB(this._libs.path.join(this._config.root, 'application.json'), ctx);
             this.mergeRecursive(appConfig, ycb);
 
-            // apply the passed-in overrides
-            this.mergeRecursive(appConfig, Y.mojito.util.copy(this._config.appConfig));
+            //NOTE edge case, eventually we may remove this as well??
+            // who use this? where is coming from?
+            this.mergeRecursive(appConfig, this._config.appConfig);
```
drewfish, line 325: `this._config.appConfig` contains the appConfig overrides that were sent to the ResourceStore constructor. I believe that (at least) one command is using this.

caridy:
@dferreiroval let's try to find out who is augmenting it then.

Also, rather than freeze the object when returning it, you can freeze it once when storing it in cache, in case we decide to freeze it.

PR #481, opened 2012-09-11
controller only gets what is -explicitely- listed in it's "requires"

drewfish:
This was uncovered by @dferreiroval while working on mojito-shaker 2.0.


caridy:
:shipit:

PR #482, opened 2012-09-12
improve error handling and messaging

FabianFrank:
notify the user with more meaningful messages when trying to run mojits that don't exist or are not configured correctly.


FabianFrank:
This also solves a production problem for us, where we might see an error log like
[ERROR](1347417986650) Server OutputHandler: Error 500: Cannot read property 'url' of undefined

but don't know which mojit should have been executed and therefore don't know what might be going on. After this PR we'll see the name of the Mojit that failed:
[ERROR](1347418062693) Server OutputHandler: Error 500: Unable to compute assets root for mojit "FooBar". This usually happens when trying to run a mojit that does not exist.

isao:
+1 though not sure why a store unit test is failing on Travis (node .8 only)

FabianFrank:
@isao it seems we are good to merge ;-)

PR #483, opened 2012-09-12
perf: adding store:getMojitTypeDetails metric

caridy:


caridy:
We need to corroborate if we really need to make any change in the cache behavior for this method.

lib/store.server.js
```diff
@@ -563,9 +564,14 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             delete spec.appConfig.specs;
 
             try {
+                perf = Y.mojito.perf.timeline('mojito', 'store:getMojitTypeDetails',
+                    'expand mojit spec', spec.type);
```
drewfish, line 568: The description isn't correct. It should be something like "get details for the mojit" or "expand mojit details".

PR #484, opened 2012-09-12
adding cache for getRoutes function (resourceStore)

dmorilha:


caridy:
:shipit:

lib/store.server.js
```diff
@@ -740,17 +757,21 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+                    Y.mix(out, routes, true);
                 }
             }
 
             if (!Object.keys(out).length) {
-                return this._fwConfig.defaultRoutes;
+                Y.mix(out, this._fwConfig.defaultRoutes, true);
             }
+
+            this._routesCache[key] = Object.freeze(out);
```
FabianFrank, line 773: @dmorilha Is there a particular reason for using freeze? According to http://jsperf.com/freeze-vs-seal-vs-normal/3 frozen/sealed objects are a lot slower handled by v8 than "normal" objects.

caridy, line 773: The option are Copy or Freeze, and we are expecting Object.freeze to become more performance in the upcoming V8. We can't allow the original object to be modified otherwise we will be bleeding between requests, or even within the same request between mojits.

caridy, line 773: One more note on this one: you freeze once, you _have_ to copy everytime, so this is not apple to apple.

FabianFrank, line 773: The problem is not about the freeze call itself, but about the performance hit that appears whenever doing something with an object that has been frozen. In my mind the questions is, what is faster:
1. a request that operates on a frozen object
2. a request that does a clone and then operates on a regular object

i.e. is the clone overhead worse than the overhead of operating on a frozen object?

caridy, line 773: agreed, raising the concern in the board right now! thanks

PR #485, opened 2012-09-12
perf: adding support for perf.log

caridy:
introducing perf->log to write to disk.

```
"perf": {
    "timeline": true,
    "mark": true,
    "log": true,
    "exclude": {
        "groupName": true
    }
}
```

As a result, perf metrics will be log into disk under `./artifacts/perf/request-###.log`, where each line will have:

```
TIMELINE | timestamp:duration | group | label | id | desc
```

or

```
MARK | timestamp:duration | group | label | id | desc
```

Here is an example:

```
TIMELINE|1347418448352586:3947|mojito|ac.done|master.index|time to execute ac.done process
MARK|1347418448352593:795666|mojito|action:stop|master.index|after the action
```

### DIFF
--- a/lib/app/autoload/perf.server.js
+++ b/lib/app/autoload/perf.server.js
@@ -20,9 +20,9 @@ YUI.add('mojito-perf', function (Y, NAME) {
     }
 
 
-    var fs = require('fs'),
-        path = require('path'),
-        existsSync = fs.existsSync || path.existsSync,
+    var libfs = require('fs'),
+        libpath = require('path'),
+        existsSync = libfs.existsSync || libpath.existsSync,
         buffer     = YUI._mojito._perf,
         perfConfig = Y.config.perf || {},
         requestId  = 0,
@@ -46,7 +46,7 @@ YUI.add('mojito-perf', function (Y, NAME) {
             i;
 
         try {
-            outstream = fs.createWriteStream(filename, {
+            outstream = libfs.createWriteStream(filename, {
                 flags: 'a' // append
             });
             for (i = 0; i < logs.length; i += 1) {
@@ -322,7 +322,7 @@ YUI.add('mojito-perf', function (Y, NAME) {
     if (perfConfig.logFile && existsSync(perfConfig.logFile)) {
         Y.log("Removing the previous perf log file '" +
             perfConfig.logFile + "'", 'warn', NAME);
-        fs.unlinkSync(perfConfig.logFile);
+        libfs.unlinkSync(perfConfig.logFile);
     }
 
 });
--- a/lib/app/autoload/perf.server.js
+++ b/lib/app/autoload/perf.server.js
@@ -68,18 +68,21 @@ YUI.add('mojito-perf', function (Y, NAME) {
             // if we already have milliseconds, good
             // if not, we can compute it based on request init
             time = o.time,
-            duration = (o.ms || (o.time - getgo)),
+            offset = o.time - getgo,
+            duration = o.ms || '',
             desc = o.msg || 'no description',
             label = o.label,
             id = o.id;
 
         if ((perfConfig.mark && !o.ms) || (perfConfig.timeline && o.ms)) {
 
-            Y.log(group + ':' + key + ' ' + type + '=' +
-                colorRed + duration + colorReset + ' (' + desc + ')',
+            Y.log(group + ':' + key + ' ' + type + colorReset +
+                ' offset=' + colorRed + offset + colorReset +
+                (o.ms ? ' duration=' + colorRed + duration + colorReset : '') +
+                ' (' + desc + ')',
                 'mojito', NAME);
 
-            return [type, time + ':' + duration, group, label, id, desc];
+            return [type, requestId, time, duration, group, label, id, desc];
 
         }
 
@@ -216,7 +219,7 @@ YUI.add('mojito-perf', function (Y, NAME) {
      *
      * @method dump
      * @return {array} collection of perf logs. Each item will expose:
-     *     {type, time + ':' + duration, group, label, id, desc}
+     *     {type, requestId, time, duration, group, label, id, desc}
      **/
     function dump() {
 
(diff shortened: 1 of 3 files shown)


lib/app/autoload/perf.server.js
```diff
@@ -70,19 +125,22 @@ YUI.add('mojito-perf', function (Y, NAME) {
...
+        var s,
+            key = label;
 
         if (!group || !label) {
             return;
         }
 
         if (id) {
-            label += '[' + id + ']';
+            // we might also accept a command object
```
drewfish, line 144: This comment is in the wrong place.  It should be part of the API docs!

drewfish:
I'd prefer instead if there was a `perf.logFile` setting, which gives the logfile name. (It is still put in `artifacts/perf/`.)

I think the line format should be:
`type | requesstID | start | type-specific | group | label | id | desc`
The `type-specific` is different for each type:  for `MARK` it is empty and for `TIMELINE` it is the duration.

drewfish:
I think having multiple files will make the analysis _harder_.  Having one file per "run" (where each run is multiple requests, perhaps each run being a separate version of mojito) seems the natural way to group things to me.

lib/app/autoload/perf.server.js
caridy, line 144: It is, line 128:

```
 * @param {string|object} id Unique identifier of the mark, usually
 *      the requestId or a command object.
```

caridy:
About the perf->logFile definition, I'm getting to think that this feature is more likely to be triggered from the command line, something like:

```
  mojito start --perf path/to/file.log
```

and by default log on ./artifacts/perf/<something>.log

It seems more natural, because you don't have to change your application.json to get a perf result. And keeping application.json->perf as the general configuration.

About unifying the results in one file, I'm not convinced, but that's Ok, I will change it.

drewfish:
Yeah, the `--perf path/to/file.log` looks nice.

drewfish:
Default should be `artifacts/perf/perf.log`, or perhaps `../log`.

mojit0:
On Sep 12, 2012, at 10:58 AM, Caridy Patino wrote:

About the perf->logFile definition, I'm getting to think that this feature is more likely to be triggered from the command line, something like:

  mojito start --perf path/to/file.log

and by default log on ./artifacts/perf/.log

+1

caridy:
Added the --perf path/to/file.log, working on the new format now.

Q: should we always append to the file? should we reset the file content on start? the catch here is that on start, we set requestId = 0, if we don't reset the file, we might ended up some clobbered data. @drewfish what do you think?

drewfish:
Good question.  My first thought is that we should append, but you're right about the requestId.

One idea:  a separate file for every server-start (filename has start timestamp).  I generally wish to avoid creating a huge number of files, but this might be a good compromise.

caridy:
The other thing I was thinking is that if you're running $ mojito start --perf something.log and you plan to run it multiple times, your script can generate the filename, rather that mojito trying to adjust it. The main problem with adjusting from mojito is that tools that are expecting a particular files will have to understand the logic, which defeats the purpose of the logFile definition. So, let's stick to the plan, resetting the file if exists, and let's move on.

drewfish:
OK

caridy:
Alright, done with it. @drewfish please review.

lib/app/autoload/perf.server.js
```diff
@@ -20,7 +20,10 @@ YUI.add('mojito-perf', function (Y, NAME) {
     }
 
 
-    var buffer     = YUI._mojito._perf,
+    var fs = require('fs'),
+        path = require('path'),
```
drewfish: It's safer to use `libfs` and `libpath` for these variable names, so that they don't get clobbered.

lib/app/autoload/perf.server.js
```diff
@@ -93,14 +154,17 @@ YUI.add('mojito-perf', function (Y, NAME) {
             msg = '';
         }
 
-        if (buffer[group][label]) {
-            Y.log('Perf metric collision for group=' + group + ' label=' + label +
+        if (buffer[group][key]) {
+            Y.log('Perf metric collision for group=' + group +
+                ' label=' + label + ' id=' + id +
                 '. Measure one thing at a time.', 'warn', NAME);
```
drewfish, line 160: I think it's completely reasonable use-case to have multiple perf marks/timelines with the same [group,label,id], so I don't think this should be a warning.

drewfish, line 160: I guess I imagine something like this instead:

``` javascript
mark = {};
mark.group = group;
mark.label = label;
...
buffer[group].push(mark);
return mark;
```

lib/index.js
```diff
@@ -189,11 +190,17 @@ MojitoServer.prototype = {
...
         // in case we want to collect some performance metrics,
-        // we can do that by setting:
-        // application.json (master) perf->metricName = true
-        YUI.applyConfig({
-            perf: appConfig.perf
-        });
+        // we can do that by defining the "perf" object in:
+        // application.json (master)
+        // You can also use the option --perf path/filename.log when
+        // running mojito start to dump metrics to disk.
```
drewfish, line 196: This documentation is in the wrong place. Most of it isn't necessary, since it's already documented in the `mojito help start`. The fact that `options` can have a `perf` should be in the API docs for addMojitoToExpressApp().  Hmm... none of the other options.\* are documented today, so you are forgiven :smile: .

drewfish:
Looks good pretty good, though some cleanups are needed. We could merge it into `search-perf` now, and cleanup when `search-perf` merges to `develop`. (`search-perf`, in general, needs a bunch of cleanups before hitting `develop`.)

lib/app/autoload/perf.server.js
caridy, line 160: Well, there are other implications on this one. Can we merge this and later on discuss how are we going to deal with collisions and multiple requests? /cc @drewfish

drewfish, line 160: Sure.

All I'm trying to say is that, if all marks/timelines for a request are separate from other requests, then there is no such thing as a "collision".  I should be able to do the following in my code, without warnings:

``` javascript
for (...) {
    ...code...
    Y.mojito.perf.mark('libfoo', 'got here');
    ...code...

}
```

PR #486, opened 2012-09-12
Removed unneeded YUI instances...

jlecomte:
Removed unneeded YUI instances. This reduces the process memory footprint by 4 MB, from 27 MB to 23 MB at the time the first request is handled.


FabianFrank:
+1

PR #124, opened 2012-04-10
[bz5749285] Computed route caching

epuidokas:
#75

PR #489, opened 2012-09-12
Fixes to middleware (contextualizer, router) and route-maker

imalberto:
Fixes to middleware (contextualizer, router); caching routes in routes-maker for a given context so that it's not calculated on every request


mridgway:
Does this supercede #124? This is a really outdated PR that needs to be closed if we're not going to merge it.

imalberto:
@mridgway Looks like it. This will first be merged with "search-perf" and hopefully "develop" down the line.

mridgway:
Sounds good. I'm going to close #124 in that case.

Issue #461, opened 2012-09-05
[bz5815270] build html5app command - the "builds.html5app" key in "application.json" is not context aware.


ricallinson:
When running the "mojito build html5app" command, the "builds.html5app" key in "application.json" is not changed when a context is provided.

rwaldura:
@jenny mentioned in SoS

ricallinson:
Pull request here https://github.com/yahoo/mojito/pull/466

rwaldura:
See #466.

rwaldura:
please update bz ticket 5815270 when you close this, thanks!

isao:
done, thank you @rwaldura

rwaldura:
this issue was resolved by applying PR #490

PR #490, opened 2012-09-12
fix issue #461 builds.html5app key was not context aware

isao:
this is a cleanup of Ric's pull #466
thanks @capecodehq


drewfish:
+1

PR #492, opened 2012-09-12
removed unnecessary validateContext

diervo:
Changed some of them to validate only when no cache is avaliable and removed unnecessary ones


drewfish:
+1

Issue #13, opened 2012-03-30
[bz5147315] ContextController cache never cleans up


add0n:
Y instances are cached within the client controller context object by mojit instance guid so that many actions on the same controller can always occur on the same instance.

If all the views of a particular mojit controller are destroyed, that Y instance is never removed.

This is a possible memory leak.

Every time a mojit proxy's private _destroy() function is called, the client needs to do a check to see if there are any more mojit proxies that exist tied to that same instance guid. If not, it should notify the ControllerContext to clear the cache for that guid, and it should dereference the cached Y instance.

PR #495, opened 2012-09-13
fix issue #13 ContextController cache never cleans up

isao:
- add method to dispatch.common to enable deleting cached
  controller contexts (cc caching occurs on the client only)
- check if cache cleanup needed during mojito-client destroyMojitProxy()
- other minor


drewfish:
In the "develop-perf" branch (which perhaps will become mojito@0.5), the ControllerContext is no longer used.

isao:
yes, I know. should we not plug this client-side memory leak in the meantime?

rwaldura:
yes we should

PR #497, opened 2012-09-13
Fixes based on Ric's changes

imalberto:
Please do Ric PR first before applying this one


rwaldura:
I believe @imalberto is referring to #494 by @capecodehq .

imalberto:
@rwaldura Thanks for adding the PR# !

PR #501, opened 2012-09-13
Bug fixes in simple examples replacing "mojit_guid" with "mojit_view_id"

ricallinson:


isao:
good catch
+1

isao:
test and lint errors coming from your perf branch. these need to be applied to/from yahoo/develop

drewfish:
I'm going to merge onto search-perf, since it's needed there too.

isao:
ok, but it has nothing to do w/ perf (they're basically non-functional changes), and that branch won't hit develop for a while..

caridy:
@isao I still see a bunch of them on `tests/fixtures/` when doing a grep.

PR #502, opened 2012-09-13
use an AC addon for accessing the config

drewfish:
ac.instance and ac.command will be going away, so there needs to be
a new way to get the app-level configs


caridy:
+1

PR #504, opened 2012-09-13
re-enabling the perf metrics

caridy:
also:
- removing the unnecessary cache on the controller object.
- cleaning up some qaperf comments.


drewfish:
lgtm.  I will merge after team is done doing a perf run.

PR #508, opened 2012-09-13
better error message for javascript parse errors

drewfish:


caridy:
+1

isao:
nice

lib/app/addons/rs/yui.server.js
```diff
@@ -626,7 +626,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                 libvm.runInNewContext(file, ctx, res.source.fs.fullPath);
             } catch (e) {
                 yui = null;
-                Y.log(e.message + '\n' + e.stack, 'error', NAME);
+                Y.log('failed to parse javascript file ' + res.source.fs.fullPath, 'error', NAME);
```
FabianFrank, line 629: @drewfish doesn't the stack contain the parsing error and line number?

drewfish, line 629: The stack doesn't.  I think the e.message contains the error, so yeah that should be in the output as well.

PR #513, opened 2012-09-14
Added "deepFreeze" and "copy" for now

imalberto:
The goal is to have the "copy" removed from expandInstanceForEnv at some point.


imalberto:
@caridy Can you please review so that we can merge into search-perf ?

caridy:
@imalberto do we need this after all? if yes, can you clean it?

imalberto:
I'll close this PR. deepFreeze cost is too high. Will handle the loading of ac.addons in a separate PR.

PR #514, opened 2012-09-14
Search perf

dmorilha:
9/12 by 6:30pm we discovered we should consider a subset of the actual instance as part of the cache key. otherwise things from the request may get included (post parameters, headers, get parameters, etc...)

### DIFF
--- a/lib/store.server.js
+++ b/lib/store.server.js
@@ -548,6 +548,26 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 config,
                 perf;
 
+            console.log();
+            console.log();
+            console.log();
+            console.log();
+            console.log();
+            console.log(cacheKey);
+
+            /*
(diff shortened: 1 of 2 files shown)


lib/store.server.js
```diff
@@ -541,7 +541,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          */
         expandInstanceForEnv: function(env, instance, ctx, cb) {
 
-            var cacheKey = JSON.stringify([instance, ctx]),
+            var cacheKey = JSON.stringify([instance.action, instance.base, action.id, action.type, ctx]),
```
imalberto, line 544: instance.base might be undefined ?
"action" is not declared or set within the scope;

caridy:
@dmorilha @imalberto do we really need this after all? if yes, can you clean it?

imalberto:
This PR has been superceded by PR#517. @dmorilha , can you please close this PR ?

PR #518, opened 2012-09-14
enhancement for perf and profiler

caridy:
- adding support for --perf filename-{req}.log
- adding support for mojito profiler --input filename-1.log


imalberto:
+1

PR #519, opened 2012-09-14
Adds a --path option to run.js to allow specifying a path to find the te...

mridgway:
...st descriptors or applications

`./tests/run.js test -u --path ../node_modules/my_mojito_module/tests/unit`

This also changes the location of the generated arrowreports to reside in the path given.


lzhan:
lgtm

PR #522, opened 2012-09-18
fixed multiple typos in the code example as well as typo in explanation ...

dmitris:
The code in the third code example of http://developer.yahoo.com/cocktails/mojito/docs/code_exs/route_config.html has three bugs/typos which prevent it from being parsed as valid JSON.  I added missing commas and quotes so that you could copy and paste the code.  Also fixed the sentence: "you could would use the following to call it"


dmitris:
Here is the code example in question:

``` javascript
[
  {
    "settings": ["master"],
    "custom-route": {
      "verbs": ["get"],
      "path": "/custom-route",
      "call": "mapped_mojit.index"
    }  // missing comma 
    "another-route": {
      "verbs": ["post"]  // missing comma
      "path": "/*",
      "call": mojit-post-route.post_params"  // missing the starting quote
    }
  }
]
```

rwaldura:
Thanks @dmitris !
@zhouyaoji will fix

PR #527, opened 2012-09-18
Fix action-context tests

imalberto:


lib/tests/autoload/app/autoload/action-context-tests.common.js
```diff
@@ -225,12 +231,14 @@ YUI.add('mojito-action-context-tests', function(Y, NAME) {
                         return "routes";
                     }
                 }
+                // models: {},
```
drewfish: Why was this added?

imalberto: This was not added, but commented out from the original tests. But can be safely removed.

lib/app/autoload/action-context.common.js
```diff
@@ -224,7 +225,7 @@ YUI.add('mojito-action-context', function(Y, NAME) {
 
         for (addonName in dependencies) {
             if (dependencies.hasOwnProperty(addonName)) {
-                if (mods.indexOf(addonName + '-addon') > -1 || addonName === 'core') {
+                if (mods.indexOf(addonName) > -1 || addonName === 'core') {
```
drewfish, line 228: Are you sure that this is safe to do? (Yeah, I know this code will be going away.) You might instead try changing the list ['first', 'second', ...] below to ['first-addon', 'second-addon', ...].

imalberto, line 228: I did that initially, but then thought of "search-app" that does not use that convention for their addons. If the norm is "xxx-addon", sure we can proceed the original way.

drewfish, line 228: This change seems dangerous to me, though I'm still trying to work out how.  However, since this code will be going away (soon?), let's just ship it.

drewfish:
+1

PR #528, opened 2012-09-19
adding metric for the handlebars renderer engine

caridy:
Note: using the tmpl path as the ID because we don't have access to the command at this point.

### DIFF
--- a/lib/app/addons/view-engines/hb.server.js
+++ b/lib/app/addons/view-engines/hb.server.js
@@ -39,6 +39,8 @@ YUI.add('mojito-hb', function(Y, NAME) {
          */
         render: function (data, mojitType, tmpl, adapter, meta, more) {
             var cacheTemplates = meta && meta.view && meta.view.cacheTemplates,
+                perf = Y.mojito.perf.timeline('mojito', 'hb:render',
+                    'time to render a template', tmpl),
                 handler = function (err, obj) {
                     var output;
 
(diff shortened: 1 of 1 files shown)


lib/app/addons/view-engines/hb.server.js
```diff
@@ -49,6 +51,8 @@ YUI.add('mojito-hb', function(Y, NAME) {
 
                     output = obj.compiled(data);
 
+                    perf.done(); // closing the 'hb:render' timeline
```
drewfish, line 54: The comment is saying what the code is doing, and so isn't very useful.  I prefer to -not- have these kinds of comments.  (Comments that answer "why?" are much better than those that answer "what?".)

caridy, line 54: sometimes we re-use perf variable for multiple timelines within the same code-path, segmenting the timeline into small pieces (like in "ActionContext" implementation), in which case there are 3 `perf.done();` , and that doesn't tell you which timeline is being closed, I find the name of the timeline to be useful to understand the code better. I do agree that half the sentence can be stashed, but still I want to know which timeline that code should be closing. Maybe we should pass a param that does nothing. something like: `perf.done('hb:render');`, in that case we definitely don't need the comment.

drewfish, line 54: OK yeah, for the re-use of the `perf` variable it does make some sense.  (I've seen other cases where a comment does nothing except say what the next line of code does, which is a bad comment in my opinion.)

drewfish:
+1

PR #531, opened 2012-09-19
fix some store unit tests in develop-perf

drewfish:
Doesn't fix all tests for store-server.

The `validateContext()` test will continue to fail until that method is reimplemented.

One `expanceInstanceForEnv()` test is still failing because `defaults.json` isn't being merged in for some reason.  I'm still looking into that.


caridy:
+1

PR #534, opened 2012-09-19
Cleaned up renderers cache and added unit tests

imalberto:


lib/app/addons/ac/output-adapter.common.js
```diff
@@ -201,23 +200,24 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
...
-            if (rendererCache.hb === undefined || rendererCache.mu === undefined) {
-                rendererCache = {
-                    hb: new (Y.mojito.addons.viewEngines.hb)(''), // viewId
-                    mu: new (Y.mojito.addons.viewEngines.mu)('', {}) // viewId, viewOptions
-                };
+            // optimize for server only
+            if (typeof context.runtime !== undefined && 'server' === context.runtime) {
+                renderer = CACHE.renderers[mojitView.engine];
+                if (!renderer) {
+                    CACHE.renderers[mojitView.engine] = renderer =
```
drewfish, line 209: Does jslint complain about this double assignment?

lib/app/addons/ac/output-adapter.common.js
```diff
@@ -201,23 +200,24 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
...
+                        new (Y.mojito.addons.viewEngines[mojitView.engine])('', {});
```
drewfish: It looks like, on the server, that Y.mojito.ViewRenderer isn't being used. That's slightly inconsistent, but it looks like it'll be OK.

drewfish:
It looks like the view engines are no longer getting `viewEngineOptions` when run on the server.

lib/app/autoload/dispatch.common.js
```diff
@@ -5,7 +5,7 @@
  */
 
 /*jslint anon:true, sloppy:true, nomen:true*/
-/*global YUI*/
+/*global YUI,window*/
```
drewfish: This seems kinda random.  What does app/autoload/dispatch.common.js have to do with view engines?

imalberto:
@drewfish
- nice catch about the viewEngineOptions
- mojito jslint did not complain about the double assignment
- undo the change to dispatch.common. nothing to do with output adapter.

drewfish:
Could you perhaps add a test, to make sure that `viewEngineOptions` doesn't get lost again in the future?

drewfish:
+1

PR #540, opened 2012-09-20
FIX mojito-mu and output-adapter-addon tests

imalberto:


isao:
this would be great on develop as well

caridy:
Do we really need app->config in there? That one is not longer set by default, instead ac.config.getAppConfig() is provided by mojito-config-addon. I think output adapter does not use that config anymore. Otherwise:

+1

drewfish:
+1

imalberto:
@caridy Constructor for action-context still set the app.config object. Can the config addon be accessed from the output-adapter addon ?

PR #541, opened 2012-09-20
adding command.ActionContext support on ControllerContext

dmorilha:
We have identified a scenario where, in certain cases, is necessary to execute arbitrary code after dispatching a mojit but before its controller gets executed. Once the ActionContext is the last mile before the controller, we would like to augment it. The way mojito is today, it is impossible to change the Y.mojito.ActionContext from inside any mojit (and expect those changes on the dispatch). The solution here would be to augment the command object which is passed around with an ActionContext property. It contains a replacement constructor to the standard ActionContext.

### DIFF
--- a/lib/app/autoload/controller-context.common.js
+++ b/lib/app/autoload/controller-context.common.js
@@ -110,7 +110,8 @@ YUI.add('mojito-controller-context', function(Y, NAME) {
                 config = command.instance.config,
                 // this is the action that will be executed
                 action = command.action,
-                ac;
+                ac,
+                o;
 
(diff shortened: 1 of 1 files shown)


caridy:
-1, `controller-context` is going away in 0.5.x.

In top of that, there is nothing that prevent you for redefining `Y.mojito.ActionContext` at any given time, just by requiring a particular YUI module that redefine the original definition across the app.

dmorilha:
very true, i tried before but wasn't able to use like you suggested, I sorted out it was my mistake. Thank you Caridy

PR #545, opened 2012-09-22
FIX route-maker and router-middleware for search-perf

imalberto:
- store.getRoutes are now cached, no need to opt any further
- deprecated querystring routing feature and save a call to store.getAppConfig !
- tests for caching


imalberto:
Commit 777b0f4 updated with feedback from @caridy

caridy:
+1

lib/app/middleware/mojito-router.js
```diff
@@ -43,7 +43,7 @@ Router.prototype = {
     handle: function(store, globalLogger, RouteMakerClass) {
         logger = globalLogger;
 
-        var self = this;
+        var that = this;
```
drewfish: The convention in the mojito codebase is to use `my` (instead of `that`).

imalberto:
commit fea41a5 update from @drewfish feedback

imalberto:
This PR has been replaced by https://github.com/yahoo/mojito/pull/553. Closing.

PR #559, opened 2012-09-24
Removed references to app.config in output-adapter addon

imalberto:
FIX removed references to app.config and replaced with ac.config.getAppConfig(); updated unit tests

### DIFF
--- a/lib/app/addons/ac/output-adapter.common.js
+++ b/lib/app/addons/ac/output-adapter.common.js
@@ -51,8 +51,7 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
             renderer = null,
             contentType,
             contentPath,
-            configAddon = this.config || {},
-            appConfig,
+            appConfig = this.config.getAppConfig(),
             viewEngineOptions,
             perf = Y.mojito.perf.timeline('mojito', 'ac.done',
                 'time to execute ac.done process', this.command);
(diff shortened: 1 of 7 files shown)


imalberto:
PR based on @drewfish and @caridy feedback

lib/app/addons/ac/output-adapter.common.js
```diff
@@ -77,13 +78,14 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
         meta.http.code = meta.http.code || 200;
         meta.http.headers = meta.http.headers || {};
         meta.view = meta.view || {};
+        appConfig = (configAddon.getAppConfig && configAddon.getAppConfig()) || {};
```
drewfish: This seems strange.  Why not just call `this.config.getAppConfig()`?

imalberto: appConfig is reused a few times in that function (vs invoking getAppConfig())

drewfish: I'm talking about the `configAddon.getAppConfig && configAddon.getAppConfig()` part.  Why not just `appConfig = this.config.getAppConfig()` ?

imalberto: just in case `config` addon is not available. if that is very unlikely, then calling `config.getAppConfig()` directly is cleaner.

drewfish: I prefer to keep it cleaner.  If `store.config` is missing that's quite exceptional, and we should fail early.

imalberto: Sounds good.

imalberto:
@drewfish Closing this PR. See https://github.com/yahoo/mojito/pull/560 instead.

PR #560, opened 2012-09-25
Removed references to app.config in output-adapter addon

imalberto:
FIX removed references to app.config and replaced with ac.config.getAppConfig(); updated unit tests

This PR superceds https://github.com/yahoo/mojito/pull/559.

### DIFF
--- a/lib/app/addons/ac/output-adapter.common.js
+++ b/lib/app/addons/ac/output-adapter.common.js
@@ -51,8 +51,8 @@ YUI.add('mojito-output-adapter-addon', function(Y, NAME) {
             renderer = null,
             contentType,
             contentPath,
-            viewEngineOptions = (this.app && this.app.config &&
-                                 this.app.config.viewEngine) || {},
+            appConfig = this.config.getAppConfig(),
+            viewEngineOptions,
             perf = Y.mojito.perf.timeline('mojito', 'ac.done',
                 'time to execute ac.done process', this.command);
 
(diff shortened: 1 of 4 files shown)


caridy:
@imalberto this is problematic. the main objective was to minimize the amount of addons that were needed to create when creating AC object, if output-adapter requires config, then we have by default 2 addons instead of 1.

Instead, if this addon really need any app config definition, then you can access it directly, without requiring `mojito-config-addon` in the same way that we do there.

So far, I see this:
- instance.appConfig.viewEngine
- this.app.config.cacheViewTemplates
- this.app.config.pathToRoot   // for HTML5build

and in the config addon we are accessing the config via `this._store.getAppConfig(this._ctx)` when needed.

I wonder if `viewEngine` and `cacheViewTemplates` and `pathToRoot` really need to be computed everytime. Maybe @drewfish has more information, but in my mind, only `environment` dimension should affect it, hence we can compute it once, and use them in output out of the box. In my mind, the goal is to avoid calling `this._store.getAppConfig(this._ctx)` everytime, and this is a good opportunity.

imalberto:
@caridy 
- The original impl was using `app.config` object (based on the fact that current AC contructor does not change). so agreed here.
- In comment https://github.com/yahoo/mojito/pull/540#issuecomment-8789854, can you elaborate what you were referring to ? Perhaps some clarification as to what is the preferred way to access `appConfig` in general.
- For recomputing `viewEngine etc etc` config values, +1 if this can be done at startup and simply fetched at runtime

caridy:
+1

we can clean up the ac.app later on.

drewfish:
I think we can't assume that viewEngine will only vary based on environment. Oh... hmm... `ac.app.config` doesn't prevent that, since it's contextualized.

+1

PR #565, opened 2012-09-25
Added support for /robots.txt and fixed JSLint errors

jlecomte:
Eventually, this should be done using new, special routes...

### DIFF
--- a/lib/app/middleware/mojito-handler-static.js
+++ b/lib/app/middleware/mojito-handler-static.js
@@ -19,7 +19,7 @@
  */
 
 
-/*jslint anon:true, sloppy:true, nomen:true*/
+/*jslint node:true, anon:true, sloppy:true, nomen:true, todo:true */
 
 
 /*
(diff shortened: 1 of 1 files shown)


caridy:
Should mojito be smart enough to generate robots and manifest files?

I do think that we should be able to redefine/force specific files though, so I will give it a +1

drewfish:
You could instead have a route for `/robots.txt` that invokes a mojit, but that's probably overkill for most cases.

I'd prefer a solution that allows that to happen, but we can go with this for now, and tweak it later.

caridy:
Ok, I will merge this, and we can plan for having some sort of npm bundle to add support for robots and manifest that can be plugged into the apps by just adding them as dependencies.

caridy:
Ops, I just noticed the error msg in travis about the `todo` flag in jslint, who fails on node 0.8

http://travis-ci.org/#!/yahoo/mojito/jobs/2559216

@mridgway @mojit0 do you know what's going on with this?

mojit0:
@caridy afaik we're using the version in npm and hence the todo flag isn't supported until that package is updated from a jslint version which supports it.

jlecomte:
We should also add support for crossdomain.xml... Caridy, can you add that one line, or do you need a new pull request?

caridy:
@jlecomte u can commit that line under the same PR, and it will be aggregated automatically once u push to your branch.

caridy:
Ok, this is ready to be merged. I will hold it until I finish the update of node-jslint pkg to support `todo` flag, so we can validate the tests and move on.

rwaldura:
@caridy are we still on hold for this?
Can @lzhan help?

lzhan:
@jlecomte is it ok to remove the "todo:true"? I don't see js error from "mojito jslint" command after it is removed.

PR #568, opened 2012-09-25
Sweetandsour

bthaeler:
This code provides two things.
1. A framework for logging debugging and performance metrics data on a per request basis.
2. Some instrumentation of the core mojito code, using this api.

To use this results of this system you will also need the npm package, mojito-debug-view. This will be released separately.


rwaldura:
status on this? update or close?

caridy:
@rwaldura we discussed the details, and @bthaeler will be working with @drewfish and myself to get the whole infrastructure as a NPM pkg rather than part of mojito itself, and using hooks within mojito core to be able to control the flow. I'm closing this one.

bthaeler:
I think Ric wants to meet and discuss this some more. From talking to him, it sounds like we haven’t totally decided which way to go yet.

-Bret

PR #569, opened 2012-09-26
no longer use resource store adapter

drewfish:
- use duck typing to define the API for {client,server} stores
- moved the useful bits of the adapter into store.client (where it matters most)
- related cleanups


caridy:
+1

PR #574, opened 2012-09-27
use POSL and lang instead of context for the cache key in store.expandInstanceForEnv()

drewfish:


caridy:
+1

caridy:
I still think we should attack the action/id as a potential perf issue, because that will change a lot, specially the ID if it is generated automatically like in Julien's demo.

drewfish:
The action & ID are already part of the cache key, as part of the fakeInstance.

PR #576, opened 2012-09-27
Perf refactor

caridy:
- reinstate perf and profiler
- introduce data process scope for low-level implemenatations (which was needed for the perf stuff)

### DIFF
--- a/lib/app/autoload/mojito.common.js
+++ b/lib/app/autoload/mojito.common.js
@@ -21,6 +21,16 @@ YUI.add('mojito', function(Y, NAME) {
     Y.namespace('mojito.addons.ac');
     Y.namespace('mojito.addons.viewEngines');
 
-}, '0.1.0', {requires: [
-    'mojito-perf'
-]});
+    // this is a facade for the real implementation from mojito-perf module
+    // that will have to be plugged manually to get the metrics in the
+    // console or a log file.
+    Y.mojito.perf = {
+        timeline: function () {
+            return {
+                done: function () {}
+            };
+        },
+        mark: function () {}
+    };
+
+}, '0.1.0', {requires: []});
--- a/lib/app/autoload/mojito.common.js
+++ b/lib/app/autoload/mojito.common.js
@@ -33,4 +33,58 @@ YUI.add('mojito', function(Y, NAME) {
         mark: function () {}
     };
 
+    // setting the stage for all data-scopes implementation
+    YUI.namespace('Env.mojito');
+
+    // defining the process data scope bound to YUI.Env, which is
+    // persistent between requests on the same process on the
+    // server side and persistent per page on the client side.
+    // TODO: on the client side we might want to use sessionStorage
+    // TODO: abstract everything related with data into its own component
+    // TODO: this structure leaks, should be well documented
+    YUI.Env.mojito.DataProcess = YUI.Env.mojito.DataProcess || (function () {
+
+        var data = {};
+
+        function key(obj) {
+            return Y.Lang.isObject(obj) ? JSON.stringify(obj) : obj;
+        }
+
+        return {
+
+            /**
+            * Adds a new entry to the cache.
+            *
+            * @method add
+            * @param request  {Object} Request value.
+            * @param response {Object} Response value.
+            * @return {Object} Cached object, or null.
+            */
+            add: function (request, response) {
+                return (data[key(request)] = response);
+            },
+
+            /**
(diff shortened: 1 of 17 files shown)


lib/app/autoload/perf.server.js
```diff
@@ -9,25 +9,26 @@
...
      */
 
-    if (!YUI._mojito) {
-        YUI._mojito = {};
-    }
+    var libfs = require('fs'),
 
+        // using the mojito's process data to retrieve and
+        // store the process process buffer and the config
+        data   = YUI.namespace('Env.mojito').DataProcess,
```
drewfish: A matter of style, but if this requires `mojito` module then we should just be able to access this as `YUI.Env.mojito.DataProcess` no?

lib/app/autoload/mojito.common.js
```diff
@@ -21,6 +21,70 @@ YUI.add('mojito', function(Y, NAME) {
...
+            /**
+            * Adds a new entry to the cache.
+            *
+            * @method add
+            * @param request  {Object} Request value.
+            * @param response {Object} Response value.
+            * @return {Object} Cached object, or null.
+            */
+            add: function (request, response) {
+                return (data[key(request)] = response);
```
drewfish, line 64: The names `request` and `response` are confusing here, since they look like the "req"/"res" pair from connect/express.

I would suggest something like `name` and `data` or perhaps `entryName` and `entryData` (or something like that).

lib/index.js
```diff
@@ -142,19 +141,19 @@ MojitoServer.prototype = {
 
         Y.applyConfig({
             modules: {
```
drewfish, line 143: There are about 4 other places where we hand-make a "modules" config.  They need to be updated too.

lib/index.js
```diff
@@ -168,6 +167,9 @@ MojitoServer.prototype = {
         store.preload();
         appConfig = store.getAppConfig(null);
 
+        // storing appConfig in the process
+        YUI.Env.mojito.DataProcess.add('app-config', appConfig);
```
drewfish: Just a note that this is the "static" app config:  the app config that is contextualized by the startup context (if any).  (It might be better to call this 'static-app-config' just to avoid confusion later.)

drewfish:
How does the code in mojito-perf (real implementation) get loaded in?  If it's via YUI requires:[], that means that if _one_ module somewhere has a requirement on perf then _every_ module starts logging perf.  That seems a little odd.

Another approach is to _automatically_ load `mojito-perf` if `perf` is set in application.json.  This could perhaps be accomplished by adding `mojito-perf` to `CORE_MOJITO_MODULES` in index.js.

lib/app/autoload/perf.server.js
caridy: Agreed, I added the namespace to pass the tests, but after I fixed the mojito mock to provide that namespace, I'm off the hook to use the full namespace here.

lib/index.js
caridy: I was about to call it master-app-config but static-\* sounds better.

caridy:
I'm glad you raise that question, because it was confusing. Right now, for whatever reason unknown to me, all available "shared" modules are attached to the `Y` in `index.js`, this is happening as part of this method:

```
function configureYUI(Y, store, load) {
    var shared,
        module;
    shared = store.yui.getConfigShared('server', {}, false);
    Y.applyConfig(shared);
    // also pre-load shared modules
    for (module in shared.modules) {
        if (shared.modules.hasOwnProperty(module)) {
            load.push(module);
        }
    }
}
```

where "load" is actually a reference to `CORE_MOJITO_MODULES`, and as part of that process, perf and logger are attached, just like all the other common and server components from autoload.

The "feature" is what made me remove the `use` statement for `logger` and `perf` in case they were required, and requiring them everytime, but adding a conditional statement within `perf` to do nothing if `DataProcess->perf-config` was not set upfront.

lib/app/autoload/mojito.common.js
caridy, line 64: Well, the idea is to provide an abstraction layer, not only for process, but for any kind of data storage based on our definition of data scopes. For that, and special the client side, YUI has the `cache` module and `offline-cache` as well sharing the same api that supports expiration, maximum, etc. And this is the API that they use today where request and response are generic data structure, and they can be serialized when needed.

http://yuilibrary.com/yui/docs/cache/

imalberto:
+1

PR #578, opened 2012-09-28
adding some guarding in case perf is not enabled, in which case we do no...

caridy:
...t need to instrument the request.


caridy:
merging the guard since this is blocking the search-news-app tests.

PR #582, opened 2012-10-01
modify dimensions.json to meet ycb standard

lzhan:


caridy:
@lzhan those environments are not aligned with those defined in the manhattan workflow. Is that something we want to provide out of the box as the default config or not?

lzhan:
@caridy Thanks for point it out. I am trying to fix bug 5532859 which requires our default dimensions.json to meet an ycb doc. I am not aware of the "the manhattan workflow". Would you please give me a pointer?

redonkulus:
@caridy seems like the manhattan workflow would be a specific yahoo use case right? Is that something we need to enforce for the non-y community?

lzhan:
I am closing this pr and will send another pr with 1) keep current dimension.json for open source user. 2) An new dimension.json for yahoo internal user-- when internal user create mojito app, the dimension.json for internal user will be used.

PR #586, opened 2012-10-02
AC addon namespace should match the ac filename

caridy:
temporary solution


lib/app/addons/rs/dispatch-helper.server.js
```diff
@@ -107,7 +109,7 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
                 }
                 if ('addon' === res.type && 'ac' === res.subtype) {
                     modules[res.yui.name] = this.get('host').yui._makeYUIModuleConfig(env, res);
-                    acAddonNames[res.yui.name] = res.name;
+                    acAddonNames[res.yui.name] = libpath.basename(res.name);
```
drewfish, line 115: It might be good if there's a BIG LOUD COMMENT saying that this is a temporary hack/workaround.

PR #588, opened 2012-10-02
more careful about use of Y.Loader

drewfish:
Mainly submitting PR to get feedback.  Cleanup still needed (though should be easy).

### DIFF
--- a/lib/app/addons/rs/yui.server.js
+++ b/lib/app/addons/rs/yui.server.js
@@ -28,11 +28,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
             tests: true,
             yui_modules: true
         },
-
-        intlPath,
-        // experimental approaches
-        X_NEW   = false,
-        X_NEW2  = true;
+        intlPath;
 
     function RSAddonYUI() {
         RSAddonYUI.superclass.constructor.apply(this, arguments);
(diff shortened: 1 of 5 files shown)


lib/app/addons/rs/yui.server.js
```diff
@@ -27,7 +27,12 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
             autoload: true,
             tests: true,
             yui_modules: true
-        };
+        },
+
+        intlPath;
+        // experimental approaches
+        X_NEW   = false,
+        X_NEW2  = true;
```
drewfish: X_NEW2 has the proposed implementation.  (Ignore X_NEW.)

lib/app/addons/rs/yui.server.js
```diff
@@ -437,9 +525,10 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
          * @param {string} mojit name of the mojit
          * @param {object} modules YUI module metadata
          * @param {object} required lookup hash of YUI module names that are required
+         * @param {boolean} forceYLoader whether to force the use of Y.Loader
```
drewfish, line 451: We needed to add this because the dispatch-helper does need to do a proper YUI dependency calculation.  Considering the significant differences in implementation, this should probably be split into two methods.

caridy:
Ok, let's just clean this up for the PR @drewfish .

PR #593, opened 2012-10-04
Merge assets from config.assets to meta

diervo:
In mojito 0.4.x due to the automatic inclusion of the mojito-assets-addon, when executing the loop on line 124 it calls the metaMergeInto of the assets addon which makes the config to merge.

Since we dont have that addon as default in the new mojito version, we will need to do it manually.

If you add the assets addon as a dependencie, due to the implementation of metaMerge will not duplicate assets. So is just a small overhead on computation in that case. 

This is a must merge to Shaker.


caridy:
+1

drewfish:
lint passes, unit tests pass, +1

PR #595, opened 2012-10-04
much better about YUI module dependencies when calculating for the client

drewfish:


drewfish:
jslint and unit tests pass

caridy:
+1

I like `original` better than `old` though.

PR #597, opened 2012-10-05
removed lang dependencies from binders.

diervo:
The langs are already dependencies of the controller so we dont need it
here. It's causing the langs to get deployed always to the client.

### DIFF
--- a/lib/app/addons/ac/composite.common.js
+++ b/lib/app/addons/ac/composite.common.js
@@ -312,16 +312,25 @@ callback({
                     child = children[childName];
                     // Create a buffer for the child
                     buffer[childName] = {name: childName, data: '', meta: {}};
-
                     // Make a new "command" that works in the context of this
                     // composite
                     newCommand = {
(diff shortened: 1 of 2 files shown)


caridy:
@drewfish can you comment on this?

drewfish:
I agree with the idea, but the proposed code isn't quite right, since without calling `_setYUIDependencies()` calling `_precomputeYUIDependencies()` is wasted effort.

Whether we need to call `_precomputeYUIDependencies()` or not depends on whether we intend to keep support for `dependencyCalculations:precomputed`.

diervo:
But still, whatever calculation we do, the langs are not dependencies of the binder but the controller right? We should only need to ship those when the controller gets deployed.

drewfish:
Yes, I agree.  However, the code in this pull request isn't right.

diervo:
Ok, anyway with the Caridy changes that may be refactored somewhere else. So just to raise the error!

PR #600, opened 2012-10-08
minor fixes to cli.js

isao:
- fix exception in build cmd was invoked without parameters
- allow commands to show usage and exit
- rm unused function params


mojit0:
+1

PR #601, opened 2012-10-09
Mojito mojito-lite into develop-perf branch

caridy:
@drewfish, lets merge this!


drewfish:
+1.  If any issues come up we'll fix forward.

PR #603, opened 2012-10-09
Fix for expandinstace so we dont copy the params object

diervo:
Performance boost!


drewfish:
This is missing `child.config` which is very important I suspect.

diervo:
This is a Ric and I fix for avoid the overhead of copying the params. It seems that is working fine..
It was never there in the first place...

diervo:
This is input to the composite addon, and in the docs seems that we dont need the config at all. It only mention params: 

```
cfg = {
    children: {
        slot-1: {
            type: "default",
            action: "index"
        },
        slot-2: {
            type: "default",
            action: "index",
            params: {
                route: {},
                url: {},
                body: {},
                file: {}
            }
        }
    },
    assets: {}
}
```

drewfish:
I think `child.config` is needed by other users.

diervo:
I just added the config to the PR :)

PR #606, opened 2012-10-09
new: archetype for hybrid app/mojit combo

isao:
- add lib/app/archetypes/app/hybrid/\* which contains both an app and mojit 
  template/scaffolding; primarily for hybrid html5app configs
- tweak create.js so usage text reflects new `mojito create app hybrid` subtype
- remove lingering remnant of `create project`


isao:
This change for `mojito create app hybrid` creates an app that contains a mojit named "top_frame". This is different from `mojito create app [simple,default,full]` which only create the app, and the mojit is created in a second step. However, this was the requirement I was given.

isao:
n.b. if archetypes include binder tests, `mojito test app .` fails with dom on nodejs error; fixed for pre-existing archetypes in pull #607

Issue #525, opened 2012-09-18
Running 'mojito test app <app_name>' for the examples in 'developer-guide' returns an error.


zhouyaoji:
After you create a new app with Mojto 0.4.4, you can run the following to test the app **without**
any problems:

$ mojito test app <new_app>

If you run the same command, however, for the examples in '/examples/developer-guide/':
developer-guide$ mojito test app hello

...you get the following error (truncated):

/opt/local/lib/node_modules/mojito/node_modules/yui/yui-nodejs/yui-nodejs.js:1147
            throw (e || new Error(msg));
                        ^
TypeError: Cannot read property 'documentElement' of null

zhouyaoji:
Related GH issue that was already closed: https://github.com/yahoo/mojito/issues/387

dmitris:
I see the same error when running
mojito test app .
in any subdirectory of examples/developer-guide/ - for example examples/developer-guide/adding_view_engines or examples/developer-guide/binding_events or examples/developer-guide/configure_routing
The full error is:

dmitris$ mojito test app .
warn: (yui): NOT loaded: mojito
warn: (yui): NOT loaded: mojito
warn: (yui): NOT loaded: mojito
warn: (addon-rs-yui): your mojit "RoutingMojit" has a server affinity and these client-related deps: dom-core, dom-base, node-core, node-base, node-event-delegate, node-pluginhost, dom-style, dom-style-ie, dom-screen, node-screen, node-style
warn: (addon-rs-yui): Mojito may be unable to start, unless you have provided server-side DOM/host-object suppport
warn: (mojito-perf): microtime not found. Recorded times will not have microsecond accuracy
warn: (mojito-perf): microtime not found. Recorded times will not have microsecond accuracy

/Users/dmitris/.nvm/v0.6.21/lib/node_modules/mojito/node_modules/yui/yui-nodejs/yui-nodejs.js:1147
            throw (e || new Error(msg));
                        ^
TypeError: Cannot read property 'documentElement' of null
    at Object.fn (/Users/dmitris/.nvm/v0.6.21/lib/node_modules/mojito/node_modules/yui/dom-base/dom-base-min.js:7:50)
    at [object Object]._attach (/Users/dmitris/.nvm/v0.6.21/lib/node_modules/mojito/node_modules/yui/yui-nodejs/yui-nodejs.js:701:33)
[...]

PR #607, opened 2012-10-09
fix issue #525 `mojito test app|mojit` failed for archetype-code

isao:
This change removes binder tests from the archetypes that `mojito create` uses.
Binder unit tests are unfortunately no longer supported by `mojito test`. Users
may look ar yahoo-arrow available on npm, for unit and functional testing
mojito code, including binders.


mojit0:
+1

PR #608, opened 2012-10-10
upgrading to YUI 3.7.2 to facilitate the use of new loader

caridy:
- improving loader integration 
- adding some guarding on combo handler.
- formalizing extensions and content-type. 
- supporting ylinux/yiv
- adding support for css on the combo handler


lib/app/middleware/mojito-combo-handler.js
```diff
@@ -319,31 +334,51 @@ function staticProvider(store, globalLogger, Y) {
...
         // so errors can be found early on.
         for (i = 0; i < files.length; i += 1) {
             yui = libpath.join(__dirname,
                     '../../../node_modules/yui', files[i]);
 
-            module = libpath.basename(files[i]).split('.')[0];
-            module = module.replace(/\-(min|debug)$/, '');
+            // something like foo/bar-min.js should become just "bar"
+            module = libpath.basename(files[i]).
+                replace(/\.(js|css)$/, '').
```
drewfish: Using `libpath.basename(files[i], libpath.extname(files[i]))` instead of the `replace(/\.(js|css)$/, '')` is a little more correct/robust, but might not be worth the effort.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -288,7 +299,10 @@ function staticProvider(store, globalLogger, Y) {
 
         logger.log('serving static path: ' + url.pathname, 'debug', 'static-handler');
 
-        files = url.query.split('&');
+        // YIV might be messing around with the querystring params
+        // so we need to revert back to /
+        // TODO: this might not work in Windows
+        files = url.query.replace(/\=/g, '').replace(/\%2F/g, '/').split('&');
```
drewfish: What does `\=` match?  I didn't think that `=` was a special token in regex, and thus doesn't need escaping.  Ditto with `\%2F`, I didn't think that `%` was special.

drewfish:
Besides two small comments, +1.

lib/app/middleware/mojito-combo-handler.js
caridy: I added the extension explicitly to trigger the error if they try to access something else, but later on I added a verification of the extension, so I can use this structure now. Should be ready in the next commit.

lib/app/middleware/mojito-combo-handler.js
caridy: lol, I'm not very good at regex in general, so I escape anything that it suspicious. Feel free to clean that up.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -319,31 +334,54 @@ function staticProvider(store, globalLogger, Y) {
...
+            // something like foo/bar-min.js should become just "bar"
+            module = libpath.basename(files[i], ext).
+                replace(/\-(min|debug)$/, '');
 
             if (module === 'loader-app-base') {
                 result[i] = {
                     fullpath: module,
-                    content: "YUI.applyConfig({modules:" + appMetaData + "});"
+                    content: 'YUI.add("loader",function(Y){' +
+                                'YUI.Env[Y.version].modules=YUI.Env[Y.version].modules||' +  appMetaData + ';' +
```
drewfish: This code seems a little strange -- it seems like mojito has too much knowledge of how YUI works.  Is there a YUI API call we can make instead (loader.addGroup() or something like that) ?

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -319,31 +334,54 @@ function staticProvider(store, globalLogger, Y) {
...
+                             '}, "", {requires:["loader-base"]});'
                 };
             } else if (module === 'loader-app-full') {
                 result[i] = {
                     fullpath: module,
-                    content: "YUI.applyConfig({modules:" + appResolvedMetaData + "});"
+                    content: 'YUI.add("loader",function(Y){' +
+                                'YUI.Env[Y.version].modules=YUI.Env[Y.version].modules||' + appResolvedMetaData + ';' +
```
drewfish: Ditto previous comment.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -319,31 +334,54 @@ function staticProvider(store, globalLogger, Y) {
...
+                             '}, "", {requires:["loader-base"]});'
                 };
             } else if (urls[module]) {
                 result[i] = {
                     fullpath: urls[module],
                     content: ''
                 };
-            } else if ((module.indexOf('..') === -1) && existsSync(yui)) {
+            } else if ((files[i].indexOf('..') === -1) && existsSync(yui)) {
```
drewfish: Is `existsSync()` being called during runtime?  That might have bad performance implications.

lib/app/middleware/mojito-combo-handler.js
caridy: In theory, it should use `YUI.applyConfig()`, but there is a remaning bug in yui that prevents us from using the config API. Instead, we need to plug them directly into the main object. Dav will have this fix for the next release.

lib/app/middleware/mojito-combo-handler.js
drewfish: http://nodemanual.org/latest/js_doc/RegExp.html indicates that neither `=` nor `%` are special, so I think the regexes should be `/=/g` and `/%2F/g` instead.

lib/app/middleware/mojito-combo-handler.js
drewfish: Please add that information to a comment.  We really should circle back and do it the right way (once the right way is available).

lib/app/middleware/mojito-combo-handler.js
caridy: In theory, in production, we will use the YCS combo rather than the app combo. But I do agree we should make this more performance, maybe creating a hash in memory for all YUI modules or just not dealing with YUI modules at all. Let me think about this.

lib/app/middleware/mojito-combo-handler.js
caridy: http://yuilibrary.com/projects/yui3/ticket/2532854

drewfish:
+1

PR #612, opened 2012-10-11
Develop perf

bthaeler:
This code provides two things.
1. A framework for logging debugging and performance metrics data on a per request basis.
2. Some instrumentation of the core mojito code, using this api.

To use this results of this system you will also need the npm package, mojito-debug-view. This will be released separately.


caridy:
This PR is very long and confusing, probably because there is a mix of old code with new code, and refactors in the middle, I will need more time to make sense of it.

I do have few questions already:
- How does this work on client side?
- How a `model` for example, can have access to debugger?
- How does this work with the current perf api? They clearly overlap, one of them have to be blended.

bthaeler:
The intent is to eventual make this api work on both the client and server side. We should be able to get the Debug UI to merge both server side and client side results together. This is a feature we plan to work on next quarter.

To use the debugger in a model, one would need to pass down debugger to the getData methods in the model.

The code does work woth the current perf api. If you enable perf marks, prof.PerfMark will be an available debug flag. The results will appear in a water fall display.

caridy:
@bthaeler let's have a quick meeting to talk about this PR. /cc @drewfish

drewfish:
Hi @bthaeler,

Drew

PR #613, opened 2012-10-11
pathToRoot should be dynamic rather than static config

caridy:
adding config addon into url to make sure that pathToRoot can be customized per context.


caridy:
@imalberto can u review this?

imalberto:
NOTE: that now the url addon is reading the appConfig from the contextualized  and not the static config anymore to address html5app that are built with a specific context.
+1

PR #615, opened 2012-10-11
[doc] Added func/unit test instructions.

zhouyaoji:


lzhan:
"./run.js test -u" is changed to "./run.js test -u --path unit", and "./run.js test -f" is changed to "./run.js test -f --path func".

PR #616, opened 2012-10-11
loader metadata per lang

caridy:
- centralized configuration through `application.json->yui->config` for loader on server, loader on client, combo handler and application in general.
- `loader-app-base` and `loader-app-full` should be lang aware for performance reasons 
- define a new way to configure the metadata url
- define a way to enable/disable metadata as part of the seed (aka yui->config-url)


caridy:
More work to be done on this.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -46,6 +47,10 @@ var libfs = require('fs'),
     MODULE_META_PRIVATE_ENTRIES  = ['after', 'expanded', 'supersedes', 'ext', '_parsed', '_inspected',
                                     'skinCache', 'langCache'],
 
+    REGEX_LANG_TOKEN = /\"\{langToken\}\"/g,
+    REGEX_LANG_PATH  = /\{langPath\}/g,
+    REGEX_LOCALE     = /\_([a-z]{2}-[A-Z]{2})$/,
```
mojit0: Do we need to watch for simple 'en' or 'de' here?

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -226,59 +281,103 @@ function processMeta(resolvedMods, modules, expanded_modules, conditions) {
...
+
+    Y = YUI({
+        fetchCSS: true,
+        combine: true,
+        base: "/combo?",
+        comboBase: "/combo?",
+        root: ""
+    }, ((appConfig.yui && appConfig.yui.config && appConfig.yui.config.config) || {}));
+
+    modules_config = Y.merge((mojits.modules || {}), (shared.modules || {}));
```
mojit0, line 320: this is fine for now, but i'd like to see us adopt a pattern that doesn't create objects that might not be necessary. so anywhere we could check for null/undefined instead of creating objects which affect memory consumption we should.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -226,59 +281,103 @@ function processMeta(resolvedMods, modules, expanded_modules, conditions) {
...
         maxAge = cache;
     }
     maxAge = maxAge || 0;
 
-    processMeta(resolved.jsMods,  modules, expanded_modules, conditions);
-    processMeta(resolved.cssMods, modules, expanded_modules, conditions);
+    processMeta(resolved.jsMods,  modules, expanded_modules, langs, conditions);
+    processMeta(resolved.cssMods, modules, expanded_modules, langs, conditions);
+
+    for (i = 0; i < langs.length; i += 1) {
```
mojit0, line 342: slight performance gain if we cache length before the loop.

lib/app/middleware/mojito-combo-handler.js
```diff
@@ -226,59 +281,103 @@ function processMeta(resolvedMods, modules, expanded_modules, conditions) {
...
+        lang = langs[i];
+
+        appMetaData[lang] = {};
+        appResolvedMetaData[lang] = {};
+
+        for (name in expanded_modules) {
+            if (expanded_modules.hasOwnProperty(name)) {
```
mojit0, line 349: slight performance gain to cache expanded_modules[name] as 'slot' or whatever rather than reaccessing it repeatedly through the [] operator.

lib/app/middleware/mojito-combo-handler.js
caridy: good point, I will generalize that.

mojit0:
Other than optional nits, +1

lib/app/middleware/mojito-combo-handler.js
caridy, line 320: agreed.

PR #622, opened 2012-10-12
new `mojito create app` archetype "hybrid" & new custom archetype support

isao:
sorry for the messy PR but it essentially contains three changes:
1. small bug fix for build.js not exiting on error
2. new archetype for `mojito create app hybrid` which incorporates hybrid/crt app settings.
3. reworked create.js
   - new `mojito create custom path/to/custom/archtype`
   - allow optional user-defined key/value pairs to be passed to custom templates

In anticipation of #2 to require changes and updates, I added #3, which may restore `mojito create project xcode` functionality (without mojito needing to update the sources).


isao:
specs changed, will resubmit

PR #626, opened 2012-10-12
support for moving YUI meta computation out of combo middleware

drewfish:


caridy:
Ok, let's fix the 404 and let's merge this.

PR #632, opened 2012-10-15
update express and connect to a version supporting node0.8

lzhan:


isao:
+1 (the failed test on 0.6 only is the example app yql call again)

PR #633, opened 2012-10-16
add IDE/Mai Tai support to `mojito build` command

isao:
new build command options:
- --snapshotTag/-t package snapshot tag
- --snapshotName/-n package snapshot name

new files created:
- snapshot.json
- filelist.txt


isao:
specs changed, will resubmit

PR #636, opened 2012-10-16
store sandbox using yui-sandbox

caridy:
- isolation between store instance and the mojito Y instace
- remove experimental dataprocess


mojit0:
+1

caridy:
let's wait until after pr2 patch happens to merge this.

caridy:
Travis is failing due to a timeout rather than a code problem I guess: "Executing your script (./travis/script.sh) took longer than 30 minutes and was terminated."

drewfish:
+1.  I think that store.server.js should probably be using yui-sandbox instead, for when it uses Y.Loader.

caridy:
Yes, part of the next refactor. Trello card #88 :)

PR #623, opened 2012-10-12
Jl

lzhan:
This pull request includes changes made from pr #565 and a jslint error fix.

PR #638, opened 2012-10-16
FIX static-handler middleware for well known files

imalberto:
Appears that static middleware does not quite work for the following: /robots.txt, /crossdomain.xml and /favicon.ico.

This PR addresses this, with unit tests.

### DIFF
--- a/lib/app/middleware/mojito-handler-static.js
+++ b/lib/app/middleware/mojito-handler-static.js
@@ -227,13 +227,13 @@ function staticProvider(store, logger) {
 
         // TODO:  these adjustments should really be done by addons/rs/url
         if (!resource && (path === '/favicon.ico')) {
-            resource = store.getResource('client', {}, 'asset-ico-favicon');
+            resource = store.getResources('client', {}, 'asset-ico-favicon');
         }
         if (!resource && (path === '/robots.txt')) {
-            resource = store.getResource('client', {}, 'asset-txt-robots');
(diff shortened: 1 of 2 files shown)


lib/app/middleware/mojito-handler-static.js
```diff
@@ -227,13 +227,13 @@ function staticProvider(store, logger) {
...
-            resource = store.getResource('client', {}, 'asset-ico-favicon');
+            resource = store.getResources('client', {}, 'asset-ico-favicon');
         }
         if (!resource && (path === '/robots.txt')) {
-            resource = store.getResource('client', {}, 'asset-txt-robots');
+            resource = store.getResources('client', {}, 'asset-txt-robots');
         }
         if (!resource && (path === '/crossdomain.xml')) {
-            resource = store.getResource('client', {}, 'asset-xml-crossdomain');
+            resource = store.getResources('client', {}, 'asset-xml-crossdomain');
```
drewfish: Good catch.  However, `store.getResources()` returns a list of resources, so you might need to `[0]` on it.  Also, I think the `'asset-xml-crossdomain'` argument might instead need to be `{id: 'asset-xml-crossdomain'}`.  (Said another way: I think I rather bungled the previous version of the code :disappointed:  .)

rwaldura:
see also #565 and follow-up #623 which dealt with these special files

caridy:
Aside from those minor details in the tests +1

caridy:
+1

PR #643, opened 2012-10-18
archetype and build type for crt/updater/ide

isao:
new archetype: `mojito create app hybrid myapp`
new build type: `mojito build hybrid path/to/packages/dir`
https://gist.github.com/26623632f9c593be9078
though the build command was split into seven different sub modules, the eloc count is lower at the moment.

new support for user-specified archetypes: `mojito create custom path/to/your/archetype mything`


drewfish:
For finding custom archetypes, I would have preferred if it used the resource store (`store.getResources('server', {}, {type:'archetype', subtype:archetypeType})`), so that archetypes delivered by NPM modules could be easily used.

isao:
@drewfish We discussed this when I tried implementing the hybrid archetype as a package-- RS doesn't support it, and the existing create.js doesn't use mojito or the resource store at all yet.

This feature is pretty minor in terms of code changes, and may be useful for the IDE/CRT integration.

I agree we should support of npm packaging of archetypes in the future, as planned, so that archetypes can be sanely shared, versioned, etc. If that future/planned feature blocks this entire PR, let me know and I'll make a new one without the `mojito create custom` feature.

drewfish:
OK cool.  I don't imagine that anything in this PR will block future use of archetypes delivered via NPM modules.

So, +1.

PR #644, opened 2012-10-18
replace express with connect ">1.8.2"

isao:
Mojito does not use express-specific apis, only the underlying connect ones. This PR replaces references to express with connect.

Connect docs state: 1.x is compatible with node 0.4.x, 2.x (master) is compatible with node 0.6.x

Not sure if just saying >1.8.2 is what we want, but will submit for discussion.

### DIFF
--- a/lib/app/middleware/mojito-parser-body.js
+++ b/lib/app/middleware/mojito-parser-body.js
@@ -8,13 +8,13 @@
 /*jslint anon:true, sloppy:true*/
 
 
-var express = require('express');
+var connect = require('connect');
 
 /**
  * Export a function which can create a body parser.
  * @return {Object} The newly constructed body parser.
  */
 module.exports = function() {
-    return express.bodyParser();
+    return connect.bodyParser();
 };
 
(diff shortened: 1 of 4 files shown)


mojit0:
+1

isao:
ah, need to fix this fix: https://github.com/isao/mojito/blob/develop/lib/mojito.js#L455

connect.close() method doesn't exist (anymore?). seems like it's only called via mojito build & compile cmds.

drewfish:
Looking at the express@2.5.2 code, it looks like it implements a `res.send()` as a wrapper around `res.end()` which is part of `http.ServerResponse`.

isao:
thx @drewfish.

side note, according to this [poll](https://docs.google.com/spreadsheet/viewform?fromEmail=true&formkey=dDZXaTk1S0k1bU8zLUxmY2gtSXd6eXc6MQ) 14 groups use custom middleware (which will probably need some updates) and only one cites express 2.x as a dep.

isao:
thinking we wait till node 0.4 support is deprecated and just go to connect 2.x...

isao:
revist later

PR #645, opened 2012-10-18
fix bz5472979 server resources in build

isao:
server affinity resources do not need to be uri-
addressable. With this change store.getAllUrls()
will no longer include foo.server.js resources,
and hence be omitted from `mojito build`

### DIFF
--- a/lib/app/addons/rs/url.server.js
+++ b/lib/app/addons/rs/url.server.js
@@ -181,8 +181,9 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 rollupParts = [],
                 rollupFsPath;
 
-            // Don't clobber a URL calculated by another RS addon.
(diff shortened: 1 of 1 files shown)


isao:
thanks @drewfish for the fast diagnosis and remedy.

smoke test:

build html5app before this PR

```
% tree ../aaa-build-hybrid/ | grep server
│   │   │   ├── carrier.server.js
│   │   │   ├── cookie.server.js
│   │   │   ├── deploy.server.js
│   │   │   ├── device.server.js
│   │   │   ├── http.server.js
│   │       ├── hb.server.js
│   │       └── mu.server.js
│       ├── perf.server.js
│   ├── controller.server.js
│   │   └── foo.server.js
%
```

after:

```
% tree ../aaa-build-hybrid/ | grep server
%
```

drewfish:
+1

PR #646, opened 2012-10-18
Clean up of expanded instance

caridy:
- remove loader configuration from expanded instance
- clean boot process
- remove deprecated instance.guid
- fix client side to work with the new loader


lib/store.server.js
```diff
@@ -826,8 +826,18 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 ress = this.getResourceVersions({mojit: mojit});
                 for (r = 0; r < ress.length; r += 1) {
                     res = ress[r];
-                    if (res.yui && res.yui.name) {
+                    // exposing YUI modules with common or client affinity.
+                    // TODO: this should be handled differently, asking for
+                    // the specific resources with the correct affinity.
+                    if (res.yui && res.yui.name && res.affinity &&
```
drewfish: Since a single resource can have both.common. and .client. versions, I think this needs to be done differently. I think we should first walk common version, then walk client versions, so that the metadata about client version clobbers (has higher priority than) metadata about common versions.

Actually, this shouldn't be "common and client" but instead "common and env (current runtime)".

drewfish:
+1

PR #652, opened 2012-10-19
mv run.js/arrow deps to devDependencies for now

isao:
async, commander & node-static are only used by run.js (node-static was listed twice). 

these are currently only used for mojito-dev, so the move makes sense and saves some bandwidth.


caridy:
+1

PR #653, opened 2012-10-22
FIX: mojit achetypes; added ejs to devDependencies;

imalberto:


caridy:
+1

mridgway:
What is ejs being used for?

imalberto:
"ejs" is being used by some of the developerguide examples showcasing templates.

isao:
@imalberto can we remove ejs and the associated test that requires it? ejs is not whitelisted for internal use per @caridy

imalberto:
OK. Just sync'ed with @caridy. Removed in commit https://github.com/imalberto/mojito/commit/3c2c36d03550a0ad701349894e0e78e4db59ec23 as part of PR #682

PR #658, opened 2012-10-23
fix bz 5886351 cli bin/mojito uses wrong copy 

isao:
the cli entry point bin/mojito should always load and run the code it belongs with, and not from further up the require path chain.


drewfish:
+1

PR #661, opened 2012-10-24
Configuration for user-provided middlewares

rodnic:
Now the developer can load custom middlewares from a folder. If options.midDir is empty, the folder ./middleware is assumed to be the default location to store custom middlewares. This new configuration is useful for middlewares that need to be loaded before the server starts.

Example of configuration:
# server.js

options.midDir = 'middleware'; 
// options.midDir = './custom-scripts/middleware'; works too

// Some middlewares
options.appConfig = { 
  middleware: [
    'middleware-passport-init',
    'middleware-passport-session',
    'middleware-facebook-auth'
  ]
};

var app = Mojito.createServer(options);

### DIFF
--- a/lib/mojito.js
+++ b/lib/mojito.js
@@ -199,6 +199,11 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
     if (!options.context) {
         options.context = {};
     }
+    if (!options.midDir) {
+      options.midDir = 'middleware';
+    }
+
+
 
     // all logging that comes from YUI comes from here
     // We need to do this early, since creating a Y instance appears to copy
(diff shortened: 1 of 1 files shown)


ooskapenaar:
-1

The change you are proposing is not backwards compatible.

It seems that the following behaviour needs to be documented better:

if you use   'middleware/passport-init'   instead of   'middleware-passport-init',   you can specify any subfolder transparently and mojito will handle it correctly.  

Also, if you name one of your middleware files  beginning with mojito, e.g.  'middleware/mojito-my-Y-enabled-script'  you will disable the default mojito middleware and your script will be passed a config object.

2c Ron

rodnic:
it's just another way to set up where to load middlewares. If options.midDir was not specified, mojito will assume the folder  'middleware' as default. But you can still set up application.json.

rodnic:
@ooskapenaar hmm... It would be better to let application.json handle all of this.

PR #666, opened 2012-10-25
hybrid archetype fixes

isao:
- add a custom build dimension, and use it so hybridapp build can opt-into crt resources or ss web app
- add missing mojito package.json "type"
- rm old assertion from mojito test, old function
- fix bz 5904449 garbled console output on node 0.8+

### DIFF
--- a/lib/app/archetypes/app/hybrid/application.json.hb
+++ b/lib/app/archetypes/app/hybrid/application.json.hb
@@ -2,7 +2,7 @@
     {
         "settings": [ "master" ],
 
-        "appPort": {{port}},
+        "appPort": "{{port}}",
 
         "specs": {
             "frame": {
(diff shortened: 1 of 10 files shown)


lib/app/archetypes/app/hybrid/application.json.hb
```diff
@@ -2,7 +2,7 @@
     {
         "settings": [ "master" ],
 
-        "appPort": {{port}},
+        "appPort": "{{port}}",
```
drewfish, line 5: I'm surprised that this is needed, since the port can only be an integer, no?  (The change can't hurt.)

lib/app/archetypes/app/hybrid/dimensions.json
```diff
@@ -0,0 +1,189 @@
+[
```
drewfish, line 1: Why do hybrid apps need a separate dimensions.json file?

drewfish:
Some small small questions, even without answers +1.

lib/app/archetypes/app/hybrid/application.json.hb
isao, line 5: yes, it was int before, but the braces messed up the syntax highlighting/linting, and it didn't affect functionality.

lib/app/archetypes/app/hybrid/dimensions.json
isao, line 1: To provide the build:\* context config. I tried adding to the environment:\* context but I guess that's not permitted.

Some app configs are intended for use _only_ with a hybrid build inside CRT, and potentially per device. The build:debug and build:release were chosen to mirror Anne-Lise's scripts for copying dev vs minified resources.

isao, line 1: @zhouyaoji thought these dimensions belong in mojito, but I wanted the change set to be minimal wrt non-hybridapp users and the mojito dimension issue seems to be in flux.

PR #670, opened 2012-10-26
pr 580 , enhancements and tests

lzhan:
This pull request is based on ric's pr #580. Enhancement and code optimization have been made. Fixed one bug found. Added functional tests. Moved the unit tests to arrow.


lib/app/addons/rs/config.server.js
```diff
@@ -61,33 +62,77 @@ YUI.add('addon-rs-config', function(Y, NAME) {
...
             return Y.mojito.util.copy(json);
         },
 
+        /*
+            This function will try and read the requested file returning an Object.
+
+            It will also trim ".ext" from the filePath
+         */
+
+        /**
```
drewfish, line 82: Why are there two comments for the same function?  It should be one comment I think.

lib/app/addons/rs/config.server.js
```diff
@@ -61,33 +62,77 @@ YUI.add('addon-rs-config', function(Y, NAME) {
...
+         * Reads and parses a JSON or YAML structured file.
+         * @method readConfigSync
+         * @param {string} fullPath path to JSON or YAML file
+         * @param {int} type is the index of the file-type to read 0: .yml, 1: .yaml, 2: .json
```
drewfish: The `type` parameter is strange.  Why do we need it -- why can't we just iterate over `extensions` in the function itself, instead of recursing?

lib/app/addons/rs/config.server.js
drewfish, line 82: Also, the fact that this trims the extension from the file is sort-of an implementation details.  What this method is really doing is trying different "flavors" of the file, first file.yml, then file.yaml, then file.json.  That is what the caller cares about, I think.

lib/app/commands/start.js
```diff
@@ -55,20 +55,37 @@ exports.run = function(params, opts, callback) {
...
+        appConfig = store.config.readConfigYCB(path.join(root, 'application.json'));
     } catch (err) {
         appConfig = {};
     }
 
-    try { // Read the package.json config
-        pack = Y.JSON.parse(fs.readFileSync(path.join(root, 'package.json'),
-            'utf8'));
+    try { // Read the package config
+        pack = store.config.readConfigSimple(path.join(root, 'package.json'));
```
drewfish, line 88: This suggests that we support `package.yaml`, is that true?  Does npm support that?

lib/store.server.js
```diff
@@ -916,7 +915,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                     parents: [],
                     dir: dir
                 };
-                info.pkg = this.config.readConfigJSON(this._libs.path.join(dir, 'package.json'));
+                info.pkg = this.config.readConfigSimple(this._libs.path.join(dir, 'package.json'));
```
drewfish, line 918: This suggests that we support `package.yaml`, is that true?  Does npm support that?

package.json
```diff
@@ -18,6 +18,7 @@
         "express": "2.5.10",
         "glob": "~3.1.11",
         "jslint": "~0.1.9",
+		"js-yaml": "1.0.2",
```
drewfish: The indenting is messed up.  Try using spaces (instead of tabs).

lib/app/addons/rs/config.server.js
drewfish: The `@param {int} type` documentation is left over from an older version of this method.

drewfish:
+1

PR #675, opened 2012-10-29
More functests fixes

imalberto:
Only remaining failing tests are related to html5app and intl related.


caridy:
+1, I might be wrong, but @drewfish was fixing a similar issue, there might be conflicts.

PR #676, opened 2012-10-29
restoring controller as instance rather than singleton

caridy:
- instance.getController is not instance.createController
- this change might help to avoid memory leaks

### DIFF
--- a/lib/app/autoload/dispatch.common.js
+++ b/lib/app/autoload/dispatch.common.js
@@ -71,17 +71,21 @@ YUI.add('mojito-dispatcher', function(Y, NAME) {
                     Y.mojito.perf.mark('mojito', 'core_dispatch_start',
                         'dispatching an instance', command);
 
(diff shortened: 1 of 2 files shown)


drewfish:
I don't think we need to change the name of the function.  In my mind `getController()` works well for both cases (create or return singleton) -- it's just a factory.

Later we can add a configuration option so that config in application.json (or perhaps even in the mojit's definition.json) can decide whether a new controller instance is created or not.

drewfish:
Even without the changes mentioned above, +1.

caridy:
Yeah, I decided to change it because:
- getController or createController are only relevant for testing because the `instance` object is created per request per mojit instance using the blending mechanism. I believe it should be removed.
- it is not really a factory, because you're going to get a new method on every dispatch.
- "createController" is more descriptive, and follows our current guideline, like store.createStore() etc.

PR #677, opened 2012-10-29
optionally log some memory usage info

drewfish:
This logs the sizes of a few common objects, plus the
`process.memoryUsage()` sizes.  Enable by putting `"debugMemory":true` in
application.json


drewfish:
This was branched from 0.5.0pr2 (though it appears to apply to develop-perf just fine).

caridy:
Instead of hacking with _stringify, we might use `util.inspect` instead.

http://nodejs.org/docs/latest/api/util.html#util.inspect

caridy:
+1

PR #683, opened 2012-10-30
Adding support for handlebar partials

caridy:
# Example:

```
./MojitName/views/
    foo.hb.html
./MojitName/views/partials/
    bar.hb.html
```

by following that structure, you should be able to define `foo.hb.html` like this:

```
<div id="{{mojit_view_id}}">
    {{> bar}}
</div>
```

Few more notes:
- a partial view can be used by any view within the same context/mojit.
- global partial views are also supported.
# Details:
- adding support for partials in the store. there is now `instance.partials` with a similar structure than `instance.views` after expanding the instance. 
- views within `partials/*` folder are now consider partials, which are an special type of view.
- adding support for `application.json->viewEngine->preloadTemplates` to preload any view in memory, and this is disabled by default.
- consolidating `application.json->viewEngine->cacheTemplates` as a static config across the board.
- adding support for preloadTemplates for HB client and server to avoid loading the template if is is part of the expanded instance.
- adding support for partials on hb client and server.
- `view-renderer` is now a class and a factory to avoid creating renderer objects over and over again.
- removing viewId argument that is not really relevant for this abstraction.
- using the new `Y.mojito.ViewRenderer()` factory to access the renderer instance.
- removing unnecesary dependencies (`mojito-perf`).
- removing unnecessary configuration that is not semantically correct for `action-context` or `mojito-client` like `cacheTemplates` and `pathToRoot`.


ericf:
LGTM.

@caridy What's the reason that nested partials do not work?

caridy:
@ericf that's a good question. As today, whenever you call to render a template in mojito you will pass what we call a viewObject (yes, probably not the best name), and this object represent a `*.<engine>.html` resource, and you also pass the mojit instance, so the renderer knowns a little bit more about the actual resource. Part of this process is to ask for `partials` based on the mojit instance information, and those are compiled, cached and passed as a reference while executing the compiled template. If the engine is smart enough to support nested at that point, I don't know :)

ericf:
@caridy Based on what you have implemented here, nested partials should work just fine. Here's a very simple example:
http://jsbin.com/omital/2/edit

caridy:
Perfect.

drewfish:
Hi everyone,

Drew

rwaldura:
Was redone as #842.

PR #687, opened 2012-10-31
fix instances of appConfig.yui.config.config

drewfish:
... which are mysterious and archaic.


drewfish:
rats, wrong branch :(

PR #689, opened 2012-10-31
yui sandbox fixes

caridy:
- extending the sandbox scope to include clearTimeout, clearInterval and JSON (bugfix for search). 
- passing Y.config into the loader initialization routine when using a sandbox to preserve the filter and other properties.


drewfish:
+1

PR #691, opened 2012-10-31
fixed bleed between loader and our use of its results

drewfish:


caridy:
+1 even though this is a requirement from tests. Probably we should clone/copy at the test level and not at the app level, but in any case, this is a one time operation.

PR #694, opened 2012-11-01
provide new archetype with more ycb compliant dimensions.json for yahoo dev

lzhan:
The new archetype is for yahoo dev. The default dimensions.json in the yahoo archetype is more ycb compliant.


caridy:
+1

PR #697, opened 2012-11-02
Static handler and combo consolidation

caridy:
- combo is now part of the static handler middleware
- all static files are now combo-ables
- cache is now enable at the file level
- experimental combo-handler middleware removed


caridy:
One of the things is happening at the moment is that all requests are treated as potential combo or static requests, this is clearly an overhead. In theory, we should be able to leverage `appConfig.staticHandling.prefix` to validate that the url is a potential static url to avoid doing extra operations everytime, the problem is that `prefix` if a default value of the store and not a default value of the configs.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -161,9 +156,128 @@ function clearCache(key) {
...
+    Y.use('loader');
+    modules = (new Y.Loader(Y.config)).moduleInfo;
+
+    for (name in modules) {
+        if (modules.hasOwnProperty(name)) {
+            // faking a RS object for the sake of simplicity
+            fullpath = libpath.join(__dirname,
+                '../../../node_modules/yui', modules[name].path);
+            mimetype = libmime.lookup(fullpath);
+            charset  = libmime.charsets.lookup(mimetype);
```
drewfish, line 265: Since we know these YUI modules are javascript files, can we hardcode the MIME type and charset?

drewfish:
I see this in the travis output for the unit tests:

```
[ERROR] LibManager - Error: ENOENT, no such file or directory '../../../../../lib/app/middleware/mojito-combo-handler.js'
1666true
1667hook
1668Loading dependency: ../../../../base/mojito-test.js
1669Loading dependency: /home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/lib/common/yui-arrow.js
1670Executing test: ./test-combo-handler.js
1671
1672module.js:340
1673    throw err;
1674          ^
1675Error: Cannot find module '/home/travis/builds/yahoo/mojito/tests/unit/lib/app/middleware/test-combo-handler.js'
1676    at Function.Module._resolveFilename (module.js:338:15)
1677    at Function.Module._load (module.js:280:25)
1678    at Module.require (module.js:362:17)
1679    at require (module.js:378:17)
1680    at Object.ARROW.onSeeded (/home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/nodejs/node.js:93:9)
1681    at onYUIAvailable (/home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/lib/client/yuitest-seed.js:133:15)
1682    at /home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/lib/client/yuitest-seed.js:140:13
1683    at Object.<anonymous> (/home/travis/builds/yahoo/mojito/node_modules/yahoo-arrow/lib/client/yuitest-seed.js:152:3)
1684    at Module._compile (module.js:449:26)
1685    at Module._extensions..js (module.js:467:10)
```

drewfish:
It looks like `test-handler-static.js` has a negative combo test (in "bad or missing files") but it doesn't appear to have a positive combo test.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -182,12 +296,21 @@ function makeContentTypeHeader(res) {
...
         maxAge = options.maxAge,
-        urls = store.getAllURLResources();
+        staticRess,
+        moduleRess,
+        yuiRess;
+
+    logger = globalLogger;
+
+    moduleRess = getAppModuleResources(store);
+    yuiRess = getYUIModuleResources(appConfig);
```
drewfish, line 312: This appears to be mixing apples and oranges.  The `getYUIModuleResources()` method _looks_ like it's returning resources, but it really is returning YUI loader config info.  So I think it should either return resources (real or faked), or `yuiRess`, `getYUIModuleResources()`, and the `res` of `res: yuiRess[module]` (later on) should be renamed.

lib/app/middleware/mojito-handler-static.js
caridy, line 265: they could be CSS since YUI uses yui modules meta for css as well.

lib/app/middleware/mojito-handler-static.js
caridy, line 312: Hmm, didn't quite get what you mean. Both methods return a object in the same form of {moduleName: res}, in the case of yuiRess, they are fake "resources".

caridy:
@drewfish I remember we removed a file from "middleware/test-handler-static.js" as some point.

lib/app/middleware/mojito-handler-static.js
drewfish, line 312: Oh sorry, I didn't see that `getYUIModuleResources()` was returning fake resources.  Yeah, it's fine as is.

drewfish:
In the "valid combo url" test, the comment "combining an existing file with an invalid one should trigger 400" is wrong I think.

Otherwise, +1.

PR #702, opened 2012-11-02
app group consolidation

caridy:
- bypassing combine and base definition from yui->config when running in the server side to avoid side effects of the loader trying to load stuff from CDN when running on the server side.
- removing appConfig->yui from the client runtime init structure since it is already set by YUI.applyConfig.
- consolidating yui.config.groups.app setting.
- defining a way for other RS addons to control it through store.yui.getAppGroupConfig method.


drewfish:
+1

PR #705, opened 2012-11-05
[docs] Removed binder tests that prevented unit tests from running, corrected the server.js so apps can run on hosting containters.

zhouyaoji:


lzhan:
+1

PR #712, opened 2012-11-06
modify run.js so that when context param is null, no --context flag in mojito command

lzhan:
"mojito start --context" gives error "Invalid command line". Modify run.js to avoid this error when starting apps.  Also, when there is no port number is specified for an app in apps.json, the app should be started with port specified in app configs or using mojito default.


isao:
+1

PR #714, opened 2012-11-06
meta data and combo consolidation for shaker and offline apps

caridy:
- removing support for rollups from store. 
- consolidate path attribute for all base and resolved metadata for better integration with offline apps. 
- bugfix for failures when the first file in a combo was a synthetic file without stats.
- static seed files without combo when combine is false
- store.yui.getAppGroupConfig method is the hook for RS addons to control the composition of the seed files array

### DIFF
--- a/lib/app/addons/rs/url.js
+++ b/lib/app/addons/rs/url.js
@@ -118,7 +118,6 @@ YUI.add('addon-rs-url', function(Y, NAME) {
                 ress = store.getResourceVersions({mojit: mojit});
                 for (r = 0; r < ress.length; r += 1) {
                     res = ress[r];
-
                     skip = false;
                     if ('config' === res.type) {
                         skip = true;
@@ -171,7 +170,7 @@ YUI.add('addon-rs-url', function(Y, NAME) {
         _calcResourceURL: function(res, mojitRes) {
             var fs = res.source.fs,
                 relativePath = fs.fullPath.substr(fs.rootDir.length + 1),
-                urlParts = [];
+                urlParts = ['/', this.config.prefix];
 
             // Don't clobber a URL calculated by another RS addon, or bother to
             // proceed for server affinity resources that don't need uris
@@ -181,14 +180,13 @@ YUI.add('addon-rs-url', function(Y, NAME) {
 
             if (res.yui && res.yui.name) {
                 // any yui module in our app will have a fixed path
-                // that hash to be really short to optimize combo
-                // urls as much as possible.
-                res.url = res.yui.name + '.js';
+                // that has to be really short to optimize combo
+                // urls as much as possible. This url will also work
+                // in conjuntion with "base", "comboBase" and "root"
+                // from application.json->yui->config
+                urlParts.push(res.yui.name + '.js');
+                res.url = libpath.join.apply(libpath.join, urlParts);
                 return;
-            }
-
-            if (this.config.prefix) {
-                urlParts.push(this.config.prefix);
             }
 
             if ('shared' === res.mojit) {
@@ -211,13 +209,13 @@ YUI.add('addon-rs-url', function(Y, NAME) {
 
             if ('mojit' === res.type) {
                 if ('shared' !== res.name) {
-                    res.url = '/' + urlParts.join('/');
+                    res.url = libpath.join.apply(libpath.join, urlParts);
                 }
                 return;
             }
 
             urlParts.push(relativePath);
-            res.url = '/' + urlParts.join('/');
+            res.url = libpath.join.apply(libpath.join, urlParts);
         }
 
 
--- a/lib/app/addons/rs/url.js
+++ b/lib/app/addons/rs/url.js
@@ -4,7 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
-/*jslint anon:true, sloppy:true, nomen:true, stupid:true*/
+/*jslint anon:true, sloppy:true, nomen:true, stupid:true, node:true */
 /*global YUI*/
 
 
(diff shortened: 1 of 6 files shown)


lib/app/addons/rs/url.js
```diff
@@ -177,18 +170,23 @@ YUI.add('addon-rs-url', function(Y, NAME) {
         _calcResourceURL: function(res, mojitRes) {
             var fs = res.source.fs,
                 relativePath = fs.fullPath.substr(fs.rootDir.length + 1),
-                urlParts = [],
-                rollupParts = [],
-                rollupFsPath;
+                urlParts = ['/', this.config.prefix];
```
drewfish: I don't think this is right.  The trick is that `config.prefix` might be empty, as we need to support this use-case (or at least we used to).

lib/app/addons/rs/url.js
```diff
@@ -201,42 +199,23 @@ YUI.add('addon-rs-url', function(Y, NAME) {
...
-                    rollupParts.push(res.mojit);
-                    rollupParts.push('rollup.client.js');
-                    rollupFsPath = libpath.join(mojitRes.source.fs.fullPath, 'rollup.client.js');
-                }
             }
 
             if ('mojit' === res.type) {
                 if ('shared' !== res.name) {
-                    res.url = '/' + urlParts.join('/');
+                    res.url = libpath.join.apply(libpath.join, urlParts);
```
drewfish: I don't think we want to use libpath.join() for URLs.  If we did, if the server is windows we'd generated URLs like `/foo\bar\baz` which is really weird.

lib/app/addons/rs/url.js
```diff
@@ -201,42 +199,23 @@ YUI.add('addon-rs-url', function(Y, NAME) {
...
 
             urlParts.push(relativePath);
-
-            if (rollupFsPath && (this.assumeRollups || existsSync(rollupFsPath))) {
-                res.url = '/' + rollupParts.join('/');
-                fs.rollupPath = rollupFsPath;
-            } else {
-                res.url = '/' + urlParts.join('/');
-            }
+            res.url = libpath.join.apply(libpath.join, urlParts);
```
drewfish: I don't think we want to use libpath.join() for URLs.  If we did, if the server is windows we'd generated URLs like `/foo\bar\baz` which is really weird.

lib/app/addons/rs/yui.js
```diff
@@ -96,16 +132,25 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
         initializer: function(config) {
             this.appRoot = config.appRoot;
             this.mojitoRoot = config.mojitoRoot;
+
+            // for all synthetic files, since we don't have an actual file, we need to
+            // create a stat object, in this case we use the mojito folder stat as
+            // a replacement. We make it syncronous since it is meant to be executed
+            // once during the preload process.
+            syntheticStat = libfs.statSync(libpath.join(__dirname, '../../../..'));
```
drewfish, line 140: Neat idea, and probably OK.

My worry is that the `syntheticStat` probably marks the resource as a directory.  The way we use this stat structure today it's probably no big deal, but it might lead to interesting surprises later.

I have a slight preference for saying that "synthetic files don't have a 'stat' structure" and then make sure that trickles into all parts of the system (static handler in this case).

But, the above code will work fine for now.

lib/app/addons/rs/yui.js
```diff
@@ -368,90 +482,56 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                 langs.push('');
             }
 
-            res = {
+            this.appModulesRess['loader-app'] = res = {
```
drewfish: I thought that jslint complained about `a = b = ...`.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -296,21 +189,17 @@ function getYUIModuleResources(appConfig) {
...
+        yuiRess        = store.yui.getYUIURLResources(),
+        staticRess     = store.getAllURLResources(),
 
-    moduleRess = getAppModuleResources(store);
-    yuiRess = getYUIModuleResources(appConfig);
-    staticRess = store.getAllURLResources();
+        options = appConfig.staticHandling || {},
+        cache   = options.cache,
+        maxAge  = options.maxAge,
+        staticPath = libpath.join('/', (options.prefix || 'static'), '/'),
```
drewfish: If `staticPath` is used for URLs, it should not use libpath (which will create weird URLs with \ in them if the server is on windows).

lib/app/addons/rs/url.js
caridy: It is not empty anymore since I have a default value in `config.json`. but I do agree that `prefix` in its current form is not semantically correct. In my mind, `prefix` means a prefix path, not a prefix word. We can attempt to change that as well but it will have BC implications.

lib/app/addons/rs/url.js
caridy: Good point.

lib/app/addons/rs/url.js
drewfish: There were users (Livestand at least) who didn't want a prefix at all.  They'd put `staticHandling: { prefix: '' }` in the app.json, and we needed to be sure to do handle that.

I think the CocktailsRT stuff is based directly off the old Livestand code, so might need to `prefix:''` as well.

lib/app/addons/rs/yui.js
```diff
@@ -211,14 +256,77 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
             return Y.merge({
                 combine: (this.yuiConfig.combine === false) ? false : true,
                 maxURLLength: this.yuiConfig.maxURLLength || 1024,
-                base: "/static/",
-                comboBase: "/static/combo?",
+                base: "/.",
+                comboBase: "/combo~",
```
drewfish, line 260: Should comboBase be appConfig.staticHandling.prefix + '/combo~' ?

caridy, line 260: It was considered, but the thing is that it will not be consistent with `/static/combo~/static/foo.js~/static/bar.js`.

drewfish:
+1

PR #720, opened 2012-11-08
optimizing the cache key for instances by relying on type and base only

caridy:
this should fix any memory allocation when using id as part of the instance object before expanding it. also, shaker does not use action anymore, so we are removing it.


drewfish:
+1

PR #721, opened 2012-11-08
support for custom version of yui at the app level

caridy:
- by using `npm i yui@<some-version>` in the app folder, at the same level then `mojito`, it will force mojito to rely on that version of `yui`.
- a warning message will be issued to announce the detection of the new `yui` version.
- this version will affect server and client.
- for customization of the client side only for online apps, you can use application.json->yui->config settings instead.

TODO:
- test this in manhattan
- add unit test

### DIFF
--- a/lib/yui-sandbox.js
+++ b/lib/yui-sandbox.js
@@ -2,17 +2,32 @@
 
 'use strict';
 
-var fs = require('fs'),
-    path = require('path'),
-    vm = require('vm'),
-    code = {
-        min: fs.readFileSync(path.join(__dirname, '..',
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs-min.js'), 'utf8'),
-        raw: fs.readFileSync(path.join(__dirname, '..',
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs.js'), 'utf8'),
-        debug: fs.readFileSync(path.join(__dirname, '..',
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs-debug.js'), 'utf8')
-    };
+var libfs   = require('fs'),
+    libpath = require('path'),
+    libvm   = require('vm'),
+    appRoot = process.cwd(),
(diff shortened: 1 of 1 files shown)


mojit0:
+1

lib/yui-sandbox.js
```diff
@@ -2,17 +2,32 @@
...
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs-min.js'), 'utf8'),
-        raw: fs.readFileSync(path.join(__dirname, '..',
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs.js'), 'utf8'),
-        debug: fs.readFileSync(path.join(__dirname, '..',
-            'node_modules', 'yui', 'yui-nodejs', 'yui-nodejs-debug.js'), 'utf8')
-    };
+var libfs   = require('fs'),
+    libpath = require('path'),
+    libvm   = require('vm'),
+    appRoot = process.cwd(),
```
drewfish, line 8: This is the part that I suspect will fail in Manhattan.  In Manhattan the server handles multiple apps, and so doesn't cd to any particular app directory at startup.

ItsAsbreuk:
I hope this also involves custom base location at the client side other than used on the client side.
Because situations may exist where clientside loads from a custom comboloader through ssl. The serverside doesn't need to (preferably not).

Marco.

caridy:
@ItsAsbreuk that's already covered with the new yui->config->seed and yui->config->base / comboBase / root, they are not affecting the server side execution, only the client side. I showed an example during my yuiconf presentation, here is the link of the gist:

https://gist.github.com/4070536#file_custom_seed_2.json

ItsAsbreuk:
Caridy, you are great!
I haven;t seen your presentation yet. I've started looking at your sheets (http://www.slideshare.net/caridy/yuiconf2012-mojito-for-yui-developers) right now :)

Regards,
Marco.

drewfish:
Hi everyone,

Drew

PR #722, opened 2012-11-08
fix bug 4404935, add yaml suport, including work from pr580 and pr670

lzhan:
Revert pr 693. As js-yaml and it dependent packages are whitelisted.


isao:
+1 (again) now that the yaml lib is whitelisted

PR #727, opened 2012-11-09
two minor changes: use latest arrow,  modify run.js handle no port num specified case

lzhan:


tests/run.js
```diff
@@ -307,7 +307,9 @@ function runMojitoApp (basePath, path, port, params, callback) {
     params = params || '';
     console.log('Starting ' + path + ' at port ' + port + ' with params ' + (params || 'empty'));
     var p;
-    if (params === '') {
+    if (!port) {
+        p = runCommand(basePath + '/' + path, cwd + "/../bin/mojito", ["start"], function () {});
+    } else if (params === '') {
         p = runCommand(basePath + '/' + path, cwd + "/../bin/mojito", ["start", port], function () {});
```
drewfish: It might be better to create a separate cmdArgs array, and push things to it as needed.  Something like:

``` javascript
var cmdArgs = ['start'];
if (port) {
    cmdArgs.push(port);
}
if (params) {
    cmdArgs.push('--context');
    cmdArgs.push(params);
}
...
p = runCommand(basePath + '/' + path, cwd + "/../bin/mojito", cmdArgs, function () {});
```

isao: good suggestion.

+1 for now assuming you need this in for something else.

[commander](https://github.com/visionmedia/commander.js), the cli option parser used supports [default values](http://tjholowaychuk.com/post/9103188408/commander-js-nodejs-command-line-interfaces-made-easy) as well. You may want to explicitly define defaults and initial values with it's api first and save a few `if(foo)`'s

[edit] my comment doesn't actually apply to your change.. I assumed port  and context were a cli params... srry

lzhan:
Thanks Drew and Isao. I will update run.js.

isao:
+1

PR #738, opened 2012-11-12
update run.js to add --descriptor and --config

lzhan:


lzhan:
Thanks Isao. It is good to know. But in this case, only yui 3.4.1 can be used for func tests to avoid a extra insertion caused by later version yui.

isao:
+1

PR #742, opened 2012-11-12
Composite dispatch on the client

caridy:
- moving the logic to attach controllers, view engines and other requirements on demand during the dispatch rather than mojito-client to support composition on demand. 
- as a side effect, we can have a tunnel client at the server runtime as well to delegate the dispatch to another runtime
- tunnel client is now available in server and client runtime to delegate commands


lib/app/autoload/dispatch.common.js
```diff
@@ -35,16 +39,128 @@ YUI.add('mojito-dispatcher', function(Y, NAME) {
...
+                adapter.error(new Error('RPC tunnel is not available in the ' +
+                    command.context.runtime + ' runtime.'));
+
+            }
+        },
+
+        /**
+         * Dispatch a command in the current runtime, or fallback
+         * to a remote runtime when posible.
+         * @method prc
```
drewfish: Wrong method name I think.

lib/app/autoload/dispatch.common.js
```diff
@@ -35,16 +39,128 @@ YUI.add('mojito-dispatcher', function(Y, NAME) {
...
+                Y.log(e.message, 'error', NAME);
+                Y.log(e.stack, 'error', NAME);
+                adapter.error(e);
+            }
+            perf.done(); // closing the 'ac:ctor' timeline
+        },
+
+        /**
+         * Executes a command in a remote runtime if possible.
+         * @method prc
```
drewfish: I think this should be `@method rpc`.

drewfish:
+1

PR #746, opened 2012-11-13
fix bz 5914052: do real buffer calculus

drewfish:


lib/app/middleware/mojito-handler-static.js
```diff
@@ -239,9 +240,17 @@ function staticProvider(store, logger, Y) {
                 }
             }
 
+            len = 0;
```
caridy, line 245: Can you add a block of comment explaining the details about this routine, and why we need to do the two cycles to pre-compute the length?

caridy:
+1

PR #749, opened 2012-11-13
updated archetypes for 0.5

drewfish:


lib/app/archetypes/app/full/application.json
```diff
@@ -4,30 +4,21 @@
...
-            "client": {
-                "level": "info",
-                "yui": false
-            },
-            "server": {
-                "level": "info",
-                "yui": false
-            }
-        },
         "shareYUIInstance": false,
```
caridy: this is not used anymore.

lib/app/archetypes/app/full/application.json
```diff
@@ -4,30 +4,21 @@
...
         "staticHandling": {
             "appName": "app",
             "cache": false,
```
caridy: this should probably go into the staticHandling on environment:dev section as false, and in here as true.

lib/app/archetypes/app/full/application.json
```diff
@@ -4,30 +4,21 @@
 
         "appPort": 8666,
         "cacheViewTemplates": true,
-        "deferAllOptionalAutoload": false,
         "embedJsFilesInHtmlFrame": false,
```
caridy: is this property still used?

caridy:
+1

PR #753, opened 2012-11-14
guard against empty module info return value

isao:
Should prevent the following error, referenced in the favicon trello task

```
  TypeError: Cannot read property 'name' of undefined

  at RSAddonYUI.getResourceContent (/home/y/share
  node/manhattan_app_search_apps_news/node_modules
  /mojito/lib/app/addons/rs/yui.js:573:31)
```

also rm sloppy jslint config, as it's no longer needed


drewfish:
+1

PR #762, opened 2012-11-16
[code_exs] Updated developer-guide code exs for 0.5:

zhouyaoji:
- Removed 'init' methods from controllers, obsolete app configs.
- Required addons in controllers.
- Changed syntax for accessing models from controllers.
- Removed mojit tests directory.


isao:
was your branch based on develop or develop-perf? 

+1 for now

PR #763, opened 2012-11-17
resolving yui module name collision

caridy:
- solve the case where a developer wants to override mojito-client or any other client or common affinity file defined by mojito core.


caridy:
@drewfish what I don't understand is why using unshift or push for the "shared" mojit when collecting al ress will not make a difference. Can you look at this? Probably we should add some tests here.

drewfish:
+1

PR #764, opened 2012-11-19
splitting dispatch logic

caridy:
- client and server controllers have different nature (stateless vs long live)
- this solves the problem of calling the actions in the same mojit in the client side, in which case the same instance and dispatcher should be used.


drewfish:
It looks like invoking an action on a composite mojit is failing on the client side.

Steps to reproduce
- In tests/func/applications/frameworkapp/common do a `../../../../../bin/mojito start` to start the app.
- Visit http://localhost:4081/#MojitContainer in a browser.
- Click the `MyMojits` button.
- The `Output` block should contain `My Child Mojits:` with two mojits, however none are shown and errors are in the error console.

drewfish:
+1

PR #765, opened 2012-11-19
fix `mojito test mojit` and newsboxes tests

drewfish:


lib/management/yui-module-configurator.js
```diff
@@ -118,7 +118,7 @@ module.exports = function(dir, excludes) {
         return modules;
     }
 
-    if (Y.Lang.isArray(dir)) {
+    if (!Y.Lang.isArray(dir)) {
```
caridy, line 121: This condition can be replaced with:

`dir = Y.Array(dir);`

which guarantee an array as a result.

caridy:
+1

PR #767, opened 2012-11-20
Files cleanup

caridy:
removing deprecated components:
- logger
- loader
- output-adapter
- controller-context
- resource-store-adapter
- server-log
- setLog\* from mojito.js


isao:
+1

drewfish:
Perhaps also `lib/server-log.js`?

Perhaps also the `setLog*()` methods in MojitoServer?

Is `lib/app/autoload/json-stringify-hack.server.js` actually being used?

caridy:
The hack for json-stringify is still needed. I ran this gist:
https://gist.github.com/4119541

And the results are:

$ node json.js 

stringify native:  184170
Y.JSON.stringify:  375190

parse native:  187691
Y.JSON.parse:  186450

As you can see, stringify is still not using the native under the hood, should be ready in 3.8.0 though.

drewfish:
I didn't see anyone depending on "json-stringify-hack", so I'm not sure we're even using it.

Regardless, +1.

caridy:
@drewfish yes, it is used on the server side because all available files are attached to Y, no matter if they are required or not :)

PR #768, opened 2012-11-20
rename TunnelClient to TunnelCommon everywhere

drewfish:
Also fixed the RPC to server.

### DIFF
--- a/lib/app/autoload/dispatch.client.js
+++ b/lib/app/autoload/dispatch.client.js
@@ -31,7 +31,7 @@ YUI.add('mojito-dispatcher', function (Y, NAME) {
          * @method init
          * @public
          * @param {Y.mojito.ResourceStore} resourceStore the store to use.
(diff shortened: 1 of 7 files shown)


isao:
+1

caridy:
I'm not so sure I like that name. In my mind, the tunnel is still a "client" in a sense that it is suppose to ack like a client to a remote mojito server.

drewfish:
I see.  It's affinity is "common" however.

drewfish:
I'll rename it to just "Tunnel".  (It -is- a client, as in "mysql client" or "http client", but "client" can get confused with our affinity:client.)

drewfish:
I'll also add a note to the `delete command.rpc` that it might need to move to a different place in the future.

drewfish:
Actually, TunnelClient does make more sense once I look at the context more.  Closing this, will issue a new PR with necessary other changes.

PR #769, opened 2012-11-20
removing hack to use a custom yui-test script

caridy:
this should fix all the func tests failures due to the problem to collect results.


drewfish:
+1

idearat:
+1

PR #771, opened 2012-11-20
0.5-inspired cleanups

drewfish:
... mainly in `examples/` and `tests/`


examples/performance/string/application.json
```diff
@@ -1,14 +1,9 @@
...
-			"level": "none",
-			"yui": false
-		}
-	}
-}]
\ No newline at end of file
+    "yui": {
+        "config": {
+            "debug": false,
+            "logLevel": "error"
```
caridy, line 6: if 'debug' is false, logLevel is irrelevant. In this particular case, the translation from "log" to yui->config will be having debug: false, without logLevel.

btw, debug: false is the default setting :)

tests/fixtures/gsg5-appConfig/application.json
```diff
@@ -11,18 +9,8 @@
...
-        "log": {
-            "client": {
-                "level": "warn",
-                "yui": false
-            },
-            "server": {
-                "level": "warn",
-                "yui": false
-            }
+            "base": "/foo/",
```
caridy: what's the purpose of redefining `base` here? I guess u will have to use `combine: false` otherwise it will continue use `comboBase` instead.

tests/fixtures/gsg5-appConfig/application.json
```diff
@@ -11,18 +9,8 @@
...
+            "logLevel": "warn"
```
caridy: If you want to get logLevel to warn, remember to add `debug: true` as well.

tests/fixtures/gsg5-appConfig/application.json
drewfish: No idea, this was in the previous version of the code too.  I can remove it if desired.  I'm pretty sure it's not being used anyways.  (The fixtures are only used for unit tests, and the one unit tests which loads gsg5-appConfig doesn't test for "foo".)

caridy:
+1

PR #777, opened 2012-11-21
rm sections supporting run app parallelly as it causes hanging on CI

lzhan:
This can be revisited if necessary. The hanging may caused by failed application start.

### DIFF
--- a/tests/run.js
+++ b/tests/run.js
@@ -133,7 +133,7 @@ function runUnitTests (cmd, callback) {
     var commandArgs = [
         cwd + "/../node_modules/yahoo-arrow/index.js",
         "--descriptor=" + cmd.unitPath + "/**/*_descriptor.json",
-        "--coverage=true",
+        "--report=true",
         "--reportFolder=" + arrowReportDir
     ];
     if ('phantomjs' !== cmd.unitBrowser) {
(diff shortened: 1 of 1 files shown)


tests/run.js
```diff
@@ -318,37 +321,22 @@ function runMojitoApp (cliOptions, basePath, path, port, params, callback) {
...
-    if (cliOptions.debugApps) {
-        p.stdout.on('data', function(data) {
-            console.error('---DEBUG ' + port + ' STDOUT--- ' + data.toString());
-        });
-        p.stderr.on('data', function(data) {
-            console.error('---DEBUG ' + port + ' STDERR--- ' + data.toString());
-        });
-    }
+    // Give each app a second to start
+    setTimeout(function () { callback(null) }, 1000);
```
caridy, line 346: what's the reason for this line? is that 1000 an arbitrary number?

tests/run.js
```diff
@@ -318,37 +321,22 @@ function runMojitoApp (cliOptions, basePath, path, port, params, callback) {
...
         cmdArgs.push('--context');
         cmdArgs.push(params);
-    }
-    p = runCommand(basePath + '/' + path, cwd + "/../bin/mojito", cmdArgs, function () {});
+    } 
+    var p = runCommand(basePath + '/' + path, cwd + "/../bin/mojito", cmdArgs, function () {});
+    
     pids.push(p.pid);
     pidNames[p.pid] = libpath.basename(path) + ':' + port + (params ? '?' + params : '');
-    p.stdout.on('data', listener);
```
caridy, line 343: I will like to preserve the output options through the cliOptions

tests/run.js
```diff
@@ -175,8 +176,9 @@ function deploy (cmd, callback) {
         apps = appsConfig.applications;
 
     for (var i=0; i<apps.length; i++) {
-        (function (app) {
```
caridy, line 178: Any particular reason for reverting the changes in this block?

lzhan:
Added cliOption back. The listener part is causing problem when there is any app failed to start. (For our case, there is an example app is failed to start now.) When there is an app failed to start, run.js should continue to start the rest of the apps and run all the tests. The approach is right. The code needs to be modified and checked in later.

caridy:
+1

PR #778, opened 2012-11-21
Micro optimization

caridy:
The intent is to do a quick review across the core modules to try to optimize them for V8, specially the low hanging fruits. So far, we have two things we want to optimize:
- "delete" is evil in V8, avoid the use of delete on objects.
  for more details, check this gist:
  https://gist.github.com/4126067
- "try/catch" is slow, avoid the use of it when possible.


drewfish:

Issue #779, opened 2012-11-21
mojito compile doesn't work with custom view engines


isao:
on 0.5.x branch `mojito compile views` does not access the proper view engine

isao:
compile cmd is deprecated

PR #780, opened 2012-11-21
fix `mojito compile` fatal error

isao:
minimal fix for getting a mu or hb view renderer.
- removed store.preload() calls that were made
  redundant by store.js:createStore()
- add return callback to exit on error.
- copy utils func from develop

[trello issue](https://trello.com/card/cli-mojito-compile-is-broken-in-pr4/504e5e3b7cd1578456fc430f/150)

n.b. I'm not sure that custom view renderers were loaded successfully before, but they aren't now (issue #779).

### DIFF
--- a/lib/app/commands/compile.js
+++ b/lib/app/commands/compile.js
@@ -42,7 +42,7 @@ var libpath = require('path'),
     options,
     run,
     YuiModuleCacher,
-    Y = libutils.yuiuse({
+    Y = libutils.getYUIInstance({
         'async-queue': null,
         'json-parse': null,
         'json-stringify': null,
--- a/lib/app/commands/compile.js
+++ b/lib/app/commands/compile.js
@@ -12,9 +12,12 @@
 var libpath = require('path'),
     libfs = require('fs'),
     existsSync = libfs.existsSync || libpath.existsSync,
-    libutils = require(libpath.join(__dirname, '../../management/utils')),
-    Mojito = require(libpath.join(__dirname, '../../mojito')),
-    Store = require(libpath.join(__dirname, '../../store')),
+
+    // paths
+    BASE = libpath.resolve(__dirname, '../../../') + '/',
+    libutils = require(BASE + 'lib/management/utils'),
+    Mojito = require(BASE + 'lib/mojito'),
+    Store = require(BASE + 'lib/store'),
 
     // private compilation function container
     compile = {},
@@ -39,10 +42,13 @@ var libpath = require('path'),
     options,
     run,
     YuiModuleCacher,
-    Y = require('yui').YUI({useSync: true}).use('json-parse', 'json-stringify', 'async-queue');
-
-Y.applyConfig({useSync: false});
-
+    Y = libutils.yuiuse({
+        'async-queue': null,
+        'json-parse': null,
+        'json-stringify': null,
+        'mojito-hb': BASE + 'lib/app/addons/view-engines/mu.server.js',
+        'mojito-mu': BASE + 'lib/app/addons/view-engines/hb.server.js'
+    });
 
 usage = 'mojito compile {options} {type}\n' +
     '\nOPTIONS: \n' +
@@ -506,8 +512,6 @@ compile.rollups = function(context, options, callback) {
         shortDest,
         rollupBody;
 
-    store.preload();
-
     libutils.log((options.remove ? 'Removing' : 'Creating') + ' rollups...');
 
     if (options.app || options.core) {
@@ -615,7 +619,6 @@ compile.views = function(context, options, callback) {
         store = makeStore({root: cwd}),
         compiledFilename = '/autoload/compiled/views.common.js',
         mojits,
-        yuiConfig,
         renderer,
         renderedView,
         action = options.remove ? 'Removed' : 'Created',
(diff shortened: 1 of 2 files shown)


lib/app/commands/compile.js
```diff
@@ -651,12 +651,11 @@ compile.views = function(context, options, callback) {
             mojitNs = mojitName.replace(/\./g, '_'),
             yuiModuleCacheWriter,
             viewName,
-            MojY,
             compileFunction;
 
         mojitRes = store.getResources('server', context, {type: 'mojit', name: mojitName});
         if (!mojitRes || !mojitRes.length) {
-            callback('Unknown mojit "' + options.mojit + '"');
+            return callback('Unknown mojit "' + options.mojit + '"');
```
caridy: Use:

```
callback(...)
return;
```

it is confusing to have a return of the result of a callback, specially when the current method should return null. Not need to same one line ;)

lib/app/commands/compile.js
```diff
@@ -39,10 +42,13 @@ var libpath = require('path'),
     options,
     run,
     YuiModuleCacher,
-    Y = require('yui').YUI({useSync: true}).use('json-parse', 'json-stringify', 'async-queue');
-
-Y.applyConfig({useSync: false});
-
+    Y = libutils.yuiuse({
```
caridy: I like this way of providing a utility method, but I don't like the name, specially because we are overloading YUI.use() with this definition. I probably prefer something like `libutils.yuifactory` or `libutils.getYUIInstance()` or something like that. And it is ok to reuse the same Y under the hood, YUI does that as well.

lib/management/utils.js
```diff
@@ -500,6 +500,50 @@ function contextCsvToObject(s) {
...
+        if (val) {
+            local[name] = typeof val === 'string' ? {fullpath: val} : val;
+        }
+    });
+
+    y.applyConfig({modules: local});
+    y.use(names);
+    y.applyConfig({useSync: false});
+
+    return callback ? callback(y) : y;
```
caridy: Do not overload return, replace it with:

```
if (callback) {
     callback(y);
}
return y;
```

lib/management/utils.js
```diff
@@ -500,6 +500,50 @@ function contextCsvToObject(s) {
...
+ * @param {Object} modules A hash of module names. If value is falsey, YUI
+ *  will load it by name. If it's a string, it's the path of the module file to
+ *  load, otherwise assume it's a modules config
+ * @param {String} some_particular_yui optional require() param to load YUI lib
+ *  from somewhere besides default modules.path
+ * @param {Function} callback Optional for YUI().use(.. callback); Loading is
+ *  synchronous if omitted
+ * @return {Object} yui instance
+ */
+function yuiuse(modules, my_yui, callback) {
```
caridy: [duplicated note] I don't like the name, specially because we are overloading YUI.use() with this definition. I probably prefer something like libutils.yuifactory or libutils.getYInstance() or something like that. And it is ok to reuse the same Y under the hood, YUI does that as well.

lib/management/utils.js
```diff
@@ -500,6 +500,50 @@ function contextCsvToObject(s) {
...
+ *
+ * ...specify all modules and optional module configs in one go
+ *
+ *   var Y = yuiuse({oop: null, mymodule1: 'path/to/it', foo: {base:...}, ...}
+ *
+ * @method yuiuse
+ * @param {Object} modules A hash of module names. If value is falsey, YUI
+ *  will load it by name. If it's a string, it's the path of the module file to
+ *  load, otherwise assume it's a modules config
+ * @param {String} some_particular_yui optional require() param to load YUI lib
```
caridy: this is very confusing, should be `my_yui` but that one is also confusing. In theory, yui-sandbox.js will be the proper way to require YUI, and that one should be smart enough to figure out which version of YUI to use (the one bundle as a requirement of mojito) or the one installed at the app level, or the one the user specify somehow, but we are not there yet ;)

caridy:
done with the review.

lib/app/commands/compile.js
isao: While I prefer the one-line version, it does imply that `callback` returns something.

I'll adopt your change.

lib/management/utils.js
isao: [yui-sandbox.js](https://github.com/yahoo/mojito/blob/develop-perf/lib/yui-sandbox.js) is clever, but wasn't available when I wrote this, and doesn't address the issue I had. I needed to specify the path to YUI to `require` while working on the build stuff as an external command (i.e. an npm module mojito-cli-cmd-build).

That requirement doesn't apply anymore now that the code is in the mainline, so I'll take it out.

lib/management/utils.js
isao: the `callback(y)` may return, and then `return y` will be executed which is poor form. I suppose the more explicit and clear way would be this:

```
if (callback) {
    callback(y);
    return;
} else {
    return y;
}
```

...but that's 6 lines vs 1. How strongly do you feel about this?

lib/app/commands/compile.js
isao: ok, I'll change the method name to getYUIInstance()

lib/management/utils.js
caridy: YUI does always return the Y instance, even if the callback is passed for chaining purposes!

About the number of lines, not a big deal since this is server side only.

lib/management/utils.js
caridy: cool

caridy:
+1

PR #784, opened 2012-11-24
Update docs/dev_guide/intro/mojito_mvc.rst

lazybean:
I think the model function "get_photos" needs the callback function as parameter.

### DIFF
--- a/docs/dev_guide/intro/mojito_mvc.rst
+++ b/docs/dev_guide/intro/mojito_mvc.rst
@@ -1,5 +1,4 @@
 
-
 =============
 MVC in Mojito
 =============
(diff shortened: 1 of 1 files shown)


docs/dev_guide/intro/mojito_mvc.rst
```diff
@@ -97,7 +96,7 @@ instructs Mojito to load the YUI module ``yql`` for getting data.
          this.config = config;        
        },
        // Model function to get data
-       get_photos: function(flickr_query){
+       get_photos: function(callback, flickr_query){
```
caridy: `callback` should be the second argument instead to follow model pattern.

lazybean: right, thanks for the review!

caridy:
Thanks for the quick fix @lazybean, I will let @zhouyaoji to review and merge it.

zhouyaoji:
Thanks @lazybean for your pull request.

PR #786, opened 2012-11-26
fix for developerguide func tests

caridy:
- fix for developerguide-scrollview func test: using mojito binders instead of dumping YUI statements in the view to fix the scrollview example for func tests.
- fix for developerguide-yui_modules func test: doing regex is never a good idea for a test, instead, I'm adding a class in the binder that I can easily check in the test.


drewfish:
+1

PR #790, opened 2012-11-27
fix tests as some german chars are not supported

lzhan:


caridy:
+1

Note: these tests are really bad, we should not use regex as part of the assert, because it completely obscure the failures behind the regex result. we need to plan for a big refactor of all these tests.

lzhan:
Agree. Thanks for the suggestion.

PR #793, opened 2012-11-27
allow command execution for third party modules

diervo:
Before,  in order to execute a command:
command = require('mojito-cli-cmd-' + commandName);

Which is kinda wrong since it will look for the relative internal node_modules folder within mojito or to a global installation.

This way you can include a external command which is in the same folder as your app is executing.

### DIFF
--- a/lib/management/cli.js
+++ b/lib/management/cli.js
@@ -93,11 +93,12 @@ function parseArgs(args, optionInfo) {
 function main() {
     var args = process.argv.slice(2),
         commandName = (args.length === 0 ? 'help' : args.shift()),
+        currentPath = process.cwd(),
         command,
         argInfo;
 
     try {
-        command = require('mojito-cli-cmd-' + commandName);
+        command = require(libpath.join(currentPath, 'node_modules', 'mojito-' + commandName));
     } catch (e) {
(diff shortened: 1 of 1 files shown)


lib/management/cli.js
```diff
@@ -93,11 +93,12 @@ function parseArgs(args, optionInfo) {
 function main() {
     var args = process.argv.slice(2),
         commandName = (args.length === 0 ? 'help' : args.shift()),
+        currentPath = process.cwd(),
```
drewfish, line 96: Unfortunately `process.cwd()` doesn't point to the app's directory in Manhattan, so I think you'll need to use something like `libpath.join(__dirname, '..', '..')`.

diervo:
I think manhattan don't use mojito <command> at all, and screwdriver uses the compilers in order to execute for example Shaker.

drewfish:
Ah, right, the cli isn't used in Manhattan.

caridy:
Isn't there a property for all cli that computes the path of the app? that's what you're looking here right? If I recall correctly, that's something available as today.

drewfish:
Hi @dferreiroval,

Drew

PR #795, opened 2012-11-27
whitelist raw comboSep in yui config

drewfish:
I thought it better to take a whitelist approach, since the escaping was done for security reasons.


caridy:
+1

This will solve the issue temporary, but might have other implications down the road. I'm wondering if this is only affecting `&`, and if by replacing the encoded token right before sending it to the client for the whole string will be just enough, so we don't have to target specific items.

drewfish:
This approach whitelists all of comboSep, no matter what it contains.  If we want to whitelist `&` we can do that by changing cleanse() in lib/app/autoload/utils.common.js.

drewfish:
closing, so that I can redo this against the "develop" branch instead

PR #796, opened 2012-11-27
Added support for comments in JSON files and improved the JSLint situati...

jlecomte:
JSON does not allow for comments... Maintaining our configuration files, which are starting to get gigantic, is therefore quickly becoming a nightmare! This PR allows for comments in JSON files in Mojito by stripping those files from JavaScript-style comments before passing them to JSON.parse. Also, this PR improves a bit the situation with JSLint, although I noticed JSLint does not like the intentional fallthrough in the switch statement... I like it, personally.

### DIFF
--- a/lib/app/addons/rs/config.js
+++ b/lib/app/addons/rs/config.js
@@ -4,7 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
-/*jslint anon:true, sloppy:true, nomen:true, stupid:true*/
+/*jslint node: true, anon:true, sloppy:true, nomen:true, stupid:true, plusplus:true */
 /*global YUI*/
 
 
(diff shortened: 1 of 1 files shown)


rwaldura:
Oooooh interesting! @lzhan has done some work in this area too. 
@lzhan can you comment?

lzhan:
Mojito now supports yaml format which allows comments in the config files. Please see pr 670 for detail.

rwaldura:
see #670 (@lzhan github autolinks if you prefix with a hash sign)

drewfish:
What I've seen done is use fake keys for comments, such as...

``` javascript
{
    "//": "we'll use bar because it prevents baz from blixing",
    "foo": "bar"
}
```

That doesn't help with arrays of course.

rwaldura:
@lzhan what does Julien need to do to add comments in his config files? Just rename them to .yaml?
When was this released? Which Mojito version?

lzhan:
Need to convert the json file to yaml file(not only file name change, but also format change), and add comments.
Mojito 0.4.9 should have the support: https://github.com/yahoo/mojito/wiki/ReleaseNotes0_4_9

jlecomte:
We don't want to convert our config files to YAML at this point.

mridgway:
Another option is JSON5 (https://github.com/aseemk/json5) but I'm not sure how you would make this available on the client side.

rwaldura:
Thanks.
I thought JSON was a subset of YAML? I.e. every json file is also a valid yaml file. Is that not the case?
http://en.wikipedia.org/wiki/YAML#JSON

jlecomte:
My understanding is that, indeed, JSON is a subset of YAML since YAML version 1.2 and modern YAML parsers can parse JSON correctly. Whether that's the case in Mojito 0.4.9, I don't know... But that's not the point. If I want to share the config between the client and server, which is one of the things that Mojito excels at doing, why use a YAML parser instead of the standard JSON parser which is available on the browser? Why ship unnecessary bytes to the client? That does not make any sense!

JSON5 is nice, but I don't care about any of its features except the ability to insert comments...

caridy:
@jlecomte I don't think we are in the same page. In my mind, you will never load a yaml file from the browser, instead what mojito does it to call the store to collect the path/to/mojit/definition.json, which is not a file, but a json structure produced by the store that happens to be based on a yaml file that you have in disk, but that's all.

In other words, the store is the filesystem abstraction, and it serves json structures that are computed based on the dimensions/context, and it just happen that the url looks the same than the actual path for the file, but that's just a coincidence.

Can someone confirm that this is the actual behavior?

drewfish:
Hi @jlecomte,

Drew

PR #799, opened 2012-11-27
merge develop-perf into develop

drewfish:


drewfish:
Boss says "merge it, and we'll fix problems that come up".

PR #800, opened 2012-11-28
adding a default category (lowest in priority) if a Y.log call is issed ...

drewfish:
this fixes the case when Y.log('msg') is call without a second argument

### DIFF
--- a/lib/mojito.js
+++ b/lib/mojito.js
@@ -384,6 +384,10 @@ MojitoServer.prototype._configureLogger = function(Y, yuiConfig) {
             var c = Y.config,
                 cat = e && e.cat && e.cat.toLowerCase();
 
+            // this covers the case Y.log(msg) without category
+            // by using the low priority category from logLevelOrder.
+            cat = cat || c.logLevelOrder[0];
+
             // applying logLevel filters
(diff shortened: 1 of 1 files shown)


caridy:
+1 :p

caridy:
@isao is right, in fact logLevelOrder should be in config.json rather than in mojito.js directly. This is not part of this PR though.

zhouyaoji:
Unfortunately, It doesn't seem to fix the problem with Y.log("message w/o specifiying log level in second parameter") either. I installed 0.5.0pr5 and ran the sample logging app and again no message. 

I added 'console.log(cat)' to the 'mojito.js' seen below and saw that 'cat' is undefined when the log level is not specified.

```
  Y.on('yui:log', function (e) {
        // we want to use Y.config instead of the original yuiConfig so
        // people can switch that value on runtime for debugging
        var c = Y.config,
            cat = e && e.cat && e.cat.toLowerCase();


        console.log(cat);  // => displays 'undefined' for Y.log("just a message, no log level");
        // applying logLevel filters

        // If 'cat' is undefined, log(c, e.msg., cat, e.src) is never executed.
        if (cat && ((c.logLevel === cat) || (c.logLevelOrder.indexOf(cat) >= 0))) {
            log(c, e.msg, cat, e.src);
        }
```

PR #801, opened 2012-11-28
whitelist raw comboSep in yui config

drewfish:
See pr #795 for original discussion.


caridy:
PR #795 which includes some notes about this commit.

caridy:
+1

PR #802, opened 2012-11-28
adding a default logLevel category

caridy:
this fixes the case when Y.log('msg') is call without a second argument


caridy:
Actually, @drewfish already did this, closing this.

PR #803, opened 2012-11-28
add coverage to the git ignore list

lzhan:
Fix command "mojito test mojit -c"


lzhan:
lib/tests is for testRunner to find tests to run.(It is not for Report. Report is written under: artifacts/test/coverage). The tests were the old unit tests which don't exist anymore. If my understanding is correct, this part is the old way to run unit test coverage for mojito src.

isao:
+1

PR #804, opened 2012-11-28
Added support for JavaScript-style comments in JSON files and improved t...

jlecomte:
JSON does not allow for comments... Maintaining our configuration files, which are starting to get gigantic, is therefore quickly becoming a nightmare! This PR allows for comments in JSON files in Mojito by stripping those files from JavaScript-style comments before passing them to JSON.parse. Also, this PR improves a bit the situation with JSLint, although I noticed JSLint does not like the intentional fallthrough in the switch statement... I like it, personally. We can revisit that later if it turns out to be a problem...

This is a new PR following PR #796 which was made from the develop-perf branch which went away... @lzhan confirmed in a private email that this PR was indeed needed because it seems that files with a .json extension are not parsed by the YAML parser (which may not even support JSON)

### DIFF
--- a/lib/app/addons/rs/config.js
+++ b/lib/app/addons/rs/config.js
@@ -4,7 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
-/*jslint anon:true, sloppy:true, nomen:true, stupid:true*/
+/*jslint anon:true, sloppy:true, nomen:true, stupid:true, plusplus:true*/
 /*global YUI*/
 
 
@@ -20,7 +20,7 @@
  * @class RSAddonConfig
  * @extension ResourceStore.server
  */
-YUI.add('addon-rs-config', function(Y, NAME) {
+YUI.add('addon-rs-config', function (Y, NAME) {
 
     var libfs = require('fs'),
         libpath = require('path'),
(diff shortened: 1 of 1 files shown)


drewfish:
We do intend to support comments in config files, though after 0.5.0GA.

The latest thinking is to support YAML, which is a superset of JSON, and which supports comments.  (So, any file that is valid JSON is also valid YAML.)  One good resource for testing what YAML supports is the [online parser](http://yaml-online-parser.appspot.com/).  (Comments start with the `#` character BTW.)

jlecomte:
Drew, this would be great for us. What is the ETA to have Mojito fully support YAML?

rwaldura:
We're really close, but postponed it till after GA purposefully.

rwaldura:
0.5.1, seeing how close we are.
Sprint after 0.5 GA release.

jlecomte:
OK, if this feature request is being scheduled for 0.5.1, you may close this PR.

drewfish:
Cool.  We'll probably leave this PR open for a while, until we've done planning for 0.5.1, to make sure it doesn't get lost.

redonkulus:
Not to hijack this PR but have you guys had a chance to look at JSON5? Is a superset of JSON which supports comments, multi line, trailing commas and other things. Obviously performance and other backwards compatibility features would need to be evaluated. However, it keeps the JSON format with the added commenting feature this PR introduces. 

https://github.com/aseemk/json5

jlecomte:
@redonkulus yes, we have. Please see PR #796 for some additional background about this thread.

PR #805, opened 2012-11-28
splitting view-renderer logic into server and client

caridy:
- fix the case where mojitoProxy.render is invoked while the controller is not available in the client side
- remove the hack to load engine during dispatch, instead view-renderer does the provision when needed
- provide more flexibilities for future refactors in the client side (like moving the caching logic into view-renderer instead of specific engines, etc).


tests/unit/lib/app/autoload/test-view-renderer.server.js
```diff
@@ -8,7 +8,7 @@
 /*jslint anon:true, sloppy:true, nomen:true, node:true*/
 /*global YUI*/
 
-YUI().use('mojito-view-renderer', 'test', function(Y) {
+YUI().use('mojito-view-renderer-server-tests', 'test', function(Y) {
```
drewfish: Why `mojito-view-renderer-server-tests`?  Where is this YUI module defined?

tests/unit/lib/app/autoload/test-view-renderer.client.js
```diff
@@ -0,0 +1,105 @@
...
+ * Copyright (c) 2011-2012, Yahoo! Inc.  All rights reserved.
+ * Copyrights licensed under the New BSD License.
+ * See the accompanying LICENSE file for terms.
+ */
+
+
+/*jslint anon:true, sloppy:true, nomen:true, node:true*/
+/*global YUI*/
+
+YUI().use('mojito-view-renderer-client-tests', 'test', function(Y) {
```
drewfish: Why `mojito-view-renderer-client-tests`?  Where is this YUI module defined?

drewfish:
+1

PR #806, opened 2012-11-28
[docs-exs] Updated documentation for configuring logging and the simple logging example.

zhouyaoji:


docs/dev_guide/code_exs/simple_logging.rst
```diff
@@ -30,48 +31,83 @@ Implementation Notes
...
+indicate important framework events are occurring.
+
+
+.. _logging_notes-levels:
+
+Example Log Configuration
+-------------------------
+
+In the example ``log`` object below, you can see that you can configure the 
+levels and whether the log message has a time stamp.
```
caridy, line 70: In this paragraph you reference `log` object, which is not defined, instead it shoudl be `logLevel` property or attribute. Remove the reference to time stamp.

docs/dev_guide/code_exs/simple_logging.rst
```diff
@@ -30,48 +31,83 @@ Implementation Notes
...
+own ``yui.config`` object as shown below:
+
+.. code-block:: javascript
+
+   [
+     {
+       "settings": [ "master" ],
+       "yui": {
+         "config": {
+           "logLevel": "debug"
```
caridy, line 97: Remember that logLevel does not work if debug is not set to true.

zhouyaoji:
Good point. It's implied, but it should be stated clearly. Let me close the PR, update my branch, and make another PR. Thanks, Caridy.

docs/dev_guide/code_exs/simple_logging.rst
```diff
@@ -238,13 +258,15 @@ To set up and run ``simple_logging``:
 #. Open the URL below in a browser and look at the output from the Mojito 
    server. You should see the log messages from the controller that start 
    with the string "\[CONTROLLER]:". Notice that the two messages have 
-   different log levels.
+   different log levels: one is the default (``debug``) and the other sets
```
caridy, line 261: debug is never default I think

docs/dev_guide/intro/mojito_configuring.rst
```diff
@@ -431,22 +431,29 @@ To see all the options for the ``config`` object, see the
...
+| ``debug``            | boolean          | true                     | Determines whether ``Y.log`` messages are written to the     |    
+|                      |                  |                          | browser console.                                             |
++----------------------+------------------+--------------------------+--------------------------------------------------------------+
+| ``logExclude``       | object           | none                     | Excludes the logging of the YUI module(s) specified.         |
+|                      |                  |                          | For example: ``logExclude: { "logModel": true }``            |  
++----------------------+------------------+--------------------------+--------------------------------------------------------------+
+| ``logInclude``       | object           | none                     | Includes the logging of the YUI module(s) specified.         |
+|                      |                  |                          | For example: ``logInclude: { "searchMojit": true }``         |
```
caridy, line 444: searchMojit is probably not a good name because we don't use mojit as part of most yui modules.

I will recommend to use DemoBinderIndex and DemoModelFoo since those are the actual stuff that they will use.

docs/dev_guide/topics/mojito_logging.rst
```diff
@@ -123,20 +151,20 @@ the ``yui.config`` object.
...
    [
      {
-       "settings": [ "runtime:client" ],
+       "settings": [ "master" ],
        ...
-	   "yui": {
+	     "yui": {
          "config": {
-           "logLevel": "warn"
+           "logLevel": "info"
```
caridy, line 158: Remember to use "debug": true, otherwise all this is irrelevant.

examples/developer-guide/simple_logging/mojits/log/controller.server.js
```diff
@@ -1,56 +1,12 @@
...
-                    html : "<h3 style='color:red;'>simple html</h3>"
-                };
-            Y.log('[CONTROLLER]: Today ' + today);
+YUI.add('log', function(Y, NAME) {
+    Y.namespace('mojito.controllers')[NAME] = {   
+        index: function(ac) {
+            Y.log('[CONTROLLER]: Default log-level message with date: ' + new Date());
+            Y.log('[CONTROLLER]: Warn message.','warn');
+            var data = {
+                log_config: ac.config.getAppConfig().yui.config.logLevel,
```
caridy, line 7: I don't think this is a good example, because getAppConfig is bound to the current context, instead of the static configuration from master. And in fact, the log level setting was already set. I think it is better to stick to Y.config.logLevel as a way to know the level at any given time at the server side.

caridy:
@zhouyaoji other than that, I think it is imperative that we add some documentation about the important of the 3rd parameter, which is normally NAME from the YUI.add statement, this parameter should always be used because that's what allow you to use logInclude and logExclude. All our examples, and demos should reflect that.

PR #808, opened 2012-11-28
Y.config.debug is true by default per YUI default settings

caridy:
- we should honor that. 
- Making sure that default log level is computed correctly based on the initial definition of logLevelOrder
- moving that default definition to the configuration config.json

### DIFF
--- a/lib/mojito.js
+++ b/lib/mojito.js
@@ -214,7 +214,7 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
         useSync: true
     });
 
-    this._configureLogger(Y, yuiConfig);
+    this._configureLogger(Y);
     this._configureYUI(Y, store, modules);
 
     // attaching all modules available for this application for the server side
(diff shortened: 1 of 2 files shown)


lib/mojito.js
```diff
@@ -342,9 +342,10 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
  * Configures YUI logger to honor the logLevel and logLevelOrder
  * TODO: this should be done at the low level in YUI.
  */
-MojitoServer.prototype._configureLogger = function(Y, yuiConfig) {
-    var logLevel = (yuiConfig.logLevel || 'debug').toLowerCase(),
-        logLevelOrder = ['debug', 'mojito', 'info', 'warn', 'error', 'none'];
+MojitoServer.prototype._configureLogger = function(Y) {
+    var logLevel = (Y.config.logLevel || 'debug').toLowerCase(),
+        logLevelOrder = Y.config.logLevelOrder || [],
+        defaultLogLevel = logLevelOrder[0] || 'info';
```
drewfish, line 348: Won't `logLevelOrder[0]` now always be `debug`, since we're always using config.json?

caridy, line 348: @drewfish No, because Y.config.logLevelOrder comes from getAppConfig() from store in the first place, it is set as part of the Y.config early on, and it lands in the _configureLogger through Y.config.logLevelOrder. So, developers my want to redefine that list, in which case the lower priority element does not need to be 'debug' necessarily.

drewfish:
+1

PR #819, opened 2012-12-03
update test as the app is updated

lzhan:


drewfish:
+1

lzhan:
Travis build on node06 is done. All func tests passed. Travis build on node08 seems very slow. Merge now.

PR #827, opened 2012-12-04
skip synthetic yui modules

drewfish:


isao:
+1

caridy:
This sounds like a temporary solution, eventually we will need to have that meta in the synthetic files as well.

PR #828, opened 2012-12-04
archtype updates

isao:
- enable -p option for `create mojit`
- update app/hybrid mojito controller to 0.5.x

### DIFF
--- a/lib/app/archetypes/app/hybrid/mojits/top_frame/controller.server.js.hb
+++ b/lib/app/archetypes/app/hybrid/mojits/top_frame/controller.server.js.hb
@@ -2,6 +2,8 @@
  * Copyright (c) 2012 Yahoo! Inc. All rights reserved.
  */
 /*jslint anon:true, sloppy:true, nomen:true*/
+/*globals YUI*/
+
 YUI.add('top_frame', function(Y, NAME) {
 
 /**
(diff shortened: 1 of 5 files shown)


lib/app/archetypes/app/hybrid/mojits/top_frame/controller.server.js.hb
```diff
@@ -43,4 +42,9 @@ YUI.add('top_frame', function(Y, NAME) {
 
     };
 
-}, '0.0.1', {requires: ['mojito', 'top_frameModelFoo']});
+}, '0.0.1', {requires: [
+    'mojito',
+    'mojito-assets-addon',
+    'mojito-models-addon',
+    'top_frameModelFoo'
```
mojit0, line 49: I think this has to change to something ala '{{name}}ModelFoo'.

mojit0:
One comment, otherwise +1

drewfish:
+1

lib/app/archetypes/app/hybrid/mojits/top_frame/controller.server.js.hb
drewfish, line 49: This is an app archetype, which contains a canned "top_frame" mojit, so I think what Isao has is correct.

isao, line 49: it's inconsistent, I know. the hybrid archetype is supposed to include both an app and a mojit and the mojit name was hard-coded (was a requirement)

PR #831, opened 2012-12-04
fix mojito create -p option

isao:
- add port: "{{port}}" to application.json
- rename with .hb extension

### DIFF
--- a/lib/app/archetypes/app/default/application.json.hb
+++ b/lib/app/archetypes/app/default/application.json.hb
@@ -1,6 +1,7 @@
 [
     {
         "settings": [ "master" ],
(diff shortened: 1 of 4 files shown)


caridy:
@isao we are transforming the port from a numeric value to a string, for some reason @drewfish change it to a numeric value, can you comment?

isao:
it was a trade-off to allow the raw json.hb file to be linted and syntax highlighted

isao:
note, this is not intended for pr6

drewfish:
I changed the port numbers in `tests/func/applications/apps.json` which is a bit different.

caridy:
+1 then

PR #836, opened 2012-12-05
more careful about when we say a resource is modified

drewfish:


caridy:
+1

add0n:
I can confirm that this fixed the problem I was seeing - thanks Drew and Caridy!

PR #839, opened 2012-12-06
Support for global models

caridy:
- a parent mojit can register a custom global model instance
- a child mojit can ask for a regular model and receive a global model if exists


isao:
We've been planning to deprecate "app-level" mojits, which are really a kind of convoluted global model containers AFAIK. We should proceed to officially deprecate app-level if we go with this.

This is an improvement, but IMO it'd be preferred to have a parent mojit (or the thing that owns page-scope) inject/specify shared resources and data.

Top-down injection would simplify control and optimization with respect to order of execution. Whereas the global model would have to deal with invocations bottom-up. I'm thinking of the case of mojit A request/context -> webservice I, ws II.. etc -> modelled data, and mojit B -> ws I, ws III -> modelled data

lib/app/addons/ac/models.common.js
```diff
@@ -13,6 +14,8 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
 
     'use strict';
 
+    var _clientCache = {};
```
drewfish, line 21: A short comment would by nice, something like "global model cache when on the client".

tests/unit/lib/app/addons/ac/test-models.common.js
```diff
@@ -0,0 +1,155 @@
...
+            var globalModel = {
+                name: 'global'
+            };
+
+            Y.mojito.models.cuba = localModel;
+            addon.registerGlobal('cuba', globalModel);
+
+            model = addon.get('cuba');
+            A.isObject(model, 'registered model should return an instance');
+            A.areSame(globalModel, model, 'global registered should have priority over local models');
```
drewfish, line 148: Really?  I think that might be controversial.  As an app developer I might want my local model to override my global model.

Looking at the implementation of `get()` I think the following will result in the local model being returned from both calls:

``` javascript
Y.mojito.models.cuba = localModel;
addon.get('cuba');  // returns localModel
addon.registerGlobal('cuba', globalModel);
addon.get('cuba');  // returns localModel
```

drewfish:
"app-level" mojits are really something different.  An "app-level" mojit isn't really a mojit, it's just a collection of resource shared with all other mojits.

The YAF convergence design has a feature in it where the user can define app-level **models**.  This is especially useful when running client side (on the page) where you might only want one "user" model for example.

drewfish:
Yeah, a lot of this would be improved with the data scopes we've been wanting.  I'm OK with most of this PR until we have those, though.

caridy:
@drewfish is right, the use case I'm trying to solve here is:

As a dev, I want to create an instance of a model and share it with other mojits in the page.

These global models are really page models since they should be part of the page data scope that is a long living structure, but we are not there yet. In fact, the whole re-hydration process has to be implemented as well, but this is the first step in that direction.

isao:
Yes, I understand, and again, it's simple enough and better than what exists. I'm just wary of opening the door to globals and treating apps like web pages.

Instead of register/request globally, can this be scoped to ancestor mojit?

Anyway, +1 with these caveats

drewfish:
+1

caridy:
@isao as today, mojits are doing PULL to get a model, which is just fine. The problem is when we want to control the mojit from outside, and use some kind of PUSH into the mojit to get control over the model that the mojit will ended up using. For that, parent-child is not enough, because in many cases, we want to define the data structure at the page/app level, and that's what this PR resolves. Eventually we will have configurations and stuff that can let the mojito framework to do more, but for now, a mojit is responsible for registering it.

rwaldura:
Guys:
I think I understand what we're trying to accomplish here, but can you make sure you document this thoroughly?
I.e. can you draft something for @zhouyaoji , or patch the existing doc? 
http://developer.yahoo.com/cocktails/mojito/docs/intro/mojito_mvc.html#models
which maps to
https://github.com/yahoo/mojito/blob/develop/docs/dev_guide/intro/mojito_mvc.rst
Thanks!

caridy:
@rwaldura I'm not so sure we are ready for primetime for this new feature. Let's call it experimental for now!

PR #840, opened 2012-12-06
mojito start output version of mojito being run

lzhan:
Fix trello card 219.

### DIFF
--- a/lib/app/commands/start.js
+++ b/lib/app/commands/start.js
@@ -10,7 +10,7 @@
 var path = require('path'),
     utils = require(path.join(__dirname, '../../management/utils')),
     Mojito = require(path.join(__dirname, '../../mojito')),
-    mojitoVersion = require('../../../package.json').version,
+    mojitoVersion = require(path.join(__dirname, '../../../package.json')).version,
     fs = require('fs');
 
 
--- a/lib/app/commands/start.js
+++ b/lib/app/commands/start.js
@@ -10,7 +10,7 @@
 var path = require('path'),
     utils = require(path.join(__dirname, '../../management/utils')),
     Mojito = require(path.join(__dirname, '../../mojito')),
-    mojitoVersion = require(path.join('../../../package.json')).version,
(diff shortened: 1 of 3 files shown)


mojit0:
+1

lzhan:
I see. Thanks Isao. I will update.

caridy:
@lzhan can we also remove the internal static reference to version from mojito.common.js in autoload?

```
YUI.add('mojito', function(Y, NAME) {

    Y.namespace('mojito').version = '0.2';
```

lzhan:
Yes, I will remove it.

drewfish:
Beware that `lib/app/mojits/HTMLFrameMojit/controller.server.js` uses `Y.mojito.version`.  I know that `lib/app/autoload/store.server.js` has the mojito version, so perhaps we can expose that somehow.  Also I think the RS isn't quite getting the mojito version correctly, so that needs to be updated.

Hmm... perhaps we could have the RS populate `Y.mojito.version`.

lzhan:
controller.server.js does not get right mojito version: 'Powered by Mojito ' + Y.mojito.version;  Y.mojito.version gives undefined.

drewfish:
Yes exactly, since you removed Y.mojito.version :)  I would be OK with changing HTMLFrameMojit's controller to just say "Powered by Mojito".

lzhan:
I see. I was wondering if it is used anywhere :). I will check if RS can populate Y.mojito.version.

lib/app/commands/start.js
```diff
@@ -10,6 +10,7 @@
 var path = require('path'),
     utils = require(path.join(__dirname, '../../management/utils')),
     Mojito = require(path.join(__dirname, '../../mojito')),
+    mojitoVersion = require('../../../package.json').version,
```
drewfish: I think you want to do the `path.join(__dirname, ...)` here too.

drewfish:
+1

PR #841, opened 2012-12-06
Micro optimization - Second Attempt

caridy:
The intent is to do a quick review across the core modules to try to optimize them for V8, specially the low hanging fruits. So far, we have two things we want to optimize:
- "delete" is evil in V8, avoid the use of delete on objects.
  for more details, check this gist:
  https://gist.github.com/4126067
- "try/catch" is slow, avoid the use of it when possible.

PR #778

### DIFF
--- a/lib/app/addons/ac/composite.common.js
+++ b/lib/app/addons/ac/composite.common.js
@@ -205,6 +205,8 @@ callback({
                 buffer = {},
                 content = {},
                 meta = {},
+                sanitizedChildren = {},
+                name,
                 perf;
 
             cfg.children = cfg.children || {};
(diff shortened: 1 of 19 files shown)


drewfish:
Just to be pedantic, your gist suggests that `delete` isn't the problem, but `if (obj.member_doesnt_exist)` is the problem.

caridy:
@drewfish the gist is just a simple demostration, I do have a test based on benchmark.js that reflects the same though.

Now, about the `delete` vs `if`, the problem here is the V8 low level optimizations. The `if` is in every test, but only on the object that was transformed using `delete` the `if` will take longer. Apparently, by using `delete`, V8 will change the type of object in memory, which affects the members resolution.

drewfish:
+1

isao:
@caridy actually `delete` removes the key, so javascript must check the prototype chain. by assigning something falsey as a value of the key, the falsey check is immediate. I wouldn't say `delete` is evil, but +1 :)

PR #842, opened 2012-12-06
Support for handlebar partials

caridy:
## Example:

```
./MojitName/views/
    foo.hb.html
./MojitName/views/partials/
    bar.hb.html
```

by following that structure, you should be able to define `foo.hb.html` like this:

```
<div id="{{mojit_view_id}}">
    {{> bar}}
</div>
```

Few more notes:
- a partial view can be used by any view within the same context/mojit.
- global partial views are also supported.
## Details:
- adding support for partials in the store. there is now `instance.partials` with a similar structure than `instance.views` after expanding the instance. 
- views within `partials/*` folder are now consider partials, which are an special type of view.
- adding support for `application.json->viewEngine->preloadTemplates` to preload any view in memory, and this is disabled by default.
- consolidating `application.json->viewEngine->cacheTemplates` as a static config across the board.
- adding support for preloadTemplates for HB client and server to avoid loading the template if is is part of the expanded instance.
- adding support for partials on hb client and server.
- `view-renderer` is now a class and a factory to avoid creating renderer objects over and over again.
- removing viewId argument that is not really relevant for this abstraction.
- using the new `Y.mojito.ViewRenderer()` factory to access the renderer instance.
- removing unnecesary dependencies (`mojito-perf`).
- removing unnecessary configuration that is not semantically correct for `action-context` or `mojito-client` like `cacheTemplates` and `pathToRoot`.

Second attempt for PR #683


caridy:
This is set to land on 0.5.2.

caridy:
Ready for review.

drewfish:
From reading the description:

<del>I worry about getting rid of `pathToRoot`.  It was necessary for some users of `mojito build html5app`.</del> Evidence strongly suggests that this was no longer being used (dead code) anyways.

lib/app/addons/view-engines/hb.client.js
```diff
@@ -16,25 +16,30 @@ YUI.add('mojito-hb', function(Y, NAME) {
...
+    }
 
     HandleBarsAdapter.prototype = {
 
         /**
          * Renders the handlebars template using the data provided.
          * @param {object} data The data to render.
-         * @param {string} mojitType The name of the mojit type.
-         * @param {string} tmpl The name of the template to render.
+         * @param {object} instance The expanded mojit instance.
```
drewfish, line 34: Passing the whole instance, nice.

lib/app/autoload/store.server.js
```diff
@@ -695,16 +697,23 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+                                    this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url) :
+                                    res.source.fs.fullPath),
+                            'content': res.content,
+                            'engine': res.view.engine
+                        };
+                        // we want to separate partials from actual templates
+                        // in case the engine supports partials
+                        if (res.name.indexOf('partials/') === 0) {
+                            // removing the "partials/" prefix
+                            details.partials[this._libs.path.basename(res.name)] = template;
```
drewfish, line 713: This looks like it won't support subdirectories in `views/partials/`, which might be nice to have.

drewfish:
+1 good stuff.

One comment about supporting subdirectories in `views/partials/`, but we can add that feature in the future when/if requested.

caridy:
Thanks @drewfish, I will resolve the conflicts and merge.

PR #854, opened 2012-12-11
fix for ac.url.make weird behavior after micro-optmiziation

caridy:
- ac.url.make issues
- func tests failures after the first iteration of the optimization


lib/app/addons/ac/url.common.js
```diff
@@ -65,35 +43,40 @@ YUI.add('mojito-url-addon', function(Y, NAME) {
...
-
-            url = this.getRouteMaker().make(query, verb);
+            params = Y.merge(routeParams || {});
 
             if (urlParams) {
-                urlParams = objectToQueryStr(urlParams, true);
-
-                if (urlParams && urlParams.length) {
-                    url = url + '?' + urlParams;
+                if (Y.Lang.isObject(urlParams)) {
```
drewfish: I think `Y.Lang.isObject()` returns true for arrays too, although it probably doesn't matter too much for this situation.

lib/app/addons/ac/url.common.js
```diff
@@ -65,35 +43,40 @@ YUI.add('mojito-url-addon', function(Y, NAME) {
...
+                    // adding querystring params to routeParams and let
+                    // the url maker to create the proper url. Empty params
+                    // will be left out. TODO: why?
+                    for (key in urlParams) {
+                        if (urlParams.hasOwnProperty(key) && urlParams[key]) {
+                            params[key] = urlParams[key];
+                        }
+                    }
+                } else {
+                    Y.log('ac.url.make is expecting an object as urlParams instead of: ' + urlParams, 'warn', NAME);
```
drewfish: Small point:  perhaps Y.JSON.stringify(urlParams) just to be safe.

drewfish:
A couple of small comments.  +1 even without those addressed.

caridy:
@drewfish, can you do a second pass on this one?

lib/app/addons/ac/url.common.js
```diff
@@ -61,39 +39,46 @@ YUI.add('mojito-url-addon', function(Y, NAME) {
...
-                urlParams = objectToQueryStr(urlParams, true);
-
-                if (urlParams && urlParams.length) {
-                    url = url + '?' + urlParams;
+                urlParams = (typeof urlParams === 'string' ? Y.QueryString.parse(urlParams) :
+                        urlParams);
+
+                // adding querystring params to routeParams and let
+                // the url maker to create the proper url. Empty params
+                // will be left out. TODO: why?
```
drewfish, line 64: I think empty params are left out to "optimize" the URL -- so that it's as short as possible.

drewfish:
+1

PR #860, opened 2012-12-13
better handling of null children

drewfish:


caridy:
+1

isao:
need to add /_jslint continue:true_/ at top or nr the statement(s)

PR #863, opened 2012-12-14
micro-optimization: replacing runInNewContext with runInContext

caridy:
runInContext seems to be 200x faster when using a static context, and this is suppose to help with apps with many files, improving the start up time during "mojito start", and apps with many tests during "mojito test" command.

According to travis, the improvement is not considerable, few minutes compared to the previous commit.


drewfish:
+1

PR #869, opened 2012-12-28
Windows support, replaces PR #460

jenny:
This PR takes Mike's original PR #460 and updates it for the 0.5 codeline. Mojito starts ok but there is still a problem with the combo static handler, visible in the newsboxes app.


caridy:
Closing after merging #1103

PR #872, opened 2013-01-03
do not colorize test.js output if there is no tty [gh809,bz5964521]

isao:
- use process.*.isTTY instead of require('tty').istty()
- tested on node 0.6, 0.8
- n.b. some app and addon Y.log() messages remain tty unaware.

### DIFF
--- a/lib/app/commands/test.js
+++ b/lib/app/commands/test.js
@@ -17,7 +17,6 @@ var libpath = require('path'),
     targetMojitoPath = BASE,
     mojitoTmp = '/tmp/mojitotmp',
     mojitoInstrumentedDir = '/tmp/mojito-lib-inst',
-    fwTestsRoot = BASE + 'lib/tests',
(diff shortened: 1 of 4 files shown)


lib/app/commands/test.js
```diff
@@ -4,7 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
-/*jslint anon:true, regexp:true, nomen:true, stupid:true*/
+/*jslint anon:true, regexp:true, nomen:true, stupid:true node:true*/
```
mojit0: comma after stupid:true ?

mojit0:
+1

lib/app/commands/test.js
isao: interesting looks like the jslint I have and jslint.com don't care about the comma in the config. I'll add it for consistency

PR #873, opened 2013-01-04
Functional test changes to support SD

lzhan:
1. functional app will launched one by one after tests for each app are executed.
2. callback function is used instead of timeout after each app is started.
3. port number is not fixed and can be specified from commandline.
4. tests for different apps are put in separated test descriptor.
5. html5app is hosted/started by static command.


mojit0:
+1

drewfish:
How come some of the descriptor files have config.application.type but others don't?  (I see two times `mojito` and `static`, but only common_descriptor.json has `"type":"mojito"`.

lzhan:
The default type is "mojito" if it is not specified.

drewfish:
+1

PR #878, opened 2013-01-04
[bz6023234] prefix for hybridapp/crt "//yahoo.com"

isao:
bz 6023234 - (preferred solution is to restore relative pathing in hybridapp builds)


caridy:
+1

PR #883, opened 2013-01-07
Travis-related cleanup

isao:
I'm pretty sure we don't need travis/script.sh anymore. Related cleanup.

Let's see if Travis agrees.


isao:
hm, travis not seeing the updates, will redo pr

PR #887, opened 2013-01-08
more unit tests for ActionContext

drewfish:


lib/app/autoload/action-context.common.js
```diff
@@ -607,6 +607,10 @@ YUI.add('mojito-action-context', function(Y, NAME) {
                 clearTimeout(this._timer);
                 this._timer = null;
             }
+            // in case someone calls ac.done() after calling ac.error()
+            this.done = function() {
```
caridy, line 611: nice catch!

caridy:
+1

PR #889, opened 2013-01-08
Add more unit tests for params.common and view engine handlebar

lzhan:


lib/app/addons/ac/params.common.js
```diff
@@ -204,10 +204,10 @@ YUI.add('mojito-params-addon', function(Y, NAME) {
          * @return {string|object} param value, or all params if no key
          *     specified.
          */
-        getFromFiles: function() {
+         /*getFromFiles: function() {
```
drewfish: Instead of commenting this out, why not just delete it?

If it should be kept, there should be a comment saying -why- it should be kept.

lib/app/addons/ac/params.common.js
```diff
@@ -217,9 +217,9 @@ YUI.add('mojito-params-addon', function(Y, NAME) {
          * @return {string|object} param value, or all params if no key
          *     specified.
          */
-        files: function() {
+        /*files: function() {
```
drewfish: Instead of commenting this out, why not just delete it?

If it should be kept, there should be a comment saying -why- it should be kept.

lzhan:
I am not sure what is our plan about those two APIs. Will remove it for now. We can add them back when there is a customer request.

drewfish:
+1

PR #897, opened 2013-01-10
.travis.yml debug, misc

isao:
- show selenium stderr
- no sleep (jslint and unit tests run first)
- use default `npm test` instead of script config


caridy:
+1

PR #898, opened 2013-01-10
[gh238] [bz5667423] application.json files in NPM packages can contribute as well

drewfish:


lib/app/autoload/store.server.js
```diff
@@ -223,7 +223,15 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._validDims = this._parseValidDims(this.config.getDimensions());
             this.validateContext(this._config.context);
             this._fwConfig = this.config.readConfigSimple(this._libs.path.join(this._config.mojitoRoot, 'config.json'));
+
+            // We use this during preload() before we even know about all the
+            // application.jsons.  This might be OK, since we'll have the real
+            // values during the second pass.
+            // In practice, this only applies to "mojitDirs" and "mojitsDirs",
+            // which can only be given in the app-level application.json.
```
caridy, line 231: This is interesting. If we can specify the list of folders for mojits, we probably can do something similar to specify the list of folders to pick up other application.json files as well rather than doing the regular walk (in case the app owner wants to control this process).

lib/app/autoload/store.server.js
```diff
@@ -349,13 +357,20 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
 
+            ctx = this.blendStaticContext(ctx);
             key = JSON.stringify(ctx || {});
 
             if (this._appConfigCache[key]) {
                 return JSON.parse(this._appConfigCache[key]);
             }
 
-            ycb = this.config.readConfigYCB(this._libs.path.join(this._config.root, 'application.json'), ctx);
+            if (this._appConfigYCB) {
```
caridy, line 367: This is very confusing. Having `ycb` to hold the app level application.json then to hold the merged version. I wonder if we should use a separate structure.

lib/app/autoload/store.server.js
```diff
@@ -374,10 +402,19 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
         preload: function() {
             // We need to do an initial sweep to find the resource store addons.
             this.preloadResourceVersions();
+
+            // Now that we've found all the application.jsons, we can aggregate them.
+            this._initAppConfig();
```
caridy, line 407: Does this means we need to call store.preload() before getting any app.config value?

lib/app/autoload/store.server.js
drewfish, line 231: We could do that, but the goal of this feature is to automatically find and merge the application.jsons in the NPM packages.

lib/app/autoload/store.server.js
drewfish, line 367: Yeah, this approach seemed a little too clever for me as well.  I will rework.

lib/app/autoload/store.server.js
drewfish, line 407: Yes.  That is because the NPM-package-walk during preload needs to happen so that the other application.jsons can be detected.

I'm strongly considering making a `preloadFast()` or something like that, which is a much lighter version of preload(), so that can be used instead when a full preload() is undesired.

drewfish:
Closing this, as we're going to take a different approach, of adding `import_settings` support in the YCB files.

PR #905, opened 2013-01-10
[gh238] [bz5667423] support for multiple application configuration files

caridy:
Continuation of PR #898.

By specifying the list of files to aggregate with the application.json, developers will have the flexibility to split the application configuration into various files, e.g:

```
[
    {
        "settings": [ "master" ],
        "applicationConfigFiles": [
            "node_modules/devices/application.json",
            "relative/path/to/custom.json"
        ],
        "testKey1": "testVal1",
        "testKey2": "testVal2",
        "testKey3": "testVal3"
    }
]
```


lib/app/addons/rs/config.js
```diff
@@ -165,6 +176,48 @@ YUI.add('addon-rs-config', function(Y, NAME) {
...
+                path,
+                config,
+                s,
+                section,
+                settings = {},
+                bundle = [];
+            bundle.push(this.getDimensions()[0]);
+            for (p = 0; p < paths.length; p += 1) {
+                path = paths[p];
+                config = this.readConfigSync(path);
```
drewfish, line 197: This should use readConfigSimple() instead.  The work to wire in YAML will update readConfigSimple() to also understand YAML files.

drewfish:
Does this support application.json which includes application2.json which includes application3.json?

caridy:
@drewfish No, that's not supported. It developers want to keep control over what they have/use, there is not need for that. Also the error tolerance is important, since application.json is optional.

lib/app/addons/rs/config.js
caridy, line 197: At the moment, `readConfigSimple` does not provide that capability, and the multi-extension capability as well. Once that land it, we can change it I guess.

drewfish, line 197: That's part of what I'm saying.  Right now in develop, we should not be trying to support YAML.  So, you should use readConfigSimple().  The work to support YAML will do the needful, whether that's update readConfigSimple() to readConfigSync() or not.

(I am working on the YAML support.  readConfigSync() is going away, so calling it will cause a merge conflict of sorts. It's better to use the established API (readConfigSimple) than to use readConfigSync() which isn't used by anyone else and which is going away.)

Said another way, this pull request should -only- be focused on implementing multiple application.jsons, and not also trying to partially support YAML.

caridy:
@drewfish I will merge this so we can work on the PR #907 to accomodate the rest of the changes.

PR #907, opened 2013-01-11
re-enable yaml support

drewfish:
closes #804

### DIFF
--- a/lib/app/autoload/store.server.js
+++ b/lib/app/autoload/store.server.js
@@ -364,6 +364,20 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
 
             return appConfig;
         },
+
+
+        /**
+         * Does initial preload of many parts of the application and framework.
+         * The full preload is done by preload().
+         *
+         * @method preloadInitial
+         * @return {nothing}
+         */
+        preloadInitial: function() {
+            this.preloadResourceVersions();
+        },
+
+
         /**
          * Preloads everything in the app, and as well pertinent parts of
          * the framework.
@@ -373,11 +387,14 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          */
         preload: function() {
             // We need to do an initial sweep to find the resource store addons.
-            this.preloadResourceVersions();
+            this.preloadInitial();
+
             // And then use them.
             this.loadAddons();
+
             // Then, do another sweep so that the loaded addons can be used.
             this.preloadResourceVersions();
+
             this.makeResourceVersions();
             this.resolveResourceVersions();
         },
--- a/lib/app/autoload/store.server.js
+++ b/lib/app/autoload/store.server.js
@@ -1096,7 +1096,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                     parents: [],
                     dir: dir
                 };
-                info.pkg = this.config.readConfigSimple(this._libs.path.join(dir, 'package.json'));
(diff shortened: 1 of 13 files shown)


lib/app/autoload/store.server.js
```diff
@@ -364,6 +364,20 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+
+
+        /**
+         * Does initial preload of many parts of the application and framework.
+         * The full preload is done by preload().
+         *
+         * @method preloadInitial
+         * @return {nothing}
+         */
+        preloadInitial: function() {
```
caridy, line 376: What happen if preloadInitial and then preload is called? Will be good if we support that use-case, so we can do some stuff before getting to the .preload() execution.

lib/app/autoload/store.server.js
```diff
@@ -373,11 +387,14 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          */
         preload: function() {
             // We need to do an initial sweep to find the resource store addons.
-            this.preloadResourceVersions();
+            this.preloadInitial();
```
caridy, line 390: related to the comment above.

lib/app/commands/start.js
```diff
@@ -55,31 +56,29 @@ exports.options = [
...
+    store = Store.createStore({
+        root: root,
+        preload: 'initial',
+        context: inputOptions.context || {}
+    });
+    appConfig = store.getAppConfig();
+        
+    pack = store.config.readConfigJSON(path.join(root, 'package.json'));
 
     options.port = params[0] || appConfig.appPort || process.env.PORT || 8666;
```
caridy, line 79: Finally, different port per different context :) nice

lib/store.js
```diff
@@ -79,7 +83,14 @@ Store.createStore = function(options) {
         appConfig: options.appConfig
     });
 
-    store.preload();
+    switch (options.preload) {
```
caridy, line 86: isn't switch one of the bad parts in javascript? A couple of "if" statements should be just enough considering that the value was already validated/preset.

caridy, line 86: Travis failure:

```
458+ /home/travis/builds/yahoo/mojito/bin/mojito jslint -p
459store.js
460   1 87,9: Expected 'case' at column 5, not column 9.
461     case 'skip': break;
462   2 87,22: Expected 'break' at column 9, not column 22.
463     case 'skip': break;
464   3 88,9: Expected 'case' at column 5, not column 9.
465     case 'initial':
466   4 89,13: Expected 'store' at column 9, not column 13.
467     store.preloadInitial();
468   5 90,13: Expected 'break' at column 9, not column 13.
469     break;
470   6 91,9: Expected 'default' at column 5, not column 9.
471     default:
472   7 92,13: Expected 'store' at column 9, not column 13.
473     store.preload();
474app/commands/start.js
475   1 76,1: Unexpected '(space)'.
```

lib/app/autoload/store.server.js
drewfish, line 376: Yeah, I had thought of that too.  I will implement.

lib/store.js
drewfish, line 86: I had no idea switch was contraindicated.  The errors are because I forgot to run jslint, which I'll do.

caridy:
Closing this in favor of issue #916 which addressed the comments here and more to accomodate this commit for the multi-part application config files as well.

PR #910, opened 2013-01-11
unformatted logs when running in production systems

caridy:
when running on production, we should rely on console.log directly instead of std.err used by yui-log-nodejs.

fixes #bz5908204


isao:
+1

PR #916, opened 2013-01-11
re-enable yaml support

caridy:
- continuation of issue #907 from @drewfish
- closes #804
- appConfig can be used as part of mojito start
- comments are now supported

### DIFF
--- a/lib/store.js
+++ b/lib/store.js
@@ -32,7 +32,9 @@ var Store = {};
  * @param {string} options.string Path to application directory.
  * @param {object} options.context Runtime context to apply to all requests.
  * @param {object} options.appConfig Overrides for the application.json file.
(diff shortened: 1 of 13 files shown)


lib/store.js
```diff
@@ -79,7 +83,14 @@ Store.createStore = function(options) {
         appConfig: options.appConfig
     });
 
-    store.preload();
+    if ('initial' === options.preload) {
+        store.preloadInitial();
+    } else if ('skip' !== options.preload) {
+        // preload by default when the store is created, but
+        // if you only care about appConfig and package.json
+        // you probably should use 'skip' option.
```
drewfish: This comment should be part of the method docs, not buried in the code.

caridy: I will do that.

drewfish:
+1

PR #917, opened 2013-01-12
add more mojito.js tests

isao:


lib/mojito.js
```diff
@@ -149,6 +149,84 @@ MojitoServer.prototype._startupTime = null;
 //  Private Methods
 //  ---------------
 
+MojitoServer.prototype._makeMwList = function (app_mw, mojito_mw) {
```
caridy: There is not need to shorten the names when it comes to nodejs. Let's use the full "Middleware" instead of "Mw".

caridy: Let's also add the documentation for these two new methods.

lib/mojito.js
```diff
@@ -282,13 +331,7 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
...
         });
         if (appConfig.debugMemory) {
-            outputHandler.setMemoryDebugObjects({
-                'YUI.Env': YUI.Env,
-                'YUI.config': YUI.config,
-                'YUI.process': YUI.process,
-                Y: Y,
-                store: store
-            });
+            outputHandler.setMemoryDebugObjects(debugConfig);
```
caridy: We can eliminate the setMemoryDebugObject, we don't need that anymore.

lib/mojito.js
```diff
@@ -151,6 +151,101 @@ MojitoServer.prototype._startupTime = null;
...
+            // backwards-compatibility: user-provided middleware is
+            // specified by path
+            midPath = libpath.join(midDir, midName);
+            //console.log("======== MIDDLEWARE user " + midPath);
+            midBase = libpath.basename(midPath);
+            if (0 === midBase.indexOf('mojito-')) {
+                // Same as above (case of Mojito's special middlewares)
+                // Gives a user-provided middleware access to the YUI
+                // instance, resource store, logger, context, etc.
+                //                                                ...dead code?
```
drewfish: This isn't dead code, this is in fact the new API.  The "else" branch is actually the old code/API which we'll deprecate at some point.

tests/unit/lib/test-mojito.js
```diff
@@ -250,5 +261,333 @@ YUI().use('mojito', 'test', function (Y) {
...
+
+        'test listen 1': function () {
+            var port = 1234,
+                host = 'letterman';
+
+            this._startupTime = +new Date();
+            this._options = {verbose: true};
+
+            Mojito.Server.prototype.listen.call(this, port, host);
+        },
```
drewfish, line 332: Where are the asserts in this test?

tests/unit/lib/test-mojito.js
```diff
@@ -250,5 +261,333 @@ YUI().use('mojito', 'test', function (Y) {
...
+            Y.Mock.expect(app, {
+                method: 'listen',
+                arguments: [port, host, cb]
+            });
+
+            this._startupTime = null;
+            this._options = {verbose: false};
+            this._app = app;
+
+            Mojito.Server.prototype.listen.call(this, port, host, cb);
```
drewfish: This approach, of passing "this" to the function call, is weird.  Reading it, I don't understand what kind of object "this" is, or how it relates to the tests we're attempting.  If you'd created a mock object and passed it as the first argument to call() that'd make more sense to me.

isao: agreed, that's much cleaner. I'll update, thx

lzhan:
+1

PR #922, opened 2013-01-14
[bz5800733] Handlebars Template Helpers

twobit:
Now that Handlebars is the default view engine, it would be great to expose more functionality. Template helpers can simplify common patterns in templates. I added `HandleBarsAdapter.register` in `hb.server.js` and a test. Since this is only applicable to Handlebars, user must require 'mojito-hb' and register like so:

```
Y.namespace('mojito.addons.viewEngines.hb').register('helper', function (options) {
    return 'CustomHelper';
});
```

You would use it like so within your template:

```
{{helper}} or {{#helper}}{{/helper}}
```

For now, server side only. See http://handlebarsjs.com/block_helpers.html for more info.

### DIFF
--- a/lib/app/addons/view-engines/hb.server.js
+++ b/lib/app/addons/view-engines/hb.server.js
@@ -26,6 +26,16 @@ YUI.add('mojito-hb', function(Y, NAME) {
         this.viewId = viewId;
     }
 
+    /**
+     * Registers a Handlebars template helper method.
+     * See http://handlebarsjs.com/block_helpers.html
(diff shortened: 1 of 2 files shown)


caridy:
The problem here is the isolation. The helpers registered are global, and that might be very confusing in complex apps. The approach that we are exploring is to have a level of isolation by default at the mojit level for partials and helpers. I will keep you posted. Let's keep this open in the meantime.

caridy:
Helpers will be ready in 0.5.5 later today!

PR #925, opened 2013-01-15
add unit test for command info

lzhan:


tests/cli/lib/app/commands/test-info1.js
```diff
@@ -0,0 +1,55 @@
...
+        libpath = require('path'),
+        cmdpath = libpath.join(__dirname, '../../../../../lib/app/commands/info.js'),
+
+        mockery = require('mockery'),
+        mockPath,
+        info;
+
+    suite.add(new Y.Test.Case({
+
+        name: 'info test cases',
```
drewfish: This suite has the same name as the suite found in tests/cli/lib/app/commands/test-info.js.  I don't know if that'll cause a problem or not.

tests/cli/lib/app/commands/test-info1.js
```diff
@@ -0,0 +1,55 @@
...
+        setUp: function() {
+        },
+
+        tearDown: function() {
+            mockery.deregisterAll();
+            mockery.disable({
+                'useCleanCache': true,
+                'warnOnUnregistered': false,
+                'warnOnReplace': false
+            });
```
drewfish, line 47: It's a little weird having cleanup in tearDown() but the initialization isn't being done in setUp().  I suggest moving those mockery.whatever() calls from the test itself into setUp().

lzhan:
Thanks Drew. Will update as you suggested.

drewfish:
+1

PR #926, opened 2013-01-15
moving part of the logic from HTMLFrame to mojito-frame-addon

caridy:
The idea here is to make the creation of Frames Mojits a breeze.

Based on the current implementation we can create a mojit that inherit from HTMLFrameMojit or just create a brand new that uses a set of addons to create a frame. In this gist I demonstrated both approaches:
- https://gist.github.com/4543247

Todo:
- Document re-usability of the HTMLFrameMojit methods (this is for Joe)


drewfish:
Perhaps instead move the asset management stuff (renderListAsHtmlAssets() etc) into the "assets" AC addon.  That way you (perhaps) don't need a new AC addon.

caridy:
Tests are still unstable, one of jobs passed, the other fail here:

Failed Common: simplemodelclient onMozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:18.0) Gecko/20100101 Firefox/18.0
Passed, 1 Failed , 0 skipped 
in testCase_yui_3_7_3_2_1358546278029_33
test simplemodelclient:Unexpected error: Y.one(...) is null

I did a manual verification, and it works just fine. Seems like a timer issue in Arrow.

Anyway, this is ready for review!

isao:
+1

PR #935, opened 2013-01-17
Add support for a shared app levellang bundle

gomezd:
This change is to allow using a "shared app level" lang bundle for getting localized strings available for the whole application.

The yui lang module will be located in: `<app_root>/lang/shared.js`.

This is similar to the `config` addon, where you have `ac.config.get()` for mojit level configs vs `ac.config.getAppConfig()` to get the app level config.
In this case it would be `ac.intl.lang()` for mojit level langs vs `ac.intl.appLang()` for app level strings.

For example:
The default shared app level lang bundle would be:
`<app_root>/lang/shared.js`

``` javascript
YUI.add("lang/shared", function (Y) {
    "use strict";

    Y.Intl.add(
        "shared",
        "",
        {
            "APP_TITLE": "This is my app title"
        }
    );
}, "0.0.1", {
    requires: ["intl"]
});
```

The localized files (i.e. spanish), would be:
`<app_root>/lang/shared_es-MX.js`

``` javascript
YUI.add("lang/shared_es-MX", function (Y) {
    "use strict";

    Y.Intl.add(
        "shared",
        "es-MX",
        {
            "APP_TITLE": "Este es el título de mi aplicación"
        }
    );
}, "0.0.1", {
    requires: ["intl"]
});
```

And so any controller can get app level strings from the `intl` addon:

``` javascript
YUI.add('CustomHTMLFrameMojit', function (Y, NAME) {

    Y.mojito.controllers[NAME] = {

        index: function (ac) {
            var title = ac.intl.appLang('APP_TITLE');
```

An example use case is for dynamically changing/translating the application title.

### DIFF
--- a/lib/app/addons/ac/intl.common.js
+++ b/lib/app/addons/ac/intl.common.js
@@ -14,6 +14,19 @@
  */
 YUI.add('mojito-intl-addon', function(Y, NAME) {
 
+    function getLang(ac, module, label, args) {
+        var lang, string;
+        lang = Y.mojito.util.findClosestLang(ac.context.lang, ac.instance.langs) ||
+                ac.instance.defaultLang || 'en';
+        Y.Intl.setLang(module, lang);
+        string = Y.Intl.get(module, label);
+        if (string && args) {
+            // simple string substitution
+            return Y.Lang.sub(string, Y.Lang.isString(args) ? [args] : args);
+        }
+        return string;
+    }
+
     /**
      * <strong>Access point:</strong> <em>ac.intl.*</em>
      * Internationalization addon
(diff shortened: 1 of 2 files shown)


caridy:
Since the lang bundles are attached only if the host module is use when mojito runs in the client side, this solution will not work unless you create a dummy module called "shared" and require it in the controller you want to access the app level langs. On the server this is not a problem because all bundles are attached.

rwaldura:
let's resolve this: @gomezd, @caridy, do you have any more comments?

caridy:
I think the new custom frame mojit plug the new global helpers solves most of the problems related to this PR. Aside from that, we should have a better infrastructure once we mix in the modown pieces.

PR #936, opened 2013-01-17
add unit tests for command mojito docs

lzhan:


isao:
+1

is there a reason for two separate test files though?

PR #937, opened 2013-01-17
Clean up process for junebug/hook infrastructure

caridy:
- Run missing tests
- Restore perf levels by not doing extra work when hooks are not enabled
- Remove hooks from store in preparation for the split


isao:
+1

side note, seems like you could add some mock verifications for some of the `adapter = Y.Mock()` you added.

caridy:
There are a bunch of tests without the verify here and there, others that are not using mock at all when they should. Those that I just added don't need verification because I'm not mocking any actual method, but just preparation for later on add more mocks for some of the adapter tests..

PR #938, opened 2013-01-18
`mojito build rescache` for faster startups

drewfish:


isao:
I didn't see any cache invalidation code. Perhaps log cache info and add a cache invalidation flag to `mojito start` for the developer use case?

It looks like you're saving everything in single cache file, so any code/lib change will invalidate. I was going to suggest discrete caches for app, yui, lang, etc. to enable partial invalidation for the developer use case, but I figure that's not what this PR is about.

Also, the `build` command creates new stand-alone applications, whereas the (deprecated) `compile` command creates app work-products. The "rescache" sub-command and might make more sense there. Of course the deprecation makes things confusing too, but just my $0.02.

lib/store.js
```diff
@@ -82,7 +83,8 @@ Store.createStore = function(options) {
     store = new Y.mojito.ResourceStore({
         root: options.root,
         context: options.context,
-        appConfig: options.appConfig
+        appConfig: options.appConfig,
+        skipCache: options.skipCache
```
caridy, line 87: isn't "useCache" more descriptive for production, in which case useCache is false by default during development.

lib/app/autoload/store.server.js
```diff
@@ -404,23 +405,90 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          * @return {nothing}
          */
         preload: function() {
-            // We need to do an initial sweep to find the resource store addons.
-            this.preloadInitial();
-
-            // And then use them.
-            this.loadAddons();
+            var useCache = !this._config.skipCache;
```
caridy, line 408: this._config.useCache sound better!

lib/app/autoload/store.server.js
```diff
@@ -182,6 +182,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._getMojitTypeDetailsCache = {};    // env+posl+lang+mojitType: value
             this._expandSpecCache = {}; // env+ctx+spec: value
 
+            this._resCachePath = this._libs.path.join(this._config.root, 'rescache.json');
```
caridy, line 185: I will recommend to use a folder similar to artifacts, potentially called "build/rescache.json" or "cache/rscache.js" or "cache/rs/cache.json". Two main reasons:
- mixing user-code with generated code could be problematic, specially when it comes to SVN/Git, where a folder is easier to deal with.
- we might want to cache more things, like application.json (merged), or compiled templates, etc.

drewfish:
Re: omitting the `@return`, I'd rather specify explicitly that it returns nothing, instead of having the return value undefined.  (Otherwise, a reader could wonder whether we _forgot_ to document the return type.)

drewfish:
Yeah, I had looked at adding this to `mojito compile` then noticed the deprecation warning.  I didn't know that `mojito build` was for building applications -- I just thought it was useful for building things.  I think I don't understand the semantic difference between "build" and "compile".  I'll move to `mojito compile` (and remove the deprecation warning).

Cache invalidation is accomplished by deleting the file :smile: .  This is something we'll have to document.

drewfish:
I though about `useCache` instead, but unfortunately if it's missing that's falsy, which would lead one to think that the cache shouldn't be used.  By default we want the cache to be used (if it exists).  (Sure, we could play tricks with `hasOwnProperty()` but that doesn't solve the semantic inconsistency.)

drewfish:
Yeah, I'll move the file out of the app root.  I'll put it under `artifacts/` (probably `artifacts/cache/`) since the whole purpose of `artifacts/` is to hold generated files.

caridy:
- I'm not so sure about artifacts, in fact, SD is ignoring artifacts/ by default when creating a pkg.
- useCache: I'm not so sure we want to have that enabled by default, it sounds like a manual process to build the cache, and enable it in the application.json, and it is definitely an advanced option.
- @return, I think yui is just avoiding adding the return statement if there is nothing to return. Yes, it might lead to missed docs but it is not a big deal.

drewfish:
RE useCache/skipCache:  the idea is that if the cache file exists it will be used.  So, by default the cache is disabled, because it doesn't exist :smile: .  If the user doesn't run `mojito compile rescache` then there won't be any caching.  If the user runs `mojito compile rescache` then the cache file is created and the resource store will pick it up.  Don't want to use the cache anymore? Delete it!

I argue that the resource cache isn't that advanced a feature -- it's a production optimization and will hopefully become a common build step for our users.  Skipping the cache if it exists is a more advanced use case, I think.

drewfish:
RE `artifacts/`, hmm... good point.  The old/deprecated `mojito compile ...` commands in fact would put their results in `mojits/{mojit}/autoload/compiled/`.

The trick with `cache/` is that there are some cases where caches are writing to a filesystem, but those caches _shouldn't_ be shipped with the application.  Imagine CI step that starts up the app and runs functional tests on it.  The caching that happens as part of the functional tests might not be appropriate to ship to production.

However, this new `rescache.json` definitely is something that should be shipped.  Yet, it shouldn't be checked into SCM.  This is unusual.

As a side note, I think `npm pack` will also ignore files listed in a `.gitignore` file (if the `.npmignore` file doesn't exist).

drewfish:
Closing pull request, since this has some tricky issues and doesn't address the underlying needs of the customer (search).

PR #942, opened 2013-01-20
Reduced memory consumption at server start

jlecomte:
This meta-data copy operation seems unnecessary to me, although I would like someone else to confirm before merging this PR. This change reduces the memory consumption at server start by almost 40% in our case (from ~160MB to <100 MB) because we have so many language resource bundles.


drewfish:
Yeah, that copy seems unnecessary to me as well.  Travis builds passed.

+1

isao:
+1

Issue #948, opened 2013-01-23
ac.composite.done() to specify the view to follow the behavior of ac.done()


kinlum:
If we do a 

```
ac.done(data, {"view": {"name": "detail"}});
```

but then, if now this Mojit has a child or children, then we need to use `ac.composite` like this:

```
ac.composite.done({ "template" : data });
```

So it works, but we can't specify the  `"view": {"name": "detail"}`  part easily.  If `ac.done` and `ac.composite.done` work mostly the same except one has no children and one does, would it be ok if `ac.composite.done` can have the feature to work like the following:

```
ac.composite.done({ "template" : data, "view": {"name": "detail"} });
```

or 

```
ac.composite.done({ "template" : data, "view": {"name": "index"} });
```

---

As a side note, I wonder why we don't just have `ac.done()` and it will check automatically whether it has children or not, so that if it does, then it will process just like `ac.composite.done()` does.  Is it really for speed concerns?  Otherwise in some GUI frameworks, if a view has a subview, it is automatically drawn, and we won't need to pay special attention as to whether a view has children or not.

caridy:
I can take this for 0.5.3, it seems a low-impact change. About the ac.done vs ac.composite.done, that have to be discussed. But basically, adding complexity into the mix is a recipe for failure. They are not the same thing, while composite is way more complex.

caridy:
Closing this now that PR #950 was merged. It will be ready in 0.5.3, in the meantime, you can try it using "develop" branch. The details on of syntax is in the PR. Basically ac.done and ac.composite.done are equivalent in terms of API.

PR #950, opened 2013-01-23
support custom meta and custom template data for ac.composite.done() issue #948

caridy:
This PR tries to match the api of ac.done() and ac.composite.done() now that we don't accept passing params and other data into all children.

The new API will work like this:

```
ac.composite.done({
    "title" : "something"
}, {
    "view": {
        "name": "detail"
    },
    "assets": {
        "top": {
            "js": ["something.js"]
        }
    }
});
```

The old API is still supported, in which case this will also work:

```
ac.composite.done({
    "template:" {
        "title" : "something"
    }
});
```

but a warning will be issued to announce the deprecation.

TODO: 
- Update examples
- Update documentation

### DIFF
--- a/lib/app/addons/ac/composite.common.js
+++ b/lib/app/addons/ac/composite.common.js
@@ -154,17 +154,17 @@ Y.mojito.controller = {
             }
 
             this.execute(cfg, function(data, meta) {
-                // clearing the metas collected from childs to avoid a child
-                // to dictate how the parent will behave.
-                meta.view = {};
-
+                parentMeta.assets = Y.mojito.util.metaMerge(
+                    parentMeta.assets || {},
+                    meta.assets || {}
+                );
                 // 1. templateData and data are normally exclusive, in which case
-                // the prority is not relevent.
+                // the prority is not relevent. In which case we set D
                 // 2. parentMeta and meta should be merged to preserve the children
                 // binders map, assets, etc. but giving parentMeta the priority in case
                 // a custom configuration in the parent should overrule something coming
                 // from the chindren merged meta
-                ac.done(Y.merge(templateData, data), Y.mojito.util.metaMerge(meta, parentMeta));
+                ac.done(Y.merge(data, templateData), Y.mojito.util.metaMerge(parentMeta, meta));
             }, this);
         },
 
--- a/lib/app/addons/ac/composite.common.js
+++ b/lib/app/addons/ac/composite.common.js
@@ -154,15 +154,17 @@ Y.mojito.controller = {
             }
 
             this.execute(cfg, function(data, meta) {
-                if (meta.assets) {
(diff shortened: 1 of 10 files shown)


drewfish:
<del>I worry about taking away the ability to pass parameters.  Perhaps we can email the users list to see if anyone is using that feature.</del> Nevermind, that was never a feature of the `ac.composite.done()` arguments.

lib/app/addons/ac/composite.common.js
```diff
@@ -150,9 +154,15 @@ Y.mojito.controller = {
             }
 
             this.execute(cfg, function(data, meta) {
-                var merged = Y.merge(template, data);
-                ac.done(merged, meta);
+                if (meta.assets) {
```
drewfish: I think you always need to metaMerge() in parentMeta, otherwise it'll never use parentMeta.view.

drewfish:
Also, we should probably update `examples/` and `tests/` to use the new API.

lib/app/addons/ac/composite.common.js
caridy: meta.assets is merged to parentMeta "optionally", but parentMeta is the one passed into the ac.done routine.

drewfish: What about other fields of `meta` besides `meta.assets` ?  We don't want to drop those on the floor.

caridy:
Travis is still unstable, one job is green the other fails due to arrow issues. This is ready for the final review /cc @drewfish

lib/app/addons/ac/composite.common.js
```diff
@@ -150,9 +154,17 @@ Y.mojito.controller = {
...
+                    parentMeta.assets || {},
+                    meta.assets || {}
+                );
+                // 1. templateData and data are normally exclusive, in which case
+                // the prority is not relevent. In which case we set D
+                // 2. parentMeta and meta should be merged to preserve the children
+                // binders map, assets, etc. but giving parentMeta the priority in case
+                // a custom configuration in the parent should overrule something coming
+                // from the chindren merged meta
+                ac.done(Y.merge(data, templateData), Y.mojito.util.metaMerge(parentMeta, meta));
```
drewfish, line 167: If we're doing this metaMerge(), do we still need the metaMerge() on line 157 (just above)?

caridy, line 167: Actually, this is a question for you. Apparently, metaMerge does not work well for very deep structures, like meta.assets. The solution I have so far is to merge assets object, then merge the rest.

drewfish, line 167: OK, sounds good then.  I didn't realize the metaMerge() was so particular.

drewfish:
+1

PR #952, opened 2013-01-24
supporting applicationConfigFiles definition in a setting other than master

caridy:
specifically supporting runtime:common or runtime:server, e.g.: 

```
[
    {
        "settings": [ "runtime:common" ],
        "applicationConfigFiles": [
            "another/application.json",
            "master-setting.json"
        ]
    },
    {
        "settings": [ "master" ],
        "testKey1": "testVal1",
        "testKey2": "testVal2",
        "testKey3": "testVal3"
    }
]
```

So, you can get a master setting that is shared between apps, and include it as part of the root application.json

### DIFF
--- a/lib/app/addons/rs/config.js
+++ b/lib/app/addons/rs/config.js
@@ -351,7 +351,13 @@ YUI.add('addon-rs-config', function(Y, NAME) {
             // since application.json is optional, we should be careful
             // in any case there is a low-level cache mechanism going on here.
             if (Y.Lang.isArray(this.readConfigSimple(rootAppJSON))) {
-                ycb = this.readConfigYCB(rootAppJSON, {});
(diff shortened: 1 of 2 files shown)


drewfish:
+1

caridy:
Travis is still unreliable, one job passed, the other fail with timeout:
test simplemodelclient:Unexpected error: Y.one(...) is null

jlecomte:
I'm not a big fan of this PR. I would have preferred the original proposal which was to be able to have configurations set for the same contexts (included settings: master) in multiple files as long as they are not the same keys. Using runtime: common seems like a hack to me...

caridy:
@jlecomte, I'm not proud of it, lol, but we plan to remove all that once we landed the "import_settings" feature. We are just trying to unblock you guys in the meantime.

PR #953, opened 2013-01-25
skip packages that have already been visited

drewfish:


lib/app/autoload/store.server.js
```diff
@@ -1093,6 +1095,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this.selectors = {};
             this._appRVs = [];
             this._mojitRVs = {};
+            this._packagesVisited = {};
```
caridy, line 1099: Why is `_packagesVisited` initialized twice?

caridy:
+1

lib/app/autoload/store.server.js
drewfish, line 1099: It's initialized for every call to preloadResourceVersions().  Internally, preload() calls preloadResourceVersions() twice.  The first time is mainly to find RS addons (which can also be found in node_modules).  The second time is the "real" load, done once the store has all its addons plugged in.  If we didn't clear out packagesVisited then the RS addons wouldn't get a chance to see the resources in the NPM modules.

Said another way, the packages (and app itself) are used to populate _appRVs and _mojitRVs.  Anytime we (re)initialized those structures we need to (re)initialize _packagesVisited as well, so that those RV structures can get repopulated.

caridy, line 1099: got it!

PR #954, opened 2013-01-26
App fails to boot in nodejitsu

caridy:
App fails to boot in nodejitsu due to a callback wrapper that is synthetically created when calling `app.listen(8080);` in server.js. Apparently this callback interrupts the control, and make the app to hang. To solve the problem, we are removing the wrapper in favor of the original callback or nothing, in other words callback argument for app.listen() is optional now.


isao:
+1

caridy:
Travis is still unstable. One passed, the other fail :(

PR #955, opened 2013-01-28
prefer store.getStaticAppConfig()

drewfish:
... over store.getAppConfig(store.getStaticContext())


caridy:
+1

Note: I guess you will need to update the tests, for the store mock maybe.

drewfish:
Yep, I'll get tests to pass before merging.

PR #958, opened 2013-01-29
fix unit tests after newsboxes code was updated for pull #842

isao:
app no longer has that extra "template" property


drewfish:
+1

isao:
travis has errors ff/selenium unrelated

PR #973, opened 2013-02-07
Fix for client side hooks, bug fix in template hooks.

bthaeler:
To use hooks on the client side, the globalHookHandler has to be available for addons. Send pass it around using life cycle events in mojito client.

Fixed bug introduced by Caridy in hb.server.js. Hooks start call has to be called before cache code.

### DIFF
--- a/lib/app/autoload/mojito-client.client.js
+++ b/lib/app/autoload/mojito-client.client.js
@@ -332,9 +332,13 @@ YUI.add('mojito-client', function(Y, NAME) {
         init: function(config) {
             var forwardConfig = {
                 config: config,
+                // pass globalHookhandler to addons that may want to use hooks
                 globalHookHandler: globalHookHandler
             };
             fireLifecycle('pre-init', forwardConfig);
+            // if we didn't originaly have hooks enabled, copy back from config object.
+            // This is the case where an add-on module wants to turn on hooks and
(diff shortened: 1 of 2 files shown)


lib/app/autoload/mojito-client.client.js
```diff
@@ -330,7 +330,14 @@ YUI.add('mojito-client', function(Y, NAME) {
...
         init: function(config) {
-            fireLifecycle('pre-init', {config: config});
+            var forwardConfig = {
+                config: config,
+                globalHookHandler: globalHookHandler
+            };
+            fireLifecycle('pre-init', forwardConfig);
+            if (!globalHookHandler) {
+                globalHookHandler = forwardConfig.globalHookHandler;
+            }
```
drewfish, line 346: Reading the global out of forwardConfig after setting it there is a bit confusing.  I understand what the code is doing but I don't understand _why_.  There should probably be a comment explaining why we need to `globalHookHandler = forwardConfig.globalHookHandler`.

bthaeler, line 346: Comment added. Also fixed lint issue.

drewfish:
+1.  I ran all tests locally and they all passed.  I'll merge after our internal builds stabilize.

drewfish:
Sorry for the pause, it took a bit longer than I expected :(

PR #975, opened 2013-02-09
write results in dir out side of working folder as arrow0.0.73 changes

lzhan:
When switch arrow from 0.0.71 to 0.0.73, our CI report dir only has the last app tests report filed left. All reports from other apps got deleted. As arrow team suggested, if result dir is outside of working folder, all results will be saved. So I am making this change.


drewfish:
+1

Issue #865, opened 2012-12-17
Ac.Composite Error Handling Hides Errors from Parent


ooskapenaar:
In v0.5.0  composite.common.js we see the following code, which effectively hides any child errors from the caller. 

If the parent is to be allowed to "fail-fast", which I think is a good thing, then mojito must pass errors up to the parent and not swallow them like this. 

Swallowing of errors by mojito should only be allowed at the top level (i.e. to clean up after a messy parent).

``` javascript
       error: function(err) {
            Y.log("Error executing child mojit at '" + this.id + "':", 'error',
                    NAME);
            if (err.message) {
                Y.log(err.message, 'error', NAME);
            } else {
                Y.log(err, 'error', NAME);
            }
            if (err.stack) {
                Y.log(err.stack, 'error', NAME);
            }
            // Pass back some empty data so we don't fail the composite
            this.done('');
```

I am going to patch this code, by passing an error object back as the data, then the parent can sort out the mess, an alternative would be to throw an error and disrupt the composite process completely:

``` javascript

           if( err instanceof Error ) {
                this.done( err );
            } else {
                this.done( new Error( err ) );
            }

```

Opinions, ideas?

caridy:
Adding this to trello, tentative for 0.5.4:

https://trello.com/card/issue-gh865-composite-ac-composite-error-handling-hides-errors-from-parent/504e5e3b7cd1578456fc430f/367

rwaldura:
see also issue #25

caridy:
will be available in mojito@0.5.5 scheduled for next week.

Issue #25, opened 2012-03-30
[bz4788267] Child errors are not communicated clearly to parents


add0n:
The default error() function the composite plugin uses to handle child errors looks like this:

``` javascript
        error: function(err) {
            Y.log("Error executing child mojit at '" + this.id + "':", 'error', NAME);
            Y.log(err.message, 'error', NAME);
            Y.log(err.stack, 'error', NAME);
            // Pass back some empty data so we don't fail the composite
            this.done('');
        }
```

In order to allow the parent mojit to continue running, it passes back an empty string into ac.done(), but we should have a way to communicate child errors

caridy:
This is still an issue. Let's bump this up. /cc @rwaldura @jenny

caridy:
will be available in mojito@0.5.5 scheduled for next week.

PR #978, opened 2013-02-12
Fix for issue #865: composite addon error manipulation

caridy:
This PR attempts to solve issue #865 and #25.

An important child fails thru a call to `ac.error`, as a result, the parent should also fail. For this use-case, we are proposing a flag at the child level configuration:

```
{
    "type": "ParentMojit",
    "config": {
        "children": {
            "slot1": {
                "type": "ReallyImportant",
                "propagateFailure": true,
                "config": {}
            },
            "slot2": {
                "type": "NotImportant",
                "propagateFailure": false,
                "config": {}
            },
            "slot3": {
                "type": "NotSoImportantWithDefaultSetting",
                "config": {}
            }
        }
    }
}
```

In the example above, `slot2` and `slot3` can fail, in which case the data passed into the parent view will be an empty string. And that's the default behavior. In the other hand, if the mojit in `slot1` fails, the error will be propagated, and `ParentMojit` will also fail by calling `ac.error()` and piping the original error from child on `slot1` due to the special configuration `propagateFailure`, which is set to `true`. This give us the control needed to define important childs.


ooskapenaar:
+1

caridy:
this is ready for review. /cc @drewfish

lib/app/addons/ac/composite.common.js
```diff
@@ -232,147 +229,150 @@ callback({
...
         },
 
+        abort: function (err) {
+            var adapter = this.adapter;
+            // an error occur in a child marked with propagateFailure flag
+            adapter.error(err);
+            // cleaning up the house
+            this.queue.results = [];
+            // cancel any pending job by invalidating the callback of the queue
+            this.queue.skip = true;
```
drewfish: The `skip` property doesn't appear to be part of the Parallel API.  Can we do `this.skip` instead?

lib/app/addons/ac/composite.common.js
```diff
@@ -232,147 +229,150 @@ callback({
...
-                    // Make a new "command" that works in the context of this
-                    // composite
-                    newCommand = {
-                        instance: child,
-                        // use action in child spec or default to index
-                        action: child.action || 'index',
-                        context: command.context,
-                        params: child.params || command.params
-                    };
+                    return {
```
drewfish, line 362: Might be nice to add a comment here that this datastructure is part of the `buffer` argument to the callback of `this.queue.done`.  (It took me a little bit to figure this out, in part because the documentation for Y.Parallel is so thin.)

lib/app/addons/ac/composite.common.js
```diff
@@ -232,147 +229,150 @@ callback({
...
-                    }
+            meta.children = children;
+
+            for (child in children) {
+                if (children.hasOwnProperty(child)) {
+                    this.add(child, children[child]);
+                }
+            }
+
+            this.queue.done(function (buffer) {
```
drewfish: Can we rename the `buffer` argument to `results` or `childResults` or something like that?  `buffer` isn't very descriptive.

lib/app/addons/ac/composite.common.js
```diff
@@ -232,147 +229,150 @@ callback({
...
+        abort: function (err) {
+            var adapter = this.adapter;
+            // an error occur in a child marked with propagateFailure flag
+            adapter.error(err);
+            // cleaning up the house
+            this.queue.results = [];
+            // cancel any pending job by invalidating the callback of the queue
+            this.queue.skip = true;
+            // invalidate this.adapter.done/error
+            adapter.error = adapter.done = function () {};
```
drewfish: I wonder if we can put do some useful logging in this replacement, instead of just no-op.

(The situation is made a little more confusing because ac.composite doesn't actually call this.adapter.done() itself, but instead calls the `cb` argument to ac.composite.execute().  However, we _do_ call this.adapter.error() whenever a child fails and propagates that failure.)

drewfish:
Looks pretty good, but the error handling could be a little cleaner I think.  Instead of `this.queue.skip` perhaps use `this.failed`.  Then, it might be a little clearer how/when we should call this.adapter.error() and when we shouldn't.

lib/app/addons/ac/composite.common.js
```diff
@@ -232,147 +229,150 @@ callback({
...
+                cb(content, meta);
+            });
         },
 
+        abort: function (err) {
+            var adapter = this.adapter;
+            // an error occur in a child marked with propagateFailure flag
+            adapter.error(err);
+            // cleaning up the house
+            this.queue.results = [];
```
drewfish: I understand why we're doing this, but I don't think this is the best approach.  Perhaps instead we can do `this.queue = null` if that works.

In fact, we should be clearing out `this.queue` regardless of whether we succeed or fail.

lib/app/addons/ac/composite.common.js
drewfish: Hmm... perhaps it might be clearer to use `this.queue.results` instead of `buffer`.

drewfish:
Here are some ideas I had:
https://github.com/drewfish/mojito/commits/composite-errors

caridy:
@drewfish few notes after the last commit:
- `this.failed` in now in place, but this.queue is not destroyed (to null). The reason for that is people can try to add childs at any time using `addChild` public method. Whether that's before of after an error occur, it does not matter.
- I made `_onChildFailure` private, we don't want people to use that. Also, I figured that by calling adapter.error() in the first place should invalidate adapter.done and any succesive call, so we are covered there.
- Finally, the `failFast` thing, I don't think we need that. If there is an explicit propagation flag, an error should fail fast. If there is not such as flag in any child, then there is not need to worry about errors, not need to aggregate them, and there is no way to consume them, so, I think we can leave that out.

drewfish:
+1

PR #981, opened 2013-02-14
some more unit test commands

drewfish:


isao:
+1

drewfish:
Yeah, I'll make that change.  I was following the same format as commands/gv, but commands/compile doesn't need that.

PR #984, opened 2013-02-14
changes for screwdriver:start browser on remote selenium host and reuse session

lzhan:


tests/run.js
```diff
@@ -37,6 +41,16 @@ program.command('test')
...
+dns.lookup(hostname, function (err, addr, fam) {
+    hostip = addr;
+    console.log('App running at......' + hostip);
+})
+
+if (process.env['SELENIUM_HUB_URL']) {
+    remoteselenium = process.env['SELENIUM_HUB_URL'];
+    console.log('selenium host.....' + remoteselenium);
+}
+
```
drewfish: I think this should go inside the `test()` method, since I generally think code outside of methods is a bad idea.

Also, since the DNS lookup could take a while, it should really be added (at the top of the) async series, so that we can wait until it's done before doing the other tasks.

tests/run.js
```diff
@@ -82,6 +93,14 @@ function test (cmd) {
     async.series(series, finalize);
 }
 
+function gethostip(callback){
+    dns.lookup(hostname, function (err, addr, fam) {
+        hostip = addr;
+        console.log('App running at.....' + hostip);
+    });
+    callback(null);
```
drewfish: I think this callback should go inside the anonymous function passed to `dns.lookup()`.

Also, couldn't hurt to sneak in some error checking, something like:

``` javascript
if (err) {
    callback(err);
    return;
}
```

drewfish:
+1

PR #985, opened 2013-02-15
leaner meaner resource store

drewfish:
This changes the resource store to hold onto less data at runtime.  Notable changes are:
- store.getResources() is now gone.
- AOP methods/events during preload() have been reworked, so that metadata can be kept in a leaner format.
- store.optimizeForEnvironment() so that store can drop data that isn't used during runtime in production.  Thus, after the server starts store.getResourceVersions() is no longer appropriate.


lib/mojito.js
```diff
@@ -430,21 +432,16 @@ MojitoServer.prototype._configureLogger = function(Y) {
...
-    Y.applyConfig({
-        modules: modules
-    });
+    modules = store.yui.getModulesConfig('server', false);
+    Y.applyConfig(modules);
 
     // pre-loading every yui module for the server runtime
-    for (module in modules) {
-        if (modules.hasOwnProperty(module)) {
+    for (module in modules.modules) {
```
caridy: We can replace this cycle with a `concat` and `Object.keys(modules)`.

caridy:
+1

It is looking pretty solid. When are we going to get travis back?

lib/mojito.js
drewfish: Sure.

drewfish:
Will this PR fix travis?  No unfortunately :( :)

caridy:
+1

PR #987, opened 2013-02-15
Update README.md

akshar100:
Added instructions that will be helpful for those who want to contribute to mojito documentation.


rwaldura:
+1
Thank you for signing our CLA!

zhouyaoji:
Thanks, akshar100. You couldn't actually run 'make html' to build the docs before because we didn't have a Makefile that would call Sphinx. I am guessing that you used Shaker before as they had a Makefile that allowed you to build the documentation. In any case, I have added a Makefile and Sphinx config and template files so that you now can run 'make html' to build the docs locally. Thanks again.

akshar100:
Thanks. My objective behind this patch was to get familiar with the whole process. Now that I have understood it I will start contributing more. :)

PR #995, opened 2013-02-20
integrated context-cache

drewfish:


lib/app/addons/rs/config.js
```diff
@@ -169,25 +169,20 @@ YUI.add('addon-rs-config', function(Y, NAME) {
...
-            ycb = this._ycbCache[fullPath][cacheKey];
+            key = 'ycb-' + fullPath;
+            ycb = store.contextCacheGet(ctx, key);
             if (!ycb) {
                 json = this.readConfigSimple(fullPath);
                 json = this._ycbDims.concat(json);
                 ycb = libycb.read(json, ctx);
-                this._ycbCache[fullPath][cacheKey] = ycb;
+                store.contextCacheSet(ctx, key, ycb);
+                ycb = Y.mojito.util.copy(ycb);
```
caridy: Why do we need to copy it? This operation is a one time operation, and at the moment we cache it, the original can be modify in the current execution, while a new copy will be produce whenever the same value is requested.

tests/unit/lib/app/addons/rs/test-config.js
```diff
@@ -46,6 +46,14 @@ YUI().use('addon-rs-config', 'mojito-util', 'mojito-test-extra', 'base', 'oop',
 
         parseResourceVersion: function(source, type, subtype, mojitType) {
             // no-op
+        },
+
+        contextCacheGet: function(ctx, key) {
```
caridy, line 51: I will like to see `store.set` and `store.get` as generic methods to cache and retrieve entries based on a context that could be used by any RS addon.

drewfish:
Closing for now, since this PR regresses performance :(

Issue #982, opened 2013-02-14
combo~ URL returning 404 in HTML5App


gvaish:
I am working on `todomvc` (https://github.com/addyosmani/todomvc) using Mojito.

The code is available at https://github.com/gvaish/todomvc-mojito/tree/master/todomvc-mojito-pure using `controller` with `common` affinity.
It works great.

However, after html5app build, the `index.html` has the combo URL as https://github.com/gvaish/todomvc-mojito/blob/gh-pages/pure/index.html#L18

This returns a 404.

What's the way around? Am I missing any option to `build` command?

gvaish:
btw, I've used the following config:

```
"builds": {
    "html5app": {
        "forceRelativePaths": false,
        "attachManifest": true
    }
},
"staticHandling": {
    "appName": "todomvc",
    "useRollups": false
}
```

caridy:
@gvaish I think we will have to revisit this use-case, specially the use of relative paths. In general, a combo is just an artifact to support your app that it happens to be part of the app itself, and it should not get affected by `forceRelativePaths` since it will not work in a pure html5 type of app.

So, change number one: disable combo for your app:

```
"yui": {
     "config": {
           "combine": false
     }
}
```

Change number two: if you want to serve yui from your app, you need some special settings to do so. If you want to serve from CDN, the default configuration should help. Here is the doc about yui configs:

http://developer.yahoo.com/cocktails/mojito/docs/topics/mojito_yui_config.html

gvaish:
@caridy Thanks.
`yui.config.combine=false` works great.

However, cannot get `base` or `root` working... I want to serve YUI from Yahoo-CDN. How do I do that?
Nor is `comboBase` working.

All in all, it's only combine or don't. Can't use YUI served from URL like http://yui.yahooapis.com/3.8.1/build/

gvaish:
Tried lot of hands... but failed to get `html5app` use relative `./static` path in `loader-app-base.js` that gets generated.  :-(

caridy:
@gvaish you're facing a configuration issue. @zhouyaoji  and myself are working on a very extensive piece of doc about yui configuration in mojito, still preliminar (https://github.com/zhouyaoji/mojito/blob/docs_yui_seed_update/docs/dev_guide/topics/mojito_yui_config.rst#getting-yui-to-your-application)

[...]

PR #997, opened 2013-02-21
seed refactor

caridy:
Goals:
- Reduce the memory allocation when YUI core modules should not be served from local server (default configuration)
- Clear separation between YUI core modules and app-level modules (including synthetic modules)
- As part of the separation, the seed will have, at least, 2 files, YUI Core Seed + App Seed.
- Support for relative path when using mojito build command.

Changes:
- moving logic to generate seed files into yui rs addon from deploy ac addon.
- making root and base for yui and app group to correlate.
- introducing staticHandling->serveYUIFromLocal configuration to disable serving yui from local app by default to preserve memory.

Basically, this PR will allow users to switch to server YUI from local, including combo, but the default configuration will be to load it from CDN. If you want to force to use YUI from local for development, or to be able to build html5apps and hybrid apps, you can use this setting:

```
{
    "settings": [ "environment:development" ],
    "staticHandling": {
        "serveYUIFromLocal": true
    }
}
```

Keep in mind that when using `mojito build` command, the `combine` flag should be set to `false` under `yui.config` because combo is not available in stand alone apps.

Other changes in this PR:
- supporting html5app->forceRelativePaths for yui loader, making root and base to be relative when needed.
- making default config.json a yaml file to add proper comments about default configurations.

TODO: 
- Update docs


caridy:
This is also suppose to fix issue #982.

isao:
(redacted)

caridy:
@isao can you elaborate more?

isao:
@caridy made this [gist](https://gist.github.com/isao/5009030) to explain, before I saw the full context of the commit (I mistakenly thought you replaced the str.replace instead of adding a second one).

sorry, nm :) 

:+1:

imalberto:
+1

PR #999, opened 2013-02-21
re-enable, upgrade selenium-server-standalone-2.30.0.jar

isao:
will this go green?


drewfish:
+1

BTW, upgrading selenium is a prerequisite to upgrading yahoo-arrow beyond 0.0.73.

PR #1007, opened 2013-02-25
re-enable context support when producing yui.config

caridy:


caridy:
This fixes Jenkins when testing custom logs per client and server with different contexts.

drewfish:
+1

Issue #1006, opened 2013-02-25
build hybrid app regression


isao:
from Ji liang 2/24:

DP-Integration-Test is failing on mojito 0.5.4-133-g443bf5e

To reproduce this:
1.install latest IDE plugins from here:
http://<redacted>/update,
 download the latest Mojito, CRT by IDE download manager.
2. Create new mojito project,right click the project ->content menu->run
as -> ipad app on simulator
3. Then index rendered was not right, and expand the
app/artifacts/builds/hybridapp/yahoo.application.appname/ there was no
index.html.

Ji liang

caridy:
It might be related to PR #997 and the upcoming fix #1007

isao:
fix confirmed by Ji Liang

PR #1008, opened 2013-02-25
fix hybrid build

drewfish:


lib/app/autoload/store.server.js
```diff
@@ -602,6 +602,10 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 if (details.defaults && details.defaults.config) {
                     details.config = Y.mojito.util.blend(details.defaults.config, details.config);
                 }
+                // we shouldn't expose this
+                if ('client' === env) {
+                    details.fullPath = undefined;
```
caridy, line 607: I recall the static handler using the fullpath at some point when computing client side YUI stuff. Let's make sure this is not breaking that part of the system.

drewfish, line 607: I was worried about this too, so (a) this change is in expandInstanceForEnv() so that it happens as late as possible (after all RS addons), and (b) I ran all the tests (unit & func) locally before issuing this PR.

isao:
+1

pls reference issue #1006 in merge commit

PR #1011, opened 2013-02-26
[bz5800733] Handlebars Template Helpers (second attempt after #922)

caridy:
This PR introduces the concept of helpers thru an addon called `mojito-helpers-addon`. Any mojito requiring this addon will be able to:
- register a helper function at the instance level
- register a helper function and expose it at the request level (when in the server)
- register a helper function and expose it at the page level (when in the client)
- expose an already registered helper function
## API
- `ac.helpers.set(name, func)` to set a local helper
- `ac.helpers.get(name)` in case you want to get a handler on a global or even local that was registered somewhere
- `ac.helpers.expose(name, func)` to make it available locally but also exposing it globally
- `ac.helpers.expose(name)` in case you want to expose an existing helper.
## How does it work?

Under the hood, any helper function, locally set using `ac.helpers.set()` or globally exposed by any of the parent mojits using `ac.helpers.expose()` will be passed into the handlebars render engine, so you can use them in your templates/views.
## Limitations and things to keep in mind
- You will be able to override any default helper provided by Y.Handlebars since the custom helper functions will have priority. This also mean you can break the default handlebar helpers, so be careful.
- You will not be able to use helpers when using `mojitProxy.render()` or `mojitProxy.refreshView()`, only thru `mojitProxy.invoke()`.
- You will not be able to use helpers when using `mojitProxy.invoke()` if the helper is supposed to be defined by another mojit that was not instantiated prior to the invoke.
## Examples

https://github.com/caridy/mojito-demos/tree/master/mojito-helpers

### DIFF
--- a/lib/output-handler.server.js
+++ b/lib/output-handler.server.js
@@ -30,6 +30,7 @@ var NAME = 'OutputHandler.server',
         this.res = res;
         this.next = next;
         this.headers = {};
+        this.page = {};
     };
 
 
--- a/lib/app/addons/ac/helpers.common.js
+++ b/lib/app/addons/ac/helpers.common.js
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
+ * Copyrights licensed under the New BSD License.
+ * See the accompanying LICENSE file for terms.
+ */
+
+/*jslint nomen:true*/
+/*global YUI*/
+
+/**
+ * @module ActionContextAddon
+ */
+YUI.add('mojito-helpers-addon', function (Y, NAME) {
+
+    'use strict';
+
+    /**
+     * <strong>Access point:</strong> <em>ac.helpers.*</em>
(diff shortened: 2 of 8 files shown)


caridy:
ping issue #922

lib/output-handler.server.js
```diff
@@ -30,6 +30,7 @@ var NAME = 'OutputHandler.server',
         this.res = res;
         this.next = next;
         this.headers = {};
+        this.page = {};
```
drewfish, line 33: Oooooooh, me like.

drewfish:
+2

rwaldura:
@gotwarlost check this out, released in 0.5.5 today 
https://github.com/yahoo/mojito/wiki/ReleaseNotes0_5_5

gotwarlost:
already incorporated :)

caridy:
Just keep in mind that this feature is EXPERIMENTAL and API might change in the future. If you use it, and it is fine to use it, you will be in the bleeding edge of mojito :).

PR #1016, opened 2013-03-01
fixing regression in mojito@0.5.5 that breaks shaker

caridy:


rwaldura:
Is this the final fix, that yields a new Shaker release?
Either way, you’re awesome.

drewfish:
+1

rwaldura:
I want to use this issue to keep track of all issues related to this compatibility break.
E.g. see also yahoo/mojito-shaker#43

caridy:
@rwaldura this is the only change for the patch. Once we get confirmation from Albert, we can issue the patch or just 0.5.6, whatever works best.

As a side note, this is not related to yahoo/mojito-shaker#43, it was a bug I introduced as part of PR #997

Issue #1026, opened 2013-03-11
Compatibility with Node.js 0.10


rwaldura:
@gotwarlost says:

Looks like the Mojito team may have some work. I installed v0.10.0 to test istanbul and then started a mojito app. This app starts up correctly with node v0.8.x.

```
path.js:360
        throw new TypeError('Arguments to path.join must be strings');
        ^
TypeError: Arguments to path.join must be strings
    at path.js:360:15
    at Array.filter (native)
    at Object.exports.join (path.js:358:36)
    at ResourceStore.Y.extend.resolveMojitDetails
(.../mojito/lib/app/autoload/
store.server.js:710:49)
    at ResourceStore.Y.extend.resolveResourceVersions
(.../mojito/lib/app/autoload/
store.server.js:1542:38)
    at i.Method.exec
(.../mojito/node_modules/yui/
event-custom-base/event-custom-base-min.js:2:1252)
    at ResourceStore.o.(anonymous function).r.(anonymous function) [as resolveResourceVersions] (.../mojito/node_modules/yui/
event-custom-base/event-custom-base-min.js:2:521)
    at ResourceStore.Y.extend.preload
(.../mojito/lib/app/autoload/
store.server.js:406:18)
    at Object.Store.createStore
(.../mojito/lib/store.js:91:1
5)
    at new MojitoServer
(.../mojito/lib/mojito.js:69:
29)
'''

```

isao:
error occurs here when res.url is null: https://github.com/yahoo/mojito/blob/develop/lib/app/autoload/store.server.js#L712

in node 0.8:

```
% node -pe 'require("path").join(null)'
.
```

in node 0.9+ 

```
% node -pe 'require("path").join(null)'

path.js:360
        throw new TypeError('Arguments to path.join must be strings');
              ^
TypeError: Arguments to path.join must be strings
```

isao:
pull #1028 fixes the exception above, and mojito starts with newsboxes.. there may be other errors related to path.join(null) on node 0.9/0.10

caridy:
Weird, `this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url)` seems to be ok, unless that there is no `url` for the `res`, which is even more weird. /cc @drewfish

caridy:
Solved in mojito@0.5.6

PR #1028, opened 2013-03-11
fix issue #1026 path.join throws on null param in node 0.9+

isao:
quick patch to prevent exception on mojito start in node 0.9, 0.10

addresses change where, in latest node versions, path.join throws if given a null value, instead of coercing it to ""

### DIFF
--- a/lib/app/autoload/store.server.js
+++ b/lib/app/autoload/store.server.js
@@ -709,7 +709,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 if (res.type === 'view') {
                     template = {
                         'content-path': (env === 'client' ?
-                                this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url) :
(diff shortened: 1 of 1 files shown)


lib/app/autoload/store.server.js
```diff
@@ -709,7 +709,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 if (res.type === 'view') {
                     template = {
                         'content-path': (env === 'client' ?
-                                this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url) :
+                                this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url || '') :
```
caridy, line 712: We know `pathToRoot` could be undefined since it is an optional flag, but res.url for a view, that's just not right. We need to get to the bottom of this.

caridy:
-1

caridy:
We should consider dropping this in favor of PR #1029.

caridy:
Closing in favor of #1029

PR #1029, opened 2013-03-11
fix issue #1026: no more exceptions for mojito mojits

caridy:
they should be playing by the same rules that regular mojits.


drewfish:
+1

Anything to put in HISTORY.md about this?

PR #1030, opened 2013-03-11
rm package.json preferGlobal:true

isao:
and add deprecation notice about installing mojito globally.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -45,6 +45,9 @@ Here are details of the API changes.
 
 Please see the API docs for details of each.
 
+Installing mojito globally (i.e. `npm install --global mojito`) is deprecated. Please install mojito locally in your application.
+
+
 Features
 ------------
 * Global models thru `ac.models.expose()` upgraded from experimental to beta.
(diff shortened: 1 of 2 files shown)


isao:
@caridy if mojito-cli will ever be installed locally in an app (as opposed to globally) as "mojito", then we also need to change or remove the "bin": {"mojito": "bin/mojito"} in package.json as well.

caridy:
@isao yes, I don't think we will need the bin in both places once the cli gets ready, and we clean up mojito.

HISTORY.md
```diff
@@ -45,6 +45,9 @@ Here are details of the API changes.
 
 Please see the API docs for details of each.
 
+Installing mojito globally (i.e. `npm install --global mojito`) is deprecated. Please install mojito locally in your application.
```
caridy: I will suggest that we don't call it deprecated that fast. We can suggest that the change is intent to clear the space for the upcoming mojito-cli, but using the current global mojito will help with the archetypes and tests. I don't know, it sounds like to early to push them not to install it globally, when that's what they are used to.

caridy:
+1

PR #1032, opened 2013-03-13
cleanup how the cli is invoked a bit

isao:
specifically allow cli.js to be invoked with passthru args, for mojito-cli


isao:
passes functional

caridy:
+1

PR #1036, opened 2013-03-19
post-install check to post `npm i -g mojito` deprecation message

isao:
when `npm i -g mojito` is performed, this post install script outputs a deprecation message instructing user to `npm i -g mojito-cli` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     },
     "bugs": "https://github.com/yahoo/mojito/issues",
     "scripts": {
+        "postinstall": "./postinstall.sh",
         "test": "./tests/runall.sh"
     },
     "yahoo": {
--- a/postinstall.sh
+++ b/postinstall.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+npm=$(which npm ynpm | grep ^/ | head -1)
+global_prefix=$($npm prefix -g)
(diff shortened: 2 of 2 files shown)


caridy:
Hmm, the post-install step is not windows friendly I suspect. Aside from that, I see references to ynpm, which is internal, and should not be there.

Can we use a JS file instead? Probably using the `require('npm')` module to analyze the registry. Something like this:

```
require('npm').load(function (e, npm){
    npm.config.set('global', true);
    console.log(npm.commands.list(['mojito']));
});
```

that code will tell you whether or not mojito was installed globally, and where.

A couple of things to keep in mind:
- last time I tried `require('npm')` under ynpm, it failed, but Martin fixed it a couple of months ago.
- `require('npm')` should behave the same under npm and ynpm, it is seamless for the script.
- you will have to add `npm` under dependencies.

isao:
Good point, though mojito is not windows compatible now anyway.

I wanted to take the js and npm api approach, but thought npm as a dependency was too heavy for this.

isao:
note: this PR works for `npm i -g mojito` but not for `npm link` or `npm link path/to/mojito`. The latter two overwrite the mojito-cli "mojito" executable symlink

isao:
closing in favor of runtime messaging in cli.js, thanks @caridy

PR #1040, opened 2013-03-20
fix bz6160815: port argument must be an integer

isao:
sanitize command line argument for mojito start's port


lexander:
Looks good.

isao:
nice ’stash. :)

caridy:
+1

PR #1041, opened 2013-03-21
emit cli deprecations msg "pls npm i -g mojito-cli"

isao:
merge when old cli gets officially deprecated (see https://github.com/yahoo/mojito-cli/wiki)


caridy:
+1

maybe we should add a note about mojito-cli saying that all the commands will work the same or something like that.

isao:
yes, good point. maybe a short url to docs too.

holding off on this during "opt-in" period discussed today in mtg

PR #1044, opened 2013-03-21
refactor tunnel middleware into two phases

ekashida:
By splitting the tunnel logic into a detection phase and a handling phase,
applications are free to override the handling logic.

In order to maintain backwards-compatibility, instead of refactoring the
existing tunnel middleware, a tunnel-demux middleware to detect the type of
tunnel request, and tunnel-(rpc|specs|type) middleware to handle each type of
request have been added. Mojito itself will now use these new middleware in
place of the single tunnel middleware, which will remain for those applications
that are specifying it in their configuration.

### DIFF
--- a/lib/app/middleware/mojito-handler-tunnel-demux.js
+++ b/lib/app/middleware/mojito-handler-tunnel-demux.js
@@ -11,15 +11,6 @@
 var liburl  = require('url'),
     libpath = require('path');
 
-function trimSlash(str) {
-    if (str.charAt(0) === '/') {
-        str = str.substring(1, str.length);
-    }
-    if (str.charAt(str.length - 1) === '/') {
-        str = str.substring(0, str.length - 1);
-    }
-    return str;
-}
 
 
 /**
@@ -36,14 +27,15 @@ module.exports = function (config) {
     tunnelPrefix = appConfig.tunnelPrefix;
 
     if (staticPrefix) {
-        staticPrefix = '/' + trimSlash(staticPrefix);
+        staticPrefix = '/' + staticPrefix;
     }
     if (tunnelPrefix) {
-        tunnelPrefix = '/' + trimSlash(tunnelPrefix);
+        tunnelPrefix = '/' + tunnelPrefix;
     }
 
-    staticPrefix = staticPrefix || '/static';
-    tunnelPrefix = tunnelPrefix || '/tunnel';
+    // normalize() will squash multiple slashes into one slash.
+    staticPrefix = libpath.normalize(staticPrefix) || '/static';
+    tunnelPrefix = libpath.normalize(tunnelPrefix) || '/tunnel';
 
     return function (req, res, next) {
         var hasTunnelPrefix = req.url.indexOf(tunnelPrefix) === 0,
@@ -79,19 +71,17 @@ module.exports = function (config) {
 
         path = liburl.parse(req.url).pathname;
 
-        // Replace multiple slashes with a single one.
-        path = libpath.resolve(path);
-
         // Normalization step to handle `/{tunnelPrefix}`, `/{staticPrefix}`,
         // and `/{tunnelPrefix}/{staticPrefix}` URLs.
         path = path.replace(staticPrefix, '')
                    .replace(tunnelPrefix, '');
 
-        parts = path.split('/');
-
-        if (parts.length) {
+        if (path) {
             // Get the basename without the .json extension.
             name = libpath.basename(path, '.json');
+
+            // Time to get dirty.
+            parts = path.split('/');
 
             // Get the mojit type.
             type = parts[1];
--- a/lib/app/middleware/mojito-handler-tunnel-demux.js
+++ b/lib/app/middleware/mojito-handler-tunnel-demux.js
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
+ * Copyrights licensed under the New BSD License.
+ * See the accompanying LICENSE file for terms.
+ */
+
+/*global require, module*/
+/*jslint sloppy:true, nomen:true*/
+
+
+var liburl = require('url'),
+    RE_REPEATING_SLASH = /\/{2,}/g;
+
+function trimSlash(str) {
+    if (str.charAt(0) === '/') {
+        str = str.substring(1, str.length);
+    }
+    if (str.charAt(str.length - 1) === '/') {
+        str = str.substring(0, str.length - 1);
+    }
+    return str;
+}
+
+
+/**
+ * Export a function which can create the handler.
+ * @param {Object} config Data to configure the handler.
+ * @return {Object} The newly constructed handler.
+ */
+module.exports = function (config) {
+    var that      = this,
+        appConfig = config.store.getAppConfig({}) || {},
+        staticPrefix,
+        tunnelPrefix;
+
+    staticPrefix = appConfig.staticHandling && appConfig.staticHandling.prefix;
+    tunnelPrefix = appConfig.tunnelPrefix;
+
+    if (staticPrefix) {
+        staticPrefix = '/' + trimSlash(staticPrefix);
+    }
+    if (tunnelPrefix) {
+        tunnelPrefix = '/' + trimSlash(tunnelPrefix);
+    }
+
+    this.staticPrefix = staticPrefix || '/static';
+    this.tunnelPrefix = tunnelPrefix || '/tunnel';
+
+    return function (req, res, next) {
+        var hasTunnelPrefix = req.url.indexOf(that.tunnelPrefix) === 0,
+            hasTunnelHeader = req.headers['x-mojito-header'] === 'tunnel',
+            name,
+            type,
+            url,
+            parts;
+
+        // If we are not tunneling get out of here fast!
+        if (!hasTunnelPrefix && !hasTunnelHeader) {
+            return next();
+        }
+
+        req._tunnel = {};
+
+        /**
+        Tunnel examples
+
+        RPC tunnel:
+        /tunnel (or it could just have the tunnel header)
+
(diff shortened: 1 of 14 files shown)


ekashida:
Still working on this (tests, docs, end-to-end testing, etc). Would like some feedback for what I have so far!

lib/app/middleware/mojito-handler-tunnel-specs.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Exports a middleware factory that can handle spec tunnel requests.
+ *
+ * @param {Object} config The configuration.
+ * @return {Function} The handler.
+ */
+module.exports = function (config) {
+    return function (req, res, next) {
+        var specsReq = req._tunnel && req._tunnel.specsReq,
+            instance = {},
```
caridy: do not create objects if they are not going to be used. move this below the `if (!specsReq)`, otherwise you will be creating an empty object per req even when it is not a tunnel request, in which case this will have to be collected by GC.

lib/app/middleware/mojito-handler-tunnel-specs.js
```diff
@@ -0,0 +1,66 @@
...
+            type,
+            name;
+
+        if (!specsReq) {
+            return next();
+        }
+
+        type = specsReq.type;
+        name = specsReq.name;
+        name = name && name.split('.').slice(0, -1).join('.');
```
caridy: isn't easier to use require('path.extname(), then slicing based on that? this line seems complex.

lib/app/middleware/mojito-handler-tunnel-demux.js
```diff
@@ -0,0 +1,120 @@
...
+        /{tunnelPrefix}/{type}/definition.json // custom prefix
+        /tunnel/static/{type}/definition.json  // according to a UT
+
+        Spec tunnel:
+        /static/{type}/specs/default.json
+        /{staticPrefix}/{type}/specs/default.json  // custom prefix
+        /tunnel/static/{type}/specs/default.json   // according to a UT
+        **/
+
+        url = req.url.split('?')[0];
```
caridy: hmm, I guess we should use `require('url').parse()` to really understand that url.

lib/app/middleware/mojito-handler-tunnel-specs.js
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!type || !name) {
+            return sendError(res, 'Not found: ' + req.url, 500);
+        }
+
+        instance.base = type;
```
caridy: here is where you use `instance` for the first time, here is where you should create the object with base on it.

lib/app/middleware/mojito-handler-tunnel-specs.js
```diff
@@ -0,0 +1,66 @@
...
+        if (!specsReq) {
+            return next();
+        }
+
+        type = specsReq.type;
+        name = specsReq.name;
+        name = name && name.split('.').slice(0, -1).join('.');
+
+        if (!type || !name) {
+            return sendError(res, 'Not found: ' + req.url, 500);
```
caridy: If it is not found, it should be 404, if it is an invalid url, then 500 is just fine.

lib/mojito.js
```diff
@@ -90,7 +90,10 @@ MojitoServer.MOJITO_MIDDLEWARE = [
     'mojito-parser-body',
     'mojito-parser-cookies',
     'mojito-contextualizer',
-    'mojito-handler-tunnel',
```
caridy, line 93: this will be problematic for BC. Can we preserve the tunnel as a sugar for all the new middleware? middlewares within a middleware is a common practice, and once we refactor mojito middleware infrastructure, in which case we know we will break BC, we can remove the sugar.

lib/app/middleware/mojito-handler-tunnel-type.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Exports a middleware factory that can handle type tunnel requests.
+ *
+ * @param {Object} config The configuration.
+ * @return {Function} The handler.
+ */
+module.exports = function (config) {
+    return function (req, res, next) {
+        var typeReq  = req._tunnel && req._tunnel.typeReq,
+            instance = {};
```
caridy: same as above.

lib/app/middleware/mojito-handler-tunnel-type.js
```diff
@@ -0,0 +1,56 @@
...
+
+        if (!typeReq) {
+            return next();
+        }
+
+        if (!typeReq.type) {
+            return sendError(res, 'Not found: ' + req.url, 500);
```
caridy: same as above

lib/app/middleware/mojito-handler-tunnel-rpc.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {Object} config The configuration.
+ * @return {Function} The handler.
+ */
+module.exports = function (config) {
+    return function (req, res, next) {
+        var command  = req.body,
+            instance = command.instance;
+
+        command.context = command.context || {};
```
caridy: move logic after the `if`

caridy:
done with the first pass @ekashida

lib/app/middleware/mojito-handler-tunnel-specs.js
ekashida: agreed

lib/app/middleware/mojito-handler-tunnel-specs.js
ekashida: I don't think it was simplified much, but the comment definitely helps :)

```
// Remove the extension name.
ext  = libpath.extname(name);                                                  
name = name && ext && name.slice(0, -ext.length);
```

ekashida: I got rid of this logic completely by passing in (from demux) just the basename without the extension:

```
// Get the basename without the .json extension.                               
name = libpath.basename(path, '.json');
```

lib/app/middleware/mojito-handler-tunnel-demux.js
ekashida: agreed

lib/app/middleware/mojito-handler-tunnel-specs.js
ekashida: agreed

lib/app/middleware/mojito-handler-tunnel-specs.js
ekashida: Damn you copy/paste!

ekashida:
@caridy I think I've addressed all your feedback, although I'm still getting those weird unit test errors. I'll keep looking at them...

lib/app/middleware/mojito-handler-tunnel-rpc.js
```diff
@@ -0,0 +1,58 @@
...
+        command.context.runtime = 'server';
+
+        // All we need to do is expand the instance given within the RPC call
+        // and attach it within a "tunnelCommand", which will be handled by
+        // Mojito instead of looking up a route for it.
+        config.store.expandInstance(
+            command.instance,
+            command.context,
+            function (err, instance) {
+                // Replace with the expanded instance.
```
caridy, line 43: If `err` is detected, we should call next(err) at least.

lib/app/middleware/mojito-handler-tunnel-parser.js
```diff
@@ -0,0 +1,108 @@
...
+
+            // Time to get dirty.
+            parts = path.split('/');
+
+            // Get the mojit type.
+            type = parts[1];
+
+            // "Spec" tunnel request
+            if (parts[parts.length - 2] === 'specs') {
+                req._tunnel.specsReq = {
```
caridy, line 91: this is a bold assumption, what is req._tunnel is not set? Why is this middleware coupled with others?

lib/app/middleware/mojito-handler-tunnel-specs.js
```diff
@@ -0,0 +1,58 @@
...
+
+        if (!specsReq) {
+            return next();
+        }
+
+        type = specsReq.type;
+        name = specsReq.name;
+
+        if (!type || !name) {
+            return req._tunnel.sendError(res, 'Not found: ' + req.url, 404);
```
caridy: same as above, bold assumption. I think we should call `next(new Error())` instead.

lib/app/middleware/mojito-handler-tunnel-type.js
```diff
@@ -0,0 +1,49 @@
...
+            type: typeReq.type
+        };
+
+        config.store.expandInstanceForEnv(
+            'client',
+            instance,
+            req.context,
+            function (err, data) {
+                if (err) {
+                    return req._tunnel.sendError(
```
caridy: same as above.

lib/app/middleware/mojito-handler-tunnel-type.js
```diff
@@ -0,0 +1,49 @@
...
+    return function (req, res, next) {
+        var typeReq = req._tunnel && req._tunnel.typeReq,
+            instance;
+
+        if (!typeReq) {
+            return next();
+        }
+
+        if (!typeReq.type) {
+            return req._tunnel.sendError(res, 'Not found: ' + req.url, 404);
```
caridy: same as above.

lib/app/middleware/mojito-handler-tunnel.js
```diff
@@ -4,202 +4,57 @@
...
+    return function (req, res, next) {
+        var middleware = [
+            parser,
+            rpc,
+            specs,
+            type
+        ];
+
+        // Helper methods.
+        req._tunnel = {
```
caridy: If this is not used in this middleware, it should not be defined here.

PR #1047, opened 2013-03-26
enable per-request tunnel url configuration

ekashida:
Configured `tunnelUrl` must still be prefixed with `tunnelPrefix`.

### DIFF
--- a/lib/app/autoload/tunnel-client.common.js
+++ b/lib/app/autoload/tunnel-client.common.js
@@ -25,6 +25,11 @@ YUI.add('mojito-tunnel-client', function(Y, NAME) {
 
             url = this._appConfig.tunnelPrefix;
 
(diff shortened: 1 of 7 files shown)


lib/app/autoload/tunnel-client.common.js
```diff
@@ -25,6 +25,11 @@ YUI.add('mojito-tunnel-client', function(Y, NAME) {
 
             url = this._appConfig.tunnelPrefix;
 
+            if (command._tunnelUrl) {
+                url = command._tunnelUrl;
+                command._tunnelUrl = null;
```
caridy: better to set it to undefined, which will skip the `JSON.stringify` process.

caridy:
done with the review!

caridy:
+1

PR #1051, opened 2013-03-28
reusing the same staticAppConfig for the request instead of for the mojit

dmorilha:
Hi,

would that something mojito team might consider? it could live behind a staticAppConfig and be disable by default. When dealing with expressive amount of config and tens of mojits this change can save more than 10% on performance.

The change intends to use a copy of staticAppConfig per request instead of per mojit dispatch. Tests with our application looked OK.

Please comment.

### DIFF
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -308,7 +308,7 @@ YUI.add('mojito-action-context', function(Y, NAME) {
         this._adapter = opts.adapter;
 
         // pathToRoot, viewEngine, amoung others will be available through this.
-        this.staticAppConfig = store.getStaticAppConfig();
+        this.staticAppConfig = opts.staticAppConfig || store.getStaticAppConfig();
 
         // Create a function which will properly delegate to the dispatcher to
         // perform the actual processing.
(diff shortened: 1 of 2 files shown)


caridy:
Interesting, you're using `ac._dispatch` as a page object to store info per request, I like it!. There are a couple of things to keep in mind though:
- The only reason this works is because `ac._dispatch` is used in composite (and in your scheduler) to reuse the dispatch instance created by `mojito.js` on every request, but that do not have to be the case all the time.
- By avoiding copying that static app config per mojit instance in a request we cross the second bridge, which means that object can be changed by one of the mojits affecting the other mojits in the same request dispatch process, which considering that we crossed the first bridge when we decided to drop the isolation between mojits, which is ok.

If we clearly state that this is the natural behavior in the docs, potentially introducing the page object to formalize this instead of just relying on `ac._dispatch` (attaching that page object into req under the hood), I will give this a +1.

dmorilha:
Wouldn't that be aligned with the idea of circles someone's in mojito's team proposed a while back?

It would be nice to have this as a configurable trade-off into the apps configuration. If that's too much of complexity and could turn to be the natural behavior, it is even better!

drewfish:
Closing this in favor of PR #1056 (merged), which implemented this idea a little differently.

drewfish:
Since PR #1056 is merged an passes CI already, here are the perf improvements:
search-perf2

PR #1054, opened 2013-04-02
add mojito configuration schemas

Yujun-Qian:
This pull request adds mojito configuration schemas. These schemas are in compliance with IETF json schema draft, and are defined according to http://developer.yahoo.com/cocktails/mojito/docs/intro/mojito_configuring.html. The related code change has been sent to mojito-cli repository.


caridy:
I will argue that the schemas should not be part of mojito source, but part of mojito-cli-validate or something similar but not core. We can also explore a similar approach to the one we have lineup for the initial archetypes that can bring some default npm devDependencies that can be used by cli to resolve few other things.

Yujun-Qian:
The configuration of mojito keeps evolving, accordingly these schemas will change with each version of mojito. I think it makes sense for the shemas to be part of mojito, and the validate command to be part of mojito-cli, so that the user can use a single "mojito validate" command to validate all his mojito applications.

isao:
We asked @Yujun-Qian to have the schemas included with core mojito source, which makes the most sense at this point-- especially if we think it will be changed.

I also think we must validate our test apps against the mojito schemas in CI. +1

PR #1055, opened 2013-04-02
fix issue 1053 non-string args to path.join in 0.10.x

isao:


lib/app/autoload/store.server.js
```diff
@@ -1692,11 +1692,11 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
 
             switch (info.pkg.yahoo.mojito.type) {
             case 'bundle':
-                dir = this._libs.path.join(info.dir, info.pkg.yahoo.mojito.location);
+                dir = this._libs.path.join(info.dir, info.pkg.yahoo.mojito.location || '');
```
caridy, line 1695: when this location member can be undefined?

lib/app/autoload/store.server.js
```diff
@@ -709,7 +709,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 if (res.type === 'view') {
                     template = {
                         'content-path': (env === 'client' ?
-                                this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url) :
+                                this._libs.path.join(this._appConfigStatic.pathToRoot || '', res.url || '') :
```
caridy: when the res.url will be undefined/null?

lib/app/autoload/store.server.js
isao, line 1695: when a user has a bundle without a location property. It's defined [here](https://github.com/yahoo/mojito/blob/develop/tests/fixtures/packages/node_modules/a/package.json) but not [here](https://github.com/yahoo/mojito/blob/develop/tests/fixtures/packages/node_modules/b/package.json) for example.

lib/app/autoload/store.server.js
isao: during [this test](https://github.com/yahoo/mojito/blob/develop/tests/unit/lib/app/autoload/test-store.server.js#L500) for [this fixture](https://github.com/yahoo/mojito/tree/develop/tests/fixtures/packages).

drewfish: I still think adding the real addon-rs-url in the descriptor is a cleaner fix (if it works).

caridy: me too!

isao: So res.url is always defined? I thought it wasn't in some cases on the client i.e. shared mojits. If no res.url is valid, then 'content-path' should not be defined with a unusable value. If the test exposes a case where res.url should be defined but isn't, then I think we should open a separate issue.

caridy: Yes, we remove that exception a while ago, and we expect that url is always defined.

lib/app/autoload/store.server.js
caridy, line 1695: So, location is not produced by the store? it just comes from the pkg information as optional then? if that's the case, we are good!

isao:
thanks for the feedback @caridy @drewfish. the root of the undefined res.url issue was test mock, not store code. unit tests pass on 0.10.2 mac.

drewfish:
+1

PR #1056, opened 2013-04-02
use page object to store/use staticAppConfig

drewfish:
Reworking of PR #1051 to hang staticAppConfig off of "page" object (instead of dispatcher).

### DIFF
--- a/lib/app/autoload/mojito-client.client.js
+++ b/lib/app/autoload/mojito-client.client.js
@@ -338,6 +338,9 @@ YUI.add('mojito-client', function(Y, NAME) {
                     // pass globalHookhandler to addons that may want to use hooks
                     globalHookHandler: globalHookHandler
                 };
+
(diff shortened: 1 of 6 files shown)


lib/app/autoload/mojito-client.client.js
```diff
@@ -338,6 +338,9 @@ YUI.add('mojito-client', function(Y, NAME) {
                     // pass globalHookhandler to addons that may want to use hooks
                     globalHookHandler: globalHookHandler
                 };
+
+            this.page.staticAppConfig = config.appConfig;
```
caridy: appConfig is a local variable pointing to this.appConfig.

drewfish: D'oh.. missed that, will use that instead.

caridy:
Lets add a couple of tests and merge this.

+1

PR #1062, opened 2013-04-05
fixes the issue with Y.config.global after the upgrade to yui@3.9.1

caridy:
With this change, the app itself will use the pure version of YUI thru `require('yui')` while other internal components that should be running in isolation should be using the infamous `yui-sandbox` to make sure that the `Y` instance for the app does not get corrupted. This will guarantee that yui upgrades will hardly impact any app but only our internal components if YUI changes the way it detects features when running in node.js.


drewfish:
There are two other users of yui-sandbox.js.  Should they be removed too?

drewfish:
+1

caridy:
@drewfish yes, store uses `yui-sandbox` in multiple places, and since it should runs in isolation, using sandbox is the right thing to do for those routines.

PR #1075, opened 2013-04-10
switch to latest arrow

lzhan:
All mojito unit tests and functional tests passed. 
But see the following error for almost all the functional tests and some unit tests:

Failed to parse javascript(YUI) file: xxx.js
Property 'YUI' of object #<Object> is not a function error

It did not affect test execution and test result. But still investigating.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
         "async": "*",
         "commander": "1.0.1",
         "mockery": "~1.4.0",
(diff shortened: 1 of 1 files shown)


drewfish:
If everything works out, +1.

caridy:
It seems like it does not work :(

isao:
+1

PR #1085, opened 2013-04-24
add unit tests for lib/app/autoload/store.client.js

lzhan:


caridy:
-1 I don't like the `_mockLib` approach.

lzhan:
@caridy Any suggestion? I got the idea from the unit test of store.server.js

caridy:
The reason why you introduced _mockLib because you want to mock the internal method called `retrieveFile`. I think, either you expose that method, along with the other 3 private methods, or you mock the things that the method is trying to use. Which in this case is `Y.io` I think.

lzhan:
@caridy I see your point. Let me try. Thanks.

tests/unit/lib/app/autoload/test-store.client.js
```diff
@@ -61,6 +77,136 @@ YUI().use('mojito-client-store', 'test', 'querystring-stringify-simple', functio
...
+                valid = this.store._validateInstance(invalidbase);
+            A.areEqual(false, valid);
+        },
+
+        'test expandInstanceForEnv1': function() {
+            var instance = {
+                type: 'test_mojit_1',
+                config: {testKey4: 'other'}
+            };
+            try {
```
caridy, line 121: You should not have try/catch in your tests, instead you can use the provision for failures, here is an example:

```
    "test expandInstanceForEnv1": function () {
        var instance = {
           type: 'test_mojit_1',
           config: {testKey4: 'other'}
        };
        A.throwsError('<expected error message goes here.....>', function () {
            // this function gets executed immediately...
            this.store.expandInstanceForEnv("client", instance, {}, function(err, base) {});
        });
    }
```

more details here:

http://yuilibrary.com/yui/docs/api/classes/Test.Assert.html#method_throwsError

tests/unit/lib/app/autoload/test-store.client.js
```diff
@@ -61,6 +77,136 @@ YUI().use('mojito-client-store', 'test', 'querystring-stringify-simple', functio
...
+            this.store.expandInstanceForEnv("client", instance, {}, function(err, base) {
+                A.areEqual("test_mojit_2", base.type);
+                A.areEqual("test_mojit_1", base.base);
+                A.areEqual("other", base.config.testKey4);
+            });
+        },
+
+        'test expandInstanceForEnv3': function() {
+            var context = this.store.getStaticContext();
+            try {
```
caridy: same as above!

tests/unit/lib/app/autoload/test-store.client.js
```diff
@@ -61,6 +77,136 @@ YUI().use('mojito-client-store', 'test', 'querystring-stringify-simple', functio
...
+
+        'test get type': function() {
+            var context = this.store.getStaticContext();
+            Y.io = function(url, config) {
+                var id = "newid",
+                    obj = {};
+                obj.responseText = '{ "status": 200 }';
+                config.on.complete('myid', obj);
+            };
+            try {
```
caridy, line 188: same as above!

tests/unit/lib/app/autoload/test-store.client.js
```diff
@@ -61,6 +77,136 @@ YUI().use('mojito-client-store', 'test', 'querystring-stringify-simple', functio
...
+        'test get spec': function() {
+            var context = this.store.getStaticContext();
+
+            Y.io = function(url, config) {
+                var id = "newid",
+                    obj = {};
+                obj.responseText = '{ "status": 200 }';
+                config.on.complete('myid', obj);
+            };
+            try {
```
caridy, line 204: same as above!

caridy:
@lzhan once you remove the try/catch, feel free to merge.

+1

lzhan:
A.throwsError() can be used for test case expecting error. I will update test "expandInstanceForEnv3"
A.throwsError() can not be used for the other three cases which are not expecting error(the tests will fail if there is an error/exception).

PR #1088, opened 2013-04-26
simplify middleware registration; dependency on mojito-contextualizer

imalberto:
=== DO NOT MERGE YET ===
- Simplified middleware API registration. Documented changes in HISTORY.md
- Removed mojito-contextualizer middleware, and instead add it as an external dependency
- package.json: updated version to 0.6.0, updated YUI to ~3.10.0, added external dependency and added myself to maintainers list
## Checklist
- [ ] Publish mojito-contextualizer to npm
- [ ] Merge this PR with mojito#next


imalberto:
Note: Travis build is failing because mojito-contextualizer package dependency is not yet published to npm.

drewfish:
I don't think we should merge this with mojito#develop.  I think it's fine to have it live _just_ in mojito#next.

imalberto:
@drewfish Good catch (typo). This PR will be merged with `mojito#next`. Updated PR description.

HISTORY.md
```diff
@@ -19,6 +19,85 @@ Bug Fixes
...
+            mojito-foo.js
+            bar.js
+        mojits/
+        application.json
+        routes.json
+        server.js
+```
+
+  The following changes are required:
+  - rename `mojito-foo.js` to `foo.js`
```
drewfish: I don't think the user needs to rename their `mojito-foo.js` middleware to `foo.js`, since we're simply stopping treating the `mojito-` prefix as special.

lib/mojito.js
```diff
@@ -212,35 +213,46 @@ MojitoServer.prototype._useMiddleware = function (app, dispatcher, midDir, midCo
         midBase,
         midFactory;
 
+    // TODO: document middleware API for app developers
```
drewfish: It would be nice if we could get this middleware API documentation as part of this PR.

lib/mojito.js
```diff
@@ -212,35 +213,46 @@ MojitoServer.prototype._useMiddleware = function (app, dispatcher, midDir, midCo
...
-                //console.log("======== MIDDLEWARE mojito -- " +
-                //    "builtin mojito-handler-dispatcher");
                 app.use(dispatcher);
             } else {
-                midPath = libpath.join(__dirname, 'app', 'middleware', midName);
-                //console.log("======== MIDDLEWARE mojito " + midPath);
-                midFactory = require(midPath);
+
+                // Mojito can now register middleware by module name, or by
+                // paths.
```
drewfish: The resource store is keeping track of _all_ middleware in all the NPM modules (well, those that are mojito bundles).   It might be nice to use the store to find these middleware, so that they don't necessarily have to be in the app themselves.  (An example might be a `search-mojito-contextualizer` which is a couple modules deep in the NPM hierarchy.  The user should still be able to specify `search-mojito-contextualizer` in the `middleware: [...]` in `application.json`.)

imalberto:
Closing this PR after discussion with team.

PR #1092, opened 2013-05-03
routes sent to the client should honor the client context

caridy:
- server was using the server context to produce the routes to be expose to the client.
- server was also compiling those routes before sending them, which is something we can perfectly do in the client side.
- other small clean up in the data we serialized for client.


drewfish:
+1

PR #1093, opened 2013-05-03
Correction in Documentation: Composite Mojits

lazybean:
Hi, 
Thanks for the great cocktail and the nice doc.

In the developer topics, composite Mojits, "Executing Child Mojits" section, first code exemple, the name of the method should be 'init' and not 'index', as said in the text just before.


zhouyaoji:
Thanks lazybean for the PR, but the method should be 'index', not 'init'. The 'init' method for controllers was deprecated a while ago, although it still exists for models.

caridy:
@lazybean it is actually the other way around, init was deprecated a while ago, and the actual action name is "index".

/ping @zhouyaoji

lazybean:
Thanks for the fast feedback.
I was actually wondering what was this init method, since I haven't used in any controller.

Then, should we update the description to fit the code? The 'init' method is mentioned in the text and in another code example (the 3rd).
At least should we  add an ac.done() in the index method?

zhouyaoji:
Thanks for catching that, lazybean. I thought I had removed all references to the 'init' method in controllers when we deprecated the 'init' method in controllers, but I obviously missed one. I also grepped the rest of the doc source files and found only references to 'init' in binder code, which is valid. I also added 'ac.done' to the 'index' method as you suggested. Thanks again for your help!

lazybean:
Looks great now, thanks a lot!

PR #1094, opened 2013-05-03
backport bugfix for build cmd: spec in mapSpecUris referred to key not value

isao:
see isao/mojito-build@c11c48100828397e3f70dd3e20477a40fbf809f6

This is low impact, the affected code supports an undocumented legacy edge-case from the Livestand days.


caridy:
+1

PR #1095, opened 2013-05-04
added doT templates view engine

spsDrop:
I've added doT templates as an alternative view engine. It's MIT license, dead simple, SUPER fast, flexible, and in this implementation more fault tolerant than HB. This implementation has support for partials. I haven't made helpers work because they aren't in doT. It would be easy to add a helpers object to the data object going into the compiler that are just references to functions to be used in the template.

I've been using it on one major project and would really like to have dot available on other smaller projects as part of the core mojito lib.

More on doT.
http://olado.github.io/doT/index.html

Benchmarks:

doT is ~46x faster

Chrome 26.0.1386
Handlebars.js:
63,263 ops/sec

Chrome 26.0.1386 (2)
doT.js:
2,943,894 ops/sec

### DIFF
--- a/lib/app/addons/view-engines/doT-base.common.js
+++ b/lib/app/addons/view-engines/doT-base.common.js
@@ -1,0 +1,128 @@
+// doT.js
+// 2011, Laura Doktorova, https://github.com/olado/doT
+// Licensed under the MIT license.
+// 'Ported' to YUI by Jesse Foltz
+
+YUI.add("doT", function(Y) {
+    "use strict";
+
+    var doT = Y.doT = {
+		version: '1.0.0',
+		templateSettings: {
+			evaluate:    /\{\{([\s\S]+?\}?)\}\}/g,
+			interpolate: /\{\{=([\s\S]+?)\}\}/g,
+			encode:      /\{\{!([\s\S]+?)\}\}/g,
+			use:         /\{\{#([\s\S]+?)\}\}/g,
+			useParams:   /(^|[^\w$])def(?:\.|\[[\'\"])([\w$\.]+)(?:[\'\"]\])?\s*\:\s*([\w$\.]+|\"[^\"]+\"|\'[^\']+\'|\{[^\}]+\})/g,
+			define:      /\{\{##\s*([\w\.$]+)\s*(\:|=)([\s\S]+?)#\}\}/g,
+			defineParams:/^\s*([\w$]+):([\s\S]+)/,
+			conditional: /\{\{\?(\?)?\s*([\s\S]*?)\s*\}\}/g,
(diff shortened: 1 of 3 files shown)


drewfish:
Hi @soljin,

You should be able to add doT support to mojito _without_ baking it into the mojito package itself.  Mojito supports "extension" packages which can add new functionality.  (This functionality can be mojits, or other things such as view engines like you're suggesting.)  There are some details here:
http://developer.yahoo.com/cocktails/mojito/docs/topics/mojito_npm.html#authoring-an-npm-module-containing-shared-mojito-resources

This approach allows you to publish a new package to NPM (perhaps called `mojito-dot`) to add this functionality to mojito.  Users of mojito who want doT support would then also depend on this package.  This allows you to maintain this package (bug fixes, new features, etc) with its own release cycle, not tied to mojito's release cycle.

Feel free to email me with any questions you might have about how to do this.
Drew

caridy:
@soljin as a side note, we are moving on the direction @drewfish mentioned, and starting on 0.6.x we will have more and more small packages to do very little. In the case of the view engines, the packages will be locator plugins, and will be responsible for compiling views and rendering compiled views. For now, you can reserve the package name as drew suggested `mojito-dot`.

Btw, those perfs numbers are impressive, looking forward to try that out.

spsDrop:
https://npmjs.org/package/mojito-dot

Well that was easy! Thanks for the tip didn't know about the npm publishing stuff very slick and simple. As I have come to expect from the mojito crew. Great work! Thanks for the advice. Try out doT templates they are very fun.

caridy:
@soljin great work, you're officially the first person to create a community extension for mojito, thanks a lot!, we expect more on this front when we start promoting the capabilities of extending mojito core functionalities.

About the component, a couple of notes:
- Please, add the step-by-step guide, including the fact that they have to install the package in the app directory using `npm i mojito-dot --save`, etc.
- In the code, I see you're using `Y.namespace('Env.Mojito.Handlebars')` as the internal cache mechanism, rename it to use `Y.namespace('Env.Mojito.dot')` so you don't run into any conflicts.
- You might want to write some unit tests on this.
- You might want to write some integration tests on this to make sure that it works with every new mojito release (we might be able to help here).

Again, great job!

spsDrop:
@caridy YaY, thanks a lot!

I didn't have to save the package. I just added it to my package.json dependencies and ran npm install, like I would for anything. And mojito loaded it up just fine. I will flesh out the docs and tests, probably this weekend.

I am going to remove mojito as one of it's dependencies since it's not really required and ends up installing a whole extra version of mojito that is ignored. I'll make it a dev dependency when I write some tests.

Thanks for all the help!

PR #1098, opened 2013-05-06
generalize tests to run with phantomjs

lzhan:


isao:
not sure how it relates to phantomjs but +1

lzhan:
Thanks. The header format on phantomjs is a little different. So searching for a string is a little different too.

PR #1101, opened 2013-05-07
Review 'app.js' proposal for discussion

imalberto:
This PR proposes some ideas to have `app.js` closely resemble what a regular `express` app should look like, and the changes required in mojito to support that paradigm shift.

CHANGELOG:
- Updated all built-in middleware to lazy init
- Moved dispatcher into its own middleware module `middleware/mojito-handler-dispatcher.js`
- Moved stateless functionality into their own module (not related to core mojito): `middleware.js` and `logger.js`
- Updated the `example/newsboxes` and `examples/quickstartguide` examples with their own `app.js` to showcase the use case. These apps can be started with `node app`.
- Fixed middleware related unit tests
- Disabled lib/mojito.js unit tests (for now)

UNIT TESTS RESULTS:

```
********************************************
Consolidated Test Report Summary
********************************************
Total Number of Executed Tests  :633
Total Number of Passed Tests  : 621
Total Number of Failed Tests  : 0
Total Number of Skipped Tests  : 12
********************************************

Total Time of Execution :69.46 seconds
```

### DIFF
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -1,0 +1,37 @@
+
+
+/*jslint node:true*/
+
+'use strict';
+
+var debug = require('debug')('app:yaml'),
+    express = require('express'),
+    mojito = require('../../'),
+    app;
+
+app = express();
+
+// register custom middleware here
+// app.use(require('./middleware/foo.js'));
+
+// Registers Mojito's default
+app.use(app.mojito.registerMiddleware());
+/*
+// By doing this, give users more control over which order middleware
+// are registered.
+app.mojito.middleware().forEach(function (mid) {
+    debug('app.use(): ' + mid);
+    app.use(app.mojito[mid]);
+});
+*/
+
+// In addition to mojito `routes.json`, user can hook up additional 
+// mounting points if necessary.
+app.get('/status', function (req, res) {
+    res.send('200 OK');
+});
+
+app.listen(app.get('port'), function () {
+    debug('Server listening on port ' + app.get('port') + ' ' +
+               'in ' + app.get('env') + ' mode');
+});
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -15,7 +15,7 @@ app = express();
 // app.use(require('./middleware/foo.js'));
 
 // Registers Mojito's default
-app.use(mojito.middleware());
+app.use(mojito.middleware(app));
 
 /*
 // By doing this, give users more control over which order middleware
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -15,13 +15,16 @@ app = express();
 // app.use(require('./middleware/foo.js'));
 
 // Registers Mojito's default
-app.use(app.mojito.registerMiddleware());
+app.use(mojito.middleware());
+
 /*
 // By doing this, give users more control over which order middleware
 // are registered.
-app.mojito.middleware().forEach(function (mid) {
+// But the recommended way to register all default middleware is:
+// `app.use(mojito.middleware());`
+mojito.defaultMiddleware().forEach(function (mid) {
     debug('app.use(): ' + mid);
-    app.use(app.mojito[mid]);
+    app.use(mojito[mid]);
 });
 */
 
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -16,7 +16,7 @@ var debug = require('debug')('app'),
 
 app = express();
 
-app.use(mojito.middleware(app));
+app.use(mojito.middleware());
 
 app.get('/status', function (req, res) {
     res.send('200 OK');
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -1,35 +1,23 @@
+/*
+* Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
+* Copyrights licensed under the New BSD License.
+* See the accompanying LICENSE file for terms.
+*/
 
 
 /*jslint node:true*/
 
 'use strict';
 
-var debug = require('debug')('app:yaml'),
+var debug = require('debug')('app'),
     express = require('express'),
     mojito = require('../../'),
     app;
 
 app = express();
 
-// register custom middleware here
-// app.use(require('./middleware/foo.js'));
-
-// Registers Mojito's default
 app.use(mojito.middleware(app));
 
-/*
-// By doing this, give users more control over which order middleware
-// are registered.
-// But the recommended way to register all default middleware is:
-// `app.use(mojito.middleware());`
-mojito.defaultMiddleware().forEach(function (mid) {
-    debug('app.use(): ' + mid);
-    app.use(mojito[mid]);
-});
-*/
-
-// In addition to mojito `routes.json`, user can hook up additional 
-// mounting points if necessary.
 app.get('/status', function (req, res) {
     res.send('200 OK');
 });
--- a/lib/app/middleware/mojito-contextualizer.js
+++ b/lib/app/middleware/mojito-contextualizer.js
@@ -58,14 +58,16 @@ RequestContextualizer.prototype = {
     handle: function () {
 
         var my = this,
+            mojito,
             defaultLang;
 
         return function (req, res, next) {
 
 
             if (!defaultLang && req.app && req.app.mojito) {
-                defaultLang = (req.app.get('mojito.context') &&
-                    req.app.get('mojito.context').lang) || DEFAULT_LANG;
+                mojito = req.app.mojito;
+                defaultLang = (mojito.context && mojito.context.lang) ||
+                                DEFAULT_LANG;
             }
 
             var query = url.parse(req.url, true).query || {};
--- a/lib/app/middleware/mojito-contextualizer.js
+++ b/lib/app/middleware/mojito-contextualizer.js
@@ -180,8 +180,10 @@ RequestContextualizer.prototype = {
 };
 
 
+var rc = new RequestContextualizer();
+
 /**
 @type {Function} express middleware
 **/
-exports = module.exports = new RequestContextualizer().handle;
+exports = module.exports = rc.handle.bind(rc);
 
--- a/lib/app/middleware/mojito-contextualizer.js
+++ b/lib/app/middleware/mojito-contextualizer.js
@@ -7,8 +7,14 @@
 
 /*jslint node:true, nomen:true*/
 
+/**
+@module mojito-contextualizer
+**/
 
-var debug = require('debug')('middleware:contextualizer'),
+'use strict';
+
+var debug = require('debug')('mojito:middleware:contextualizer'),
+    // TODO: review qs vs querystring performance
     qs = require('querystring'),
     url = require('url'),
     CONTEXT_SERVER = 'server',
@@ -35,15 +41,20 @@ var debug = require('debug')('middleware:contextualizer'),
     REGEX_LANGUAGE_MATCH = /^([a-z]+)-([a-z]+)$/;
 
 /**
- * The request contextualizer. Middleware which adds context to a request.
- * @constructor
- */
+The request contextualizer. Middleware which adds context to a request.
+@constructor
+**/
 function RequestContextualizer() {
 }
 
 
 RequestContextualizer.prototype = {
 
+    /**
+    @method handle
+    @public
+    @return {Function} express middleware
+    **/
     handle: function () {
 
         var my = this,
@@ -86,6 +97,10 @@ RequestContextualizer.prototype = {
         };
     },
 
+    /**
+    @method _device
+    @protected
+    **/
     _device: function (ua, def) {
         // debug('detecting device from UA: ' + ua);
 
@@ -119,6 +134,12 @@ RequestContextualizer.prototype = {
         return def;
     },
 
+    /**
+    @method _language
+    @protected
+    @param {String} al value of the Accept-Language header
+    @param {String} def default language
+    **/
     _language: function (al, def) {
 
         al = (al || '').trim();
@@ -147,6 +168,12 @@ RequestContextualizer.prototype = {
         return chosen;
     },
 
+    /**
+    @method _preferredLanguages
+    @protected
+    @param {String} lang
+    @param {String} defaultLang
+    **/
     _preferredLanguages: function (lang, defaultLang) {
         return [lang, defaultLang].join(',');
     }
@@ -154,11 +181,7 @@ RequestContextualizer.prototype = {
 
 
 /**
- * Export a function capable of constructing a new contextualizer.
- * @param {Object} config Data to configure the new contextualizer.
- * @return {Object} The contextualizer.
- */
-exports = module.exports = function (config) {
-    var m = new RequestContextualizer();
-    return m.handle();
-};
+@type {Function} express middleware
+**/
+exports = module.exports = new RequestContextualizer().handle;
+
--- a/lib/app/middleware/mojito-contextualizer.js
+++ b/lib/app/middleware/mojito-contextualizer.js
@@ -1,15 +1,15 @@
-/*
+/**
  * Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
  * Copyrights licensed under the New BSD License.
  * See the accompanying LICENSE file for terms.
  */
 
 
-/*jslint anon:true, sloppy:true, nomen:true*/
+/*jslint node:true, nomen:true*/
 
 
-var qs = require('querystring'),
-    logger,
+var debug = require('debug')('middleware:contextualizer'),
+    qs = require('querystring'),
     url = require('url'),
     CONTEXT_SERVER = 'server',
     DEFAULT_LANG = 'en',
@@ -44,14 +44,18 @@ function RequestContextualizer() {
 
 RequestContextualizer.prototype = {
 
-    handle: function(globalLogger, defaultLang) {
+    handle: function () {
 
-        logger = globalLogger;
-        defaultLang = defaultLang || DEFAULT_LANG;
+        var my = this,
+            defaultLang;
 
-        var self = this;
+        return function (req, res, next) {
 
-        return function(req, res, next) {
+
+            if (!defaultLang && req.app && req.app.mojito) {
+                defaultLang = (req.app.get('mojito.context') &&
+                    req.app.get('mojito.context').lang) || DEFAULT_LANG;
+            }
 
             var query = url.parse(req.url, true).query || {};
 
@@ -63,29 +67,27 @@ RequestContextualizer.prototype = {
             req.context.site = query.site || '';
             // TODO: [Issue 86] add configuration switch to detect device
             req.context.device = query.device ||
-                self._device(req.headers['user-agent'],
-                    ''
-                    );
+                my._device(req.headers['user-agent'], '');
             req.context.lang = query.lang ||
-                self._language(req.headers['accept-language'], defaultLang);
+                my._language(req.headers['accept-language'], defaultLang);
             req.context.langs = query.langs ||
-                self._preferredLanguages(req.context.lang, defaultLang);
+                my._preferredLanguages(req.context.lang, defaultLang);
             req.context.region = query.region || '';
             req.context.jurisdiction = query.jurisdiction || '';
             req.context.bucket = query.bucket || '';
             req.context.flavor = query.flavor || '';
             req.context.tz = query.tz || '';
 
-            //logger.log('detected lang: ' + req.context.lang, 'debug',
+            // debug('detected lang: ' + req.context.lang, 'debug',
             //    'request-contextualizer');
-            //logger.log('detected device: ' + req.context.device, 'debug',
+            // debug('detected device: ' + req.context.device, 'debug',
             //    'request-contextualizer');
             next();
         };
     },
 
-    _device: function(ua, def) {
-//      logger.log('detecting device from UA: ' + ua);
+    _device: function (ua, def) {
+        // debug('detecting device from UA: ' + ua);
 
         // TODO: [Issue 74] Remove regex creation within this function scope,
         // and eventually offload to device catalog
@@ -117,7 +119,7 @@ RequestContextualizer.prototype = {
         return def;
     },
 
-    _language: function(al, def) {
+    _language: function (al, def) {
 
         al = (al || '').trim();
 
@@ -145,7 +147,7 @@ RequestContextualizer.prototype = {
         return chosen;
     },
 
-    _preferredLanguages: function(lang, defaultLang) {
+    _preferredLanguages: function (lang, defaultLang) {
         return [lang, defaultLang].join(',');
     }
 };
@@ -156,7 +158,7 @@ RequestContextualizer.prototype = {
  * @param {Object} config Data to configure the new contextualizer.
  * @return {Object} The contextualizer.
  */
-module.exports = function(config) {
-    var contextualizer = new RequestContextualizer();
-    return contextualizer.handle(config.logger, config.context.lang);
+exports = module.exports = function (config) {
+    var m = new RequestContextualizer();
+    return m.handle();
 };
--- a/lib/app/middleware/mojito-handler-dispatcher.js
+++ b/lib/app/middleware/mojito-handler-dispatcher.js
@@ -34,7 +34,8 @@ module.exports = function (config) {
 
     return function (req, res, next) {
         var command = req.command,
-            outputHandler;
+            outputHandler,
+            context = req.context || {};
 
         if (!command) {
             next();
@@ -58,12 +59,12 @@ module.exports = function (config) {
             log: Y.log
         });
 
-        // TODO: In `develop` branch we remove the copy, and instead, 
-        // staticAppConfig s considered an internal object that will NEVER 
-        // change.
-        // In fact we made few other changes that will change this block, but
-        // we can solve those conflicts when trying to merge this.
-        outputHandler.page.staticAppConfig = Y.mojito.util.copy(appConfig);
+        // storing the static app config as well as contextualized 
+        // app config per request
+        outputHandler.page.staticAppConfig = appConfig;
+        outputHandler.page.appConfig = store.getAppConfig(context);
+        // compute routes once per request
+        outputHandler.page.routes = store.getRoutes(context);
 
         // HookSystem::StartBlock
         // enabling perf group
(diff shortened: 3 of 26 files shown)


examples/newsboxes/app.js
```diff
@@ -0,0 +1,37 @@
...
+    mojito = require('../../'),
+    app;
+
+app = express();
+
+// register custom middleware here
+// app.use(require('./middleware/foo.js'));
+
+// Registers Mojito's default
+app.use(app.mojito.registerMiddleware());
```
caridy: `registerMiddleware` sounds awkward. should be `registeredMiddleware()` or just `app.mojito.middlewares()`. Now, middlewares should be stateless, which means they should not be bound to `app` at all. This makes me think that probably we should do:

```
app.use(mojito.middlewares());
```

which returns a middleware that groups a set of middleware that are bundled with mojito and takes in consideration any custom middleware in `app.json`. This is exactly what I do in `modown-yui` with `yui.expose()`.

examples/newsboxes/app.js
```diff
@@ -0,0 +1,37 @@
...
+/*
+// By doing this, give users more control over which order middleware
+// are registered.
+app.mojito.middleware().forEach(function (mid) {
```
caridy: again, this makes a lot of sense, by it will be `mojito.middlewares()`, which is stateless.

examples/newsboxes/app.js
```diff
@@ -0,0 +1,37 @@
...
+    debug('app.use(): ' + mid);
+    app.use(app.mojito[mid]);
```
caridy: probably the default middlewares should be attached to the function itself, something like mojito.middlewares[mid], where the function also acts like an object with members.

examples/newsboxes/app.js
```diff
@@ -0,0 +1,37 @@
...
+});
+*/
+
+// In addition to mojito `routes.json`, user can hook up additional
```
caridy: where are the routes from routes.json mounted? I don't see the call here. How can I remove that and do my own thing?

lib/mojito.js
```diff
@@ -36,55 +50,36 @@ var express = require('express'), // TODO: [Issue 80] go back to connect?
 global._mojito = {};
```
caridy: is this really needed?

lib/mojito.js
```diff
@@ -96,278 +91,236 @@ MojitoServer.MOJITO_MIDDLEWARE = [
...
+    };
+}
 
 /**
- * The server startup time. This value is used to both provide startup/uptime
- * information and as a signifier that the server has been configured/started.
- * @type {number}
- */
-MojitoServer.prototype._startupTime = null;
+NOTE: naming is based on ExpressYUIExtension in modown-yui
```
caridy: lol

examples/newsboxes/app.js
ekashida: I also like `mojito.middleware()`. Just a note: the plural of middleware is middleware :)

caridy: lol, thanks @ekashida. I knew something was not right with that "s". jajajaja

imalberto:
Thanks guys for all the feedback. I'll be making changes to this PR based on the ongoing discussion for the next review.

examples/newsboxes/app.js
```diff
@@ -0,0 +1,40 @@
+
+
+/*jslint node:true*/
+
+'use strict';
+
+var debug = require('debug')('app:yaml'),
```
caridy: should be 'app' instead of 'app:yaml' right?

examples/newsboxes/app.js
```diff
@@ -0,0 +1,40 @@
...
+    mojito = require('../../'),
+    app;
+
+app = express();
+
+// register custom middleware here
+// app.use(require('./middleware/foo.js'));
+
+// Registers Mojito's default
+app.use(mojito.middleware(app));
```
caridy: why do we need to pass `app` here? in theory once the first request comes in. you can get a handle on the app by doing `req.app`.

examples/newsboxes/app.js
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+// By doing this, give users more control over which order middleware
+// are registered.
+// But the recommended way to register all default middleware is:
+// `app.use(mojito.middleware());`
+mojito.defaultMiddleware().forEach(function (mid) {
```
caridy: `default` sounds ambiguous here because it is not about default, but those that are set up thru `application.json` isn't it?

lib/app/middleware/mojito-contextualizer.js
```diff
@@ -1,15 +1,15 @@
...
  */
 
 
-/*jslint anon:true, sloppy:true, nomen:true*/
+/*jslint node:true, nomen:true*/
 
 
-var qs = require('querystring'),
-    logger,
+var debug = require('debug')('middleware:contextualizer'),
```
caridy: for debug we should probably use `mojito:middleware:contextualizer`, so you can do any of these:

```
$ export DEBUG=*
$ export DEBUG=mojito:*
$ export DEBUG=mojito:middleware:*
$ export DEBUG=mojito:middleware:contextualizer
```

specially because we will have a lot of components

lib/app/middleware/mojito-contextualizer.js
```diff
@@ -1,15 +1,15 @@
...
+    qs = require('querystring'),
```
caridy, line 18: I heard that `qs` is faster than `querystring`, lets add a card to review this.

examples/newsboxes/app.js
imalberto: Right.

lib/app/middleware/mojito-contextualizer.js
```diff
@@ -44,14 +44,18 @@ function RequestContextualizer() {
...
 
-        logger = globalLogger;
-        defaultLang = defaultLang || DEFAULT_LANG;
+        return function (req, res, next) {
 
-        var self = this;
 
-        return function(req, res, next) {
+            if (!defaultLang && req.app && req.app.mojito) {
+                defaultLang = (req.app.get('mojito.context') &&
```
caridy: hmm, I usually do `app.set('mojito.context')`, because `get` is overloaded, and a mistake can cost you hours of debugging.

examples/newsboxes/app.js
imalberto: In this case, the default is really the "mojito built in" middleware.

Question: Do we want to support the "middleware" configuration in application.json in addition to `app.use()` ?

lib/app/middleware/mojito-contextualizer.js
```diff
@@ -44,14 +44,18 @@ function RequestContextualizer() {
...
+                    req.app.get('mojito.context').lang) || DEFAULT_LANG;
```
caridy: Is `mojito.context` attached once during the mojito object creation? then it is ok. I will say lets use `mojito context` instead of the dot. More a la express!.

Also, this code can be simplified, not need to ask fro req.app.mojito, and you can use a local variable probably for better readability.

Final question on this: how does that app level setting plays with the request level? do we want that?

examples/newsboxes/app.js
imalberto: `app` is needed to do the `app.use()` to register the middleware, and since `mojito` in this case is stateless, there is no reference to `app`.

lib/app/middleware/mojito-contextualizer.js
```diff
@@ -156,7 +158,7 @@ RequestContextualizer.prototype = {
  * @param {Object} config Data to configure the new contextualizer.
  * @return {Object} The contextualizer.
  */
-module.exports = function(config) {
-    var contextualizer = new RequestContextualizer();
-    return contextualizer.handle(config.logger, config.context.lang);
+exports = module.exports = function (config) {
+    var m = new RequestContextualizer();
```
caridy: why is this still needed? isn't easier to just export the function wrapper?

lib/app/middleware/mojito-handler-dispatcher.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+Mojito dispatcher middleware.
+
+@module mojito-dispatcher
+**/
+
+'use strict';
+
+var debug = require('debug')('middleware:dispatcher'),
```
caridy: mojito:middleware:dispatcher

lib/app/middleware/mojito-handler-dispatcher.js
```diff
@@ -0,0 +1,71 @@
...
+            outputHandler;
+
+        if (!command) {
+            next();
+            return;
+        }
+        
+        if (!store && req.app && req.app.mojito) {
+            store = req.app.get('mojito.store');
+            Y = req.app.get('mojito.Y');
```
caridy: eventually this will become `req.app.yui.use('mojito-dispatcher')` as a one time operation. Let's add a big TODO for that.

lib/app/middleware/mojito-contextualizer.js
imalberto: Like `module.exports = new RequestContextualizer().handle();` ?

lib/app/middleware/mojito-handler-dispatcher.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @param {Object} config
+ * @return {Function} express middleware
+ */
+module.exports = function (config) {
+
+    var appConfig,
+        store,
+        Y;
```
caridy, line 33: eventually, Y should not be in this middleware, only a handle to the dispatcher itself.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -21,14 +21,19 @@
...
+@module mojito-handler-static
+**/
+
 'use strict';
 
 /*
  * Module dependencies.
  */
-var liburl   = require('url'),
+var debug = require('debug')('middleware:static'),
```
caridy: mojito:*

lib/app/middleware/mojito-handler-static.js
```diff
@@ -21,14 +21,19 @@
 
 /*jslint node:true, nomen:true */
 
+/**
+
+@module mojito-handler-static
```
caridy, line 28: I will not review this one deeply because eventually modown-yui should kick in.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -196,11 +203,25 @@ function staticProvider(store, logger, Y) {
...
 
     return function (req, res, next) {
+
         if (req.method !== 'GET' && req.method !== 'HEAD') {
             next();
             return;
         }
 
+        if (!mojito && req.app && req.app.mojito) {
+            store = req.app.get('mojito.store');
```
caridy: I'm worry about using app.set() a lot, specially for things that are very mojito specific. The problem is that we don't know how express is handling that, and how efficient it is. In my opinion those things should be just `req.app.mojito.store` for example.

lib/logger.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+Configures YUI logger to honor the logLevel and logLevelOrder
+TODO: This should be done at the low level in YUI.
+
+@method configureLogger
+@public
+@param {YUI} Y shared YUI instance on the server
+**/
+module.exports = {
```
caridy, line 29: good call. the instance created for store as a sandbox should be passed thru this as well, the same for the commands that are creating Y instances.

lib/app/middleware/mojito-handler-static.js
imalberto: I originally had it in your suggested format. @drewfish and I had a quick discuss about it and thought `app.get()` is more `express` like. But your points are valid. I'll revert it to `req.app.xxx` for now and can revisit later if necessary.

lib/middleware.js
```diff
@@ -0,0 +1,139 @@
...
+var debug = require('debug')('middleware'),
+    libpath = require('path'),
+    Mojito = {};
+
+/**
+An ordered list of the middleware module names to load for a standard Mojito
+server instance.
+@type Array
+**/
+Mojito.MOJITO_MIDDLEWARE = [
```
caridy: not need to overload `Mojito`, it is only used to hold MOJITO_MIDDLEWARE, remove it.

lib/middleware.js
```diff
@@ -0,0 +1,139 @@
...
+    need customization.
+
+    @method registerMiddleware
+    @public
+    @param {express.application} app optional
+    @return {middleware}
+    **/
+    middleware: function (app) {
+
+        module.exports.defaultMiddleware().forEach(function (mid) {
```
caridy: not need to walk then and attach them one by one. express supports this:

```
app.use([foo, bar, baz]);
```

so you can return an array of middleware!

lib/middleware.js
```diff
@@ -0,0 +1,139 @@
...
+        });
+
+
+    Or better, use `app.use(app.mojito.registerMiddleware());` sugar.
+
+    @method middleware
+    @public
+    @return {Array} list of middlewares for a default Mojito application
+    **/
+    defaultMiddleware: function () {
```
caridy: this becomes irrelevant if the previous one returns an array.

lib/middleware.js
```diff
@@ -0,0 +1,139 @@
...
+        app.use(mojito['mojito-handler-static']);
+
+    @method exposeMiddleware
+    @protected
+    @param {express.application} app `express` app instance
+    @param {Function} dispatcher mojito's dispatcher middleware
+    @param {Object} midConfig configuration object to pass to middleware
+    @param {Array} middleware middleware names
+    **/
+    exposeMiddleware: function () {
```
caridy: not so sure what this does!

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
 
+    app.use(require('./middleware/my-middleware'));
 
-/**
- * Shared global object, which isn't named 'mojito' because 'mojito' is a module
- * name defined in mojito.common.js and required via Y.use.
- */
-// TODO: Merge what we put on this object with the 'mojito' module/namespace.
-global._mojito = {};
+    app.use(mojito.middleware(app));
```
caridy: remove app from here.

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
+    var app = require('express'),
+        mojito = require('mojito'),
+        app;
 
-//  ----------------------------------------------------------------------------
-//  Mojito Global
-//  ----------------------------------------------------------------------------
+    app = express(); // Mojito instance will be created during that call
 
+    app.use(require('./middleware/my-middleware'));
```
caridy: let's not promote this too much. simple simple simple

lib/middleware.js
imalberto: This attaches the middleware to the `mojito` function so that from `app.js`, user can do things like :

`app.use(mojito['mojito-handler-static]);`

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
 
-    // TODO: Note we could pass some options to the express server instance.
-    this._app = express.createServer();
 
-    this._app.store = store.createStore(this._options);
+//  ----------------------------------------------------------------------------
+//  Util
+//  ----------------------------------------------------------------------------
+// If more helpers are needed, they could potentially be moved to their own
+// `util.js` module later.
```
caridy: I already have it in another PR, once I merge it, we can move these methods to that new script.

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
 
 /**
- * The Express application (server) instance.
- * @type {Object}
- */
-MojitoServer.prototype._app = null;
+The date/time the Mojito object was initialized.
+@type Date
+**/
+Mojito.MOJITO_INIT = new Date().getTime();
```
caridy: let's get rid of all those non-sense timers, debug does that for us.

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
+    debug('applying mojito.defaultConfiguration()');
+
+    if (!app.mojito) {
+        // something is wrong: defaultConfiguration was not run
+        throw new Error('`app.mojito` was not correctly initialized!');
     }
 
-    return middleware;
+    // The application root directory
+    root = process.cwd();
```
caridy, line 102: watch out with this, manhattan has the wrong cwd.

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
+    store = libstore.createStore({
+        root: root,
+        preload: 'skip', // only need appConfig and package.json
+        context: { }
+    });
+    appConfig = store.getAppConfig();
+
+    options = {};
+    options.port = appConfig.appPort || process.env.PORT || 8666;
+    options.context = { }; // TODO: use correct static context here
```
caridy: lets make sure we force runtime: server somewhere

lib/mojito.js
```diff
@@ -1,269 +1,238 @@
...
+    options.perf =  'perf-log.log'; // TODO: should come from application.json
```
caridy: options.perf is going away.

lib/mojito.js
```diff
@@ -305,411 +274,55 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
...
-
-        Y.mojito.Dispatcher.init(store).dispatch(command, outputHandler);
-    }
+    // expose the middlewares to `mojito` function
+    mojitoFn.exposeMiddleware();
 
-    // attach middleware pieces
-    this._useMiddleware(app, dispatcher, options.dir, midConfig, middleware);
+    // stash for runtime
+    app.set('mojito.store', store);
```
caridy: as I mentioned before, this might be more suitable to be attached to `app.mojito` itself.

lib/mojito.js
imalberto: How are we recommending users who have app specific middleware to add them to `express` ?

lib/mojito.js
```diff
@@ -305,411 +274,55 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
...
+Hook into `express` default configuration init phase.
+**/
+appProto.defaultConfiguration = function () {
+    // debug('-- Begin defaultConfiguration --');
+    defaultConfiguration.apply(this, arguments);
+
+    if (!this.mojito) {
+        // These two probably can be merged.
+        this.mojito = new ExpressMojitoExtension(this);
+        this.mojito._createMojito();
```
caridy: why two separate call? what is _createMojito doing that the constructor of the extension can't do?

lib/logger.js
imalberto, line 29: Would need to match the log level from `application.json` - reading the log level without resource store seems like a hack. For now, it appears that resource store always log to "info" level.

lib/app/middleware/mojito-handler-dispatcher.js
```diff
@@ -0,0 +1,79 @@
...
+            Y = app.mojito.Y;
+            appConfig = store.getStaticAppConfig();
+        }
+
+        outputHandler = new OutputHandler(req, res, next);
+        outputHandler.setLogger({
+            log: Y.log
+        });
+
+        outputHandler.page.staticAppConfig = Y.mojito.util.copy(appConfig);
```
caridy: In `develop` branch we remove the copy, and instead, we consider `staticAppConfig` an internal object that we will NEVER change.
In fact we made few other changes that will change this block, but we can solve those conflicts when trying to merge this.

lib/app/middleware/mojito-handler-static.js
```diff
@@ -196,11 +205,26 @@ function staticProvider(store, logger, Y) {
     }
 
     return function (req, res, next) {
+
         if (req.method !== 'GET' && req.method !== 'HEAD') {
             next();
             return;
         }
 
+        if (!mojito && req.app && req.app.mojito) {
```
caridy: where is `mojito` being set? I don't see it in this file.

lib/app/middleware/mojito-handler-tunnel-parser.js
```diff
@@ -17,27 +22,36 @@ var RE_TRAILING_SLASHES = /\/+$/;
...
+                staticPrefix = staticPrefix.replace(RE_TRAILING_SLASHES, '');
+                staticPrefix = libpath.normalize('/' + staticPrefix);
+            }
+            if (tunnelPrefix) {
+                tunnelPrefix = tunnelPrefix.replace(RE_TRAILING_SLASHES, '');
+                tunnelPrefix = libpath.normalize('/' + tunnelPrefix);
+            }
+
+            staticPrefix = staticPrefix || '/static';
+            tunnelPrefix = tunnelPrefix || '/tunnel';
```
caridy: I made some changes here as well, merge will have to be done manually~

lib/app/middleware/mojito-router.js
```diff
@@ -57,6 +69,16 @@ Router.prototype = {
...
                 routeMerge;
 
+            // One time cost
+            if (!store && app.mojito) {
+                store = app.mojito.store;
+                Y = app.mojito.Y;
+                RouteMakerClass = Y.mojito.RouteMaker;
+            }
+
+            routes = store.getRoutes(context);
```
caridy: We don't want routes to be contexualized, instead you can compute `routeMaker` once as part of the previous if and use the static context for that.

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+
+        if (!store && app.mojito) {
+            debug('init mojito middleware');
+
+            store = app.mojito.store;
+            appConfig = store.getStaticAppConfig();
+
+            if (appConfig.middleware && appConfig.middleware.length > 0) {
+                debug('loading app specific middleware first');
+                appDir = process.cwd();
```
caridy: this is problematic, the path should come from app.js

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+                mm = [].concat(appConfig.middleware).concat(MOJITO_MIDDLEWARE);
+            } else {
+                mm = [].concat(MOJITO_MIDDLEWARE);
+            }
+
+            mm.forEach(function (m) {
+                if (MOJITO_MIDDLEWARE.indexOf(m) > -1) {
+                    app.use(middleware[m]);
```
caridy: I'm not so sure we should do this. Adding more middleware on demand will cause the actual middleware will not honor the order. Instead, I think we will have to emulate express here, doing our own simple walk on those middleware, and controlling the error and the next. I did that in modown-yui->middleware for the static, please, double check.

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+                    debug('app.use() => %s', m);
+                } else {
+                    try {
+                        // debug(libpath.join(appDir, m));
+                        mid  = libpath.join(appDir, m);
+                        debug('app.use() => %s', mid);
+                        mid = require(mid);
+                        app.use(mid());
```
caridy: same as above!

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+                    } catch (e) {
```
caridy: I'm not so sure we need to be defensive here.

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+        list = [].concat(MOJITO_MIDDLEWARE);
+
+
+    for (m = 0; m < list.length; m = m + 1) {
+        midName = list[m];
+        try {
+            midFactory = require(midName);
+        } catch (e1) {
+            try {
+                midPath = libpath.join(__dirname, 'app', 'middleware', midName);
```
caridy, line 181: are these really relative to __dirname?

lib/middleware.js
```diff
@@ -0,0 +1,174 @@
...
+                midFactory = require(midPath);
+            } catch (e2) {
+                midFactory = null;
+                console.error('failed to attach middleware: ' + midName);
```
caridy: instead of console.error we should `throw`

lib/app/middleware/mojito-handler-dispatcher.js
imalberto: I'll add a TODO in the source file with your explanation, and resolve it during the next merge.

lib/app/middleware/mojito-handler-static.js
imalberto: Thanks! its supposed to be `if (!store && req.app && req.app.mojito) {`

lib/app/middleware/mojito-handler-tunnel-parser.js
imalberto: Noted as a comment in the source file to resolve manually after the next merge with `develop`.

lib/middleware.js
imalberto: Will add a note to that effect that the appRootDir should either come from `app.js` or `locator` via an api at some point when `locator` is integrated. This is similar to : https://github.com/yahoo/mojito/pull/1101/files#L18R123

Until such time, `appDir = app.mojito.options.root;`

lib/middleware.js
imalberto, line 181: Based on the current directory structure where the built-in middleware lives, Yes.

lib/app/middleware/mojito-router.js
imalberto: Will update.

lib/middleware.js
imalberto: Will remove the `try/catch` and allow the error bubble up to the application. Better for debugging.

lib/mojito.js
```diff
@@ -305,416 +233,51 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
...
-            next();
-            return;
-        }
-
-        outputHandler = new OutputHandler(req, res, next);
-        outputHandler.setLogger({
-            log: Y.log
-        });
-
-        // storing the static app config as well as contextualized app config per request
```
caridy, line 342: all these are needed in the new dispatcher middleware.

imalberto, line 342: added the missing static and contextualized app config as part of this commit 23e63ce . thanks for catching this.

lib/mojito.js
```diff
@@ -1,281 +1,208 @@
...
- *          appConfig: Object,
- *          verbose: boolean}} options An object containing server options. The
- *          default port is process.env.PORT or port 8666 if no port is given.
- *          Verbose is false by default. Dir is cwd() by default. Both the
- *          context and appConfig will default to empty objects.
- * @constructor
- * @return {MojitoServer}
- */
-function MojitoServer(options) {
+Copied from modown-yui.utils
```
caridy: This can now be moved into `lib/util.js`

imalberto: updated (commit     c3703c1 )

caridy:
I think there is nothing missing at this point.

+1

PR #1104, opened 2013-05-08
use ycb-based tokens like $$lib$$ instead of repeating subpaths

isao:
requires ycb@1.0.5; previously only the first token would be replaced.

the tests pass locally, forcing mojito and arrow to use ycb@1.0.5. Travis/CI won't pass until deps specify ycb@1.0.5 (currently tagged "beta" not "latest").

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
         "rimraf": ">2",
         "semver": "1.0.14",
         "wrench": "~1.3.9",
-        "ycb": "~1.0.0",
+        "ycb": "~1.0.5",
         "yui": "~3.9.1",
         "yuidocjs": "~0.3.14",
         "yuitest": "~0.7.4",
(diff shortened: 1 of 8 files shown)


drewfish:
once builds pass, +1

drewfish:
we should update the YCB dependency in package.json to "~1.0.5"

isao:
yes, I'll add the ~1.0.5 change to this PR

isao:
Note-- will also need to wait for arrow@0.0.76 for these updated descriptor files to work as expected...

isao:
I made the dep change locally. I'll push it when arrow@0.0.76 comes out (and mojito's CI suites pass), to get the travis checkmark...

```
% git diff -U1 head^ package.json
diff --git a/package.json b/package.json
index 4d2e442..562cde5 100644
--- a/package.json
+++ b/package.json
@@ -27,3 +27,3 @@
         "wrench": "~1.3.9",
-        "ycb": "~1.0.0",
+        "ycb": "~1.0.5",
         "yui": "~3.9.1",
@@ -55,3 +55,3 @@
         "node-static": ">0.6.8",
-        "yahoo-arrow": ">=0.0.75"
+        "yahoo-arrow": "~0.0.76"
     },
```

PR #1109, opened 2013-05-10
tweaks to assist with running tests locally

drewfish:


isao:
+1

hardcoding --reuseSession is fine and good. But the original intent of this shell script is to be able to run all the tests the same way in travis/ci et al. There already is run.js with lots of config/options, which is a convenience wrapper script for arrow. And there's using the arrow cli itself. Effort to make it more convenient in this script I think would be better spent simplifying run.js.

PR #1110, opened 2013-05-13
add command 'npm test' to run all mojito unit and functional tests

lzhan:


package.json
```diff
@@ -55,6 +55,12 @@
         "node-static": ">0.6.8",
         "yahoo-arrow": ">=0.0.75"
     },
+    "optionalDependencies": {
+        "phantomjs": ">=1.8.0"
+    },
+	"scripts": {
```
caridy: weird spacing here!

lzhan:
Thanks Caridy. Updated.

HISTORY.md
```diff
@@ -16,6 +16,27 @@ Bug Fixes
 Acknowledgements
 ----------------
 
+version 0.5.9
```
drewfish: This shouldn't be here, since we don't know what the version number of the next release will be.  (It might be 0.5.9pr2 or something like that.)  Instead add new details to the `version @VERSION@` section.

tests/run.js
```diff
@@ -188,35 +202,59 @@ function build (cmd, callback) {
...
+        }
+        console.log('phantomjs failed to start. Please check if phantomjs is installed');
+        pids.pop();
+        callback(1); // Trigger failure
+    });
+    p.stdout.on('data', listener);
+    pids.push(p.pid);
+    pidNames[p.pid] = 'phantomjs driver';
+    timeout = setTimeout(function () {
+        p.stdout.removeListener('data', listener); // Stop printing output from arrow_server
```
drewfish: Should this comment be `// Stop printing output from phantomjs` ?

HISTORY.md
lzhan: Ok. I will update it.

tests/run.js
lzhan: Good catch. Blame copy and paste.

tests/run.js
```diff
@@ -188,35 +207,59 @@ function build (cmd, callback) {
...
+            clearTimeout(timeout);
+        }
+        console.log('phantomjs failed to start. Please check if phantomjs is installed');
+        pids.pop();
+        callback(1); // Trigger failure
+    });
+    p.stdout.on('data', listener);
+    pids.push(p.pid);
+    pidNames[p.pid] = 'phantomjs driver';
+    timeout = setTimeout(function () {
```
drewfish, line 247: Another approach is to put some code in `listener` to look for a string where phantomjs says that it's ready.

tests/runall.sh
```diff
@@ -2,6 +2,7 @@
 # -e exits on any error; exit code is preserved
 
 ARGS="--reuseSession"
+ARGSFUNC= "--browser phantomjs"
```
drewfish: If `ARGSFUNC` is specific to just the functional tests, it might be nice to move it into the `# func tests` section.

drewfish:
small notes, but otherwise +1

PR #1111, opened 2013-05-13
Y.Model.Vanilla on the server side instead of Y.Model

caridy:
## Goal
- Provide a consistent api between models in the server and models in the client
- Provide a performance solution for models in the server considering that we don't want to support events in that runtime
## Investigation

Y.Model and Y.Model.Base seems to be consirable slower than vanilla, which uses the same API but does not rely on Y.BaseCore. The preliminary results are:

```
new Y.Model.Vanilla() x 119,220 ops/sec ±1.03% (97 runs sampled)
new Y.Model.Base() x 23,223 ops/sec ±2.95% (92 runs sampled)
new Y.Model() x 14,673 ops/sec ±2.56% (79 runs sampled)
```

> Fastest is new Y.Model.Vanilla()

These results are based on this operation:

```
(new Y.Model.Vanilla({miami: 'florida'})).toJSON();
```

which is pretty much what we do internally per mojits with `mojito-data-addon` enabled.
## Solution
- As an interim solution we will use `Y.Model.Vanilla`, which is hacked into the mojito framework until `Y.Model.Base` gets ready


lib/app/autoload/mojit-proxy.client.js
```diff
@@ -200,7 +200,8 @@ YUI.add('mojito-mojit-proxy', function(Y, NAME) {
                 base: this._base,
                 type: this.type,
                 instanceId: this._instanceId,
-                config: Y.mojito.util.copy(this.config)
+                config: Y.mojito.util.copy(this.config),
+                data: this.data // passing the hydrated data model to be re-used
```
caridy, line 204: this is just to re-use the data model associated to a mojit instance whenever the new instance is created from the mojitProxy, which already rehydrated the data model. This also solves the problem of the communication between proxy and instance when running in the client side.

tests/benchmarks/models.js
```diff
@@ -0,0 +1,64 @@
+console.warn('Make sure you have: npm install benchmark -g');
```
caridy, line 1: I do have few benchmarks tests locally, I plan to start adding them into the repo to be ran manually. The only problem so far is that `benchmark` is still not whitelisted, working on that with Gil.

lib/app/addons/ac/data.common.js
```diff
@@ -20,12 +20,11 @@ YUI.add('mojito-data-addon', function (Y, NAME) {
      * @class Data.common
      */
     function Addon(command, adapter, ac) {
-        // create instance of data
-        ac.data = command.instance.data = new Y.Model();
-        // creating instance of page if needed
-        ac.pageData = adapter.page.data = adapter.page.data || new Y.Model();
-        // TODO: should we try to re-hydrate the ac.data when reviving
-        // the instance of the mojit in the client side thru mojitProxy?
+        var Class = (ac.context && ac.context.runtime !== 'client' ? Y.Model.Vanilla : Y.Model);
```
caridy, line 23: using vanilla class when running on non-client runtime (aka server)

lib/app/autoload/model-vanilla.common.js
```diff
@@ -0,0 +1,797 @@
...
+    customize it to meet your needs. In particular, the `sync()` and `validate()`
+    methods are meant to be overridden by custom implementations. You may also want
+    to override the `parse()` method to parse non-generic server responses.
+
+    @class Model.Base
+    @constructor
+    @extends BaseCore
+    @since 3.8.0
+    **/
+    function ModelBase(config) {
```
drewfish: Might be clearer if this is called `ModelVanilla` in this file.

tests/benchmarks/models.js
```diff
@@ -0,0 +1,64 @@
...
+}).use('model', 'model-base', 'model-vanilla');
+
+var data = { miami: 'florida' };
+
+// warm up process
+console.log((new Y.Model.Base(data)).toJSON());
+
+console.log((new Y.Model.Vanilla(data)).toJSON());
+
+console.log((new Y.Model(data)).toJSON());
```
drewfish, line 25: Do we really need to `console.log()`?  It'd be nice if all tests (unit, func, benchmark, etc) only wrote to STDOUT when absolutely necessary.

tests/benchmarks/models.js
```diff
@@ -0,0 +1,64 @@
...
+        (new Y.Model.Base(data)).toJSON();
+
+    })
+    .add('new Y.Model()', function() {
+
+        (new Y.Model(data)).toJSON();
+
+    })
+    // run async
+    .run({ 'async': true });
```
drewfish, line 64: Is this meant to be run as part of the automated testing?

caridy, line 64: no! function benchmarks are manual operations most of the time.

tests/benchmarks/models.js
caridy, line 25: this is for manual process so far.

lib/app/autoload/model-vanilla.common.js
caridy: Agreed, I just copied the same code from ModelBase but change the way it gets built, but I wanted to preserve the actual code from @ericf

drewfish: I'm confused.  This PR has `autoload/model-base.common.js` and `autoload/model-vanilla.common.js` files.  Are these two implementations different?  If they're different then the ctor in model-vanilla.common.js should be called ModelVanilla, I argue.  If they're the same, why do we need a model-vanilla.common.js?

caridy: They are equivalents in terms of code and features, we will use one of them. The main different between them is the way the class is constructed, once of them relies on Y.BaseCore, while the other is just a prototype and few mixes.

drewfish: OK.  I still find it confusing that a file called `model-vanilla.js` is creating a prototype called `ModelBase` and then exposing it as `Y.Model.Vanilla`.  The prototype name should be `ModelVanilla` in my mind.

caridy:
`Y.Model.Vanilla` is also related to issue #259, just for reference!

lib/app/autoload/model-vanilla.common.js
caridy: @drewfish I can change the code to use `ModelVanilla`, but remember we didn't create this code, we copied as a temporary solution, and any update in the original code should be pulled in here as well, so the less changes, the better.

drewfish: I mostly agree, however note that there is also `Y.Model.Vanilla = ModelBase;` in this file.  See how confusing that is?  If this is just temporary (if we'll switch in the future to using something provided directly from YUI) then it's no big deal.

PR #1113, opened 2013-05-15
relying more on the adapter.page object to store appConfig and routes

caridy:
For the sake of performance, and in preparation for adopting a more `express`-like approach, which is a less defensive approach btw, we now:
- compute `appConfig` and `routes` once per request (by default) instead of computing them per mojit instance from within the addons.
- deprecate `ac.staticAppConfig` in favor of `adapter.page.staticAppConfig` to easy the GC process by having less big objects hanging around.

### DIFF
--- a/lib/app/addons/ac/carrier.server.js
+++ b/lib/app/addons/ac/carrier.server.js
@@ -65,6 +65,5 @@ YUI.add('mojito-carrier-addon', function(Y, NAME) {
 
 }, '0.1.0', {requires: [
     'mojito',
-    'mojito-config-addon',
     'mojito-http-addon'
 ]});
--- a/lib/app/addons/ac/deploy.server.js
+++ b/lib/app/addons/ac/deploy.server.js
@@ -58,9 +58,6 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
 
             var store = this.rs,
                 contextServer = this.ac.context,
-
-                appConfigServer = store.getAppConfig(contextServer),
-
                 contextClient,
                 appConfigClient,
                 yuiConfig = {},
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -307,9 +307,6 @@ YUI.add('mojito-action-context', function(Y, NAME) {
         this.instance = command.instance;
         this._adapter = opts.adapter;
 
-        // pathToRoot, viewEngine, amoung others will be available through this.
-        this.staticAppConfig = (this._adapter.page && this._adapter.page.staticAppConfig) || store.getStaticAppConfig();
-
         // Create a function which will properly delegate to the dispatcher to
         // perform the actual processing.
         this._dispatch = function(command, adapter) {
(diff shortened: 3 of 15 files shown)


lib/app/addons/ac/deploy.server.js
```diff
@@ -58,9 +58,6 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
 
             var store = this.rs,
                 contextServer = this.ac.context,
-
-                appConfigServer = store.getAppConfig(contextServer),
```
caridy, line 62: this was not used at all in deploy.

lib/app/autoload/action-context.common.js
```diff
@@ -307,9 +307,6 @@ YUI.add('mojito-action-context', function(Y, NAME) {
         this.instance = command.instance;
         this._adapter = opts.adapter;
 
-        // pathToRoot, viewEngine, amoung others will be available through this.
-        this.staticAppConfig = (this._adapter.page && this._adapter.page.staticAppConfig) || store.getStaticAppConfig();
```
caridy, line 311: In this case, we might want to keep it around for a while, but the objective was to be used as an internal api, I don't think it is even documented, and probably removing it just fine.

lib/mojito.js
```diff
@@ -338,7 +339,11 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
             log: Y.log
         });
```
caridy, line 341: provision all the stuff per request.

caridy:
Unit tests passed in travis!

lib/app/addons/ac/carrier.server.js
```diff
@@ -65,6 +65,5 @@ YUI.add('mojito-carrier-addon', function(Y, NAME) {
 
 }, '0.1.0', {requires: [
     'mojito',
-    'mojito-config-addon',
```
caridy, line 68: this doesn't seem to be used at all in carrier.

lib/app/autoload/action-context.common.js
drewfish, line 311: Yeah, ac.config.getAppConfig() is preferred for the user.

lib/mojito.js
```diff
@@ -338,7 +339,11 @@ MojitoServer.prototype._configureAppInstance = function(app, options) {
...
 
+        // storing the static app config as well as contextualized app config per request
         outputHandler.page.staticAppConfig = store.getStaticAppConfig();
+        outputHandler.page.appConfig = store.getAppConfig(context);
+        // compute routes once per request
+        outputHandler.page.routes = store.getRoutes(context);
```
drewfish, line 346: Hmm... the router middleware should have run before this dispatcher, and it too should have called store.getRoutes(req.context), so I wonder if that can be reused.  Probably not much of an optimization.

drewfish:
action-context.common.js done() seems to have an unused pageData variable.

drewfish:
A few small comments, even without addressing them, +1.

lib/app/autoload/action-context.common.js
caridy, line 311: good!

lib/mojito.js
caridy, line 346: I'm just trying to preserve the current structures, but I do agree with you that routes should be computed once, they should not be tied to runtime context, and that's the way we want to move forward.

PR #1114, opened 2013-05-15
revert arrow, descriptor upgrade commit

isao:


drewfish:
+1

isao:
BTW func tests on travis fail with:

```
The log length has exceeded the limit of 4 Megabytes
```

lzhan:
That is the issue Drew and I mentioned in yesterday's standard up meeting.
Drew will check to see if we can turn off the warning.

Thanks,
-lichun

PR #1116, opened 2013-05-15
add some delay for the tests as they constantly failed on sd with phantomjs

lzhan:


drewfish:
Could try using `YUI.add()` for the tests, instead of `YUI.use()`?  (We might need to change all tests in an app to use the same approach.)

lzhan:
sure. I will make the changes.

drewfish:
+1

PR #1117, opened 2013-05-16
resource store honors yui.config during preload

drewfish:
This will only really work once we move to yui@3.10.1 since earlier versions don't honor `logLevel` in the config.


isao:
+1

caridy:
:+1:

PR #1118, opened 2013-05-16
introducing mojito-output-buffer

caridy:
- partial addon should honor the original adapter configuration
- composite and partial can use `Y.mojito.OutputBuffer` as the primary interface with dispatcher and render
- composite and partial support `ac.flush`
- fixes jenkins func tests

_Note: I don't think `ac.partial.invoke()` feature should be part of partial addon, but for the sake of BC, here is the fix._

### DIFF
--- a/lib/app/addons/ac/partial.common.js
+++ b/lib/app/addons/ac/partial.common.js
@@ -14,6 +14,49 @@
  */
 YUI.add('mojito-partial-addon', function(Y, NAME) {
 
+    function InvokeAdapter(callback) {
+        this.data = '';
+        this.callback = function () {
+            if (callback) {
+                callback.apply(this, arguments);
+            }
+            callback = function () {
+                Y.log('ac.done/flush/error called multiple times in partial invoke',
+                        'warn', NAME);
+            };
+        };
+    }
+
+    InvokeAdapter.prototype = {
+
+        done: function (data, meta) {
+            this.flush(data, meta);
+            this.callback(null, this.data, this.meta);
+        },
+
+        flush: function (data, meta) {
+            this.data += data;
+            // this trick is to call metaMerge only after the first pass
+            this.meta = (this.meta ? Y.mojito.util.metaMerge(this.meta, meta) : meta);
+        },
+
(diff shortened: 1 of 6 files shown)


lib/app/addons/ac/partial.common.js
```diff
@@ -14,6 +14,50 @@
  */
 YUI.add('mojito-partial-addon', function(Y, NAME) {
 
+    function InvokeAdapter(callback) {
```
caridy: this class is 99% similar to the one used in composite. It seems like anyone who wants to invoke an action from code will be using this class as a buffer to control the output, and we should generalize it, so we have a single implementation for all components.

lib/app/addons/ac/partial.common.js
```diff
@@ -21,13 +65,12 @@ YUI.add('mojito-partial-addon', function(Y, NAME) {
     * @class Partial.common
     */
     function Addon(command, adapter, ac) {
-        this.command = command;
-        this.dispatch = ac._dispatch;
-        this.ac = ac;
+        this._command = command;
+        this._ac = ac;
+        this._adapter = adapter;
```
caridy, line 26: cosmetic changes and v8 optimizations.

lib/app/addons/ac/partial.common.js
```diff
@@ -102,41 +148,47 @@ YUI.add('mojito-partial-addon', function(Y, NAME) {
...
+            // the new adapter that inherit from the original adapter
+            newAdapter = new InvokeAdapter(function (err, data, meta) {
 
-            this.dispatch(command, {
-                data: '',
-
-                meta: {},
+                // HookSystem::StartBlock
+                Y.mojito.hooks.hook('adapterInvoke', my._adapter.hook, 'end', this);
+                // HookSystem::EndBlock
```
caridy, line 119: adding hooks for the async process.

caridy:
travis is failing big, but has nothing to do with this PR.

lzhan:
+1

drewfish:
+1

PR #1119, opened 2013-05-16
an experiment in version upgrades

drewfish:


drewfish:
experiment over.  arrow 0.0.76 just isn't cutting it for us.  (arrow reported failed tests yet didn't return a non-zero exit code so that the build could fail.)

PR #1120, opened 2013-05-16
optimize how expandInstanceForEnv() merges datastructures

drewfish:


lib/app/autoload/store.server.js
```diff
@@ -528,7 +522,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
 
             // spec
             try {
-                spec = this._expandSpec(env, ctx, fakeInstance);
+                spec = this._expandSpec(env, ctx, instance);
```
caridy, line 525: does `_expandSpec` accept a full instance? remember that this can have way more info than the `fakeInstance` has in the past.

caridy:
+1 fingers crossed!

lib/app/autoload/store.server.js
drewfish, line 525: Yeah, _expandSpec() should handle a full instance just fine.  In fact, _expandSpec() can recursively merge specs (follow a chain of "base"), so it can/should merge data.  The main benefit to passing the `instance` to _expandSpec() is that it allows us to use a blend() later (instead of blend3()).

drewfish:
Search benchmarked this with their app, results in bz 6296025.

drewfish:
Based on search's results, I think this PR should be abandoned :disappointed: .

caridy:
I will like to get this compared to `develop` and `pr1` to be sure.

drewfish:
I've run the tests myself and posted the numbers to the bug.  It looks like this is indeed faster for search.

drewfish:
From retesting by search:

```
mojito@0.5.9pr1:
10.3 conn/s - 97.5 ms

mojito@develop:
11.0 conn/s - 90.8 ms

mojito@develop + pr1120:
11.3 conn/s - 88.3 ms

So it looks like pr1120 has possibly a slight positive impact, if any. Indeed,
there is some variance between 2 consecutive tests. However, what is pretty
clear is that the develop branch has a clear advantage over pr1.
```

PR #1125, opened 2013-05-17
add command "npm run <commands>" to run unit, func, lint, docs, etc

lzhan:
This is basing on Caridy's feedback and Isao's suggestion:
"npm run testunit" or "npm run-script testunit" will run unit tests only with coverage data generated too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
         "phantomjs": ">=1.8.0"
     },
     "scripts": {
-        "test": "cd tests && ./run.js test --browser phantomjs",
-        "testunit": "cd tests && ./run.js test -u --browser phantomjs --coverage"
+        "test": "cd tests && node run.js test --browser phantomjs",
+        "testunit": "cd tests && node run.js test -u --browser phantomjs --coverage"
     },
     "homepage": "http://developer.yahoo.com/cocktails/mojito/",
     "repository": {
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
         "unit": "cd tests && node run.js test -u --browser phantomjs",
         "func": "cd tests && node run.js test -f --browser phantomjs",
         "lint": "node bin/mojito jslint",
-        "doc": "node bin/mojito docs mojito"
+        "docs": "node bin/mojito docs mojito"
     },
     "homepage": "http://developer.yahoo.com/cocktails/mojito/",
     "repository": {
--- a/package.json
+++ b/package.json
@@ -60,7 +60,10 @@
     },
     "scripts": {
         "test": "cd tests && node run.js test --browser phantomjs",
-        "testunit": "cd tests && node run.js test -u --browser phantomjs --coverage"
+        "unit": "cd tests && node run.js test -u --browser phantomjs",
+        "func": "cd tests && node run.js test -f --browser phantomjs",
+        "lint": "node bin/mojito jslint",
+        "doc": "node bin/mojito docs mojito"
     },
     "homepage": "http://developer.yahoo.com/cocktails/mojito/",
     "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,7 +59,8 @@
         "phantomjs": ">=1.8.0"
     },
     "scripts": {
-        "test": "cd tests && ./run.js test --browser phantomjs"
+        "test": "cd tests && ./run.js test --browser phantomjs",
```
caridy: please, do NOT use `./` notation, that does NOT work in windows, instead, do `node run.js`

package.json
```diff
@@ -59,7 +59,8 @@
...
+        "testunit": "cd tests && ./run.js test -u --browser phantomjs --coverage"
```
caridy: same as above!

lzhan:
Thanks. I will correct it.

package.json
```diff
@@ -59,7 +59,8 @@
         "phantomjs": ">=1.8.0"
     },
     "scripts": {
-        "test": "cd tests && node run.js test --browser phantomjs"
+        "test": "cd tests && node run.js test --browser phantomjs",
+        "testunit": "cd tests && node run.js test -u --browser phantomjs --coverage"
```
caridy: I want to have more scripts, I think it is really clear how to run then, and that might help a lot in terms of productivity. Now, assuming that npm test is what our CI will probably use, and what we, as developers, will run when we really want to validate the whole library, I will argue that all the other test command should NOT have `--coverage` to speed things up, and have some special command to generate coverage, something like:

```
npm run coverage
```

I also will like to rename this to just `unit`, and the same for `func` should be added, so we can do:

```
npm run unit
npm run func
```

I will also like to see:

```
npm run lint
npm run docs
```

:)

drewfish: This sounds nice to me too.

lzhan: Thanks for the suggestion. I will add other commands.  
About coverage: I know it is a little awkward that only one command has it. Currently we can only generate cc data for unit tests. And unit test execute is not that long. I thought it is good to see the coverage when dev add feature/unit tests.

package.json
```diff
@@ -59,7 +59,11 @@
         "phantomjs": ">=1.8.0"
     },
     "scripts": {
-        "test": "cd tests && node run.js test --browser phantomjs"
+        "test": "cd tests && node run.js test --browser phantomjs",
+        "unit": "cd tests && node run.js test -u --browser phantomjs",
+        "func": "cd tests && node run.js test -f --browser phantomjs",
+        "lint": "node bin/mojito jslint",
+        "doc": "node bin/mojito docs mojito"
```
caridy: `docs` please

package.json
isao: I like these suggestions, but fwiw, time penalty for coverage is negligible for mojito's unit tests, which is why I suggested it to @lzhan. I do have `npm run cover` in a bunch of places distinct from `npm test`, but mainly because I didn't declare istanbul as a devDependency in those cases.

package.json
isao: note to self: docs & lint will need to be edited to use mojito-cli when lib/app/commands is removed.

package.json
```diff
@@ -59,7 +59,11 @@
         "phantomjs": ">=1.8.0"
     },
     "scripts": {
-        "test": "cd tests && node run.js test --browser phantomjs"
+        "test": "cd tests && node run.js test --browser phantomjs",
```
caridy, line 62: Is this cd "path" affecting the users path? if yes, then we should find another solution or restoring the path afterwork.

caridy:
+1

package.json
lzhan, line 62: I don't think so.

PR #1128, opened 2013-05-20
 [Fixes #1127] Allow `Y.Handlebars.registerHelper` to register helpers

### LINKED ISSUE #1127: Allow YUI way of adding Handlebars helpers
As per http://yuilibrary.com/yui/docs/handlebars/#defining-custom-helpers, the Handlebars helpers are defined using `Y.Handlebars.registerHelper`.

However, any helper registered using the above way are not available while generating views from hb templates on server.

Use case: Adding set of global handlebar helpers that works on both client and server. (helpers like pluralize, lowercase, uppercase, truncate etc. which are independent of mojit). More like, I want the Handlebars engine to be enhanced where I've access to a bigger set of helpers with almost no extra effort.

I understand `ac.helpers.set` and `ac.helpers.expose` are available, but I find a couple of concerns using this approach:
1. Needs access to `ac` which is available in each function of controller, and hence every function needs to attach these helpers.
2. Incompatibility with existing YUI code, which declared helpers this way.

saich:
As per http://yuilibrary.com/yui/docs/handlebars/#defining-custom-helpers, the Handlebars helpers are defined using `Y.Handlebars.registerHelper`.

However, any helper registered using the above way are not available while generating views from hb templates on server.

Use case: Adding set of global handlebar helpers that works on both client and server. (helpers like pluralize, lowercase, uppercase, truncate etc. which are independent of mojit). More like, I want the Handlebars engine to be enhanced where I've access to a bigger set of helpers with almost no extra effort.

I understand `ac.helpers.set` and `ac.helpers.expose` are available, but I find a couple of concerns using this approach:
1. Needs access to `ac` which is available in each function of controller, and hence every function needs to attach these helpers.
2. Incompatibility with existing YUI code, which declared helpers this way.

### DIFF
--- a/lib/app/addons/view-engines/hb.server.js
+++ b/lib/app/addons/view-engines/hb.server.js
@@ -14,7 +14,7 @@ YUI.add('mojito-hb', function (Y, NAME) {
     'use strict';
 
     var fs = require('fs'),
-        HB = require('yui/handlebars').Handlebars,
+        HB = Y.Handlebars,
         cache = YUI.namespace('Env.Mojito.Handlebars');
 
     /**
@@ -209,6 +209,7 @@ YUI.add('mojito-hb', function (Y, NAME) {
 
 }, '0.1.0', {requires: [
     'mojito',
+    'handlebars',
     'parallel',
     'mojito-hooks'
 ]});


caridy:
@saich the problem here is that `ac.helpers` are, or should be, template engine agnostic. They should be stateless functions that can be serialized and sent to the client side (coming soon). What changes is the fact that every engine implementation might use them or not. A lot of changes are coming in this area, so I will close this PR and we will see how to accomodate things in the future.

PR #1137, opened 2013-05-22
Add checking of page loading.

lzhan:


drewfish:
+1

Would be nice if the YQL calls where mocked out instead.

lzhan:
Agree. Need to think about it for all apps with YQL calls.

PR #1141, opened 2013-05-24
Fixing functional tests on #next

imalberto:
CHANGELOG:
- removed the error handler from `lib/middleware.js` and instead allow application to setup their own error handler. if not specified, the default `express` one will be used.
- updated `lib/middleware/mojito-tunnel-*` for functional tests to pass
- updated `tests/run.js` to ignore set of tests to be revisited or removed later; could not find a way to disable those tests in arrow descriptor configuration.
- updated for `lib/mojito.js` to use port value from `process.env` if set first, then `application.json` to allow environment customization without `cli` support

TESTS TO REVISIT (or disabled permanently ?)

The following tests have been disabled (no html5app and context appConfig in #next yet).

```
'html5apptest_descriptor.json',                                         
'configtest0_descriptor.json',                                          
'configtest1_descriptor.json',                                          
'configtest2_descriptor.json',                                          
'configtest3_descriptor.json',                                          
'configtest4_descriptor.json',                                          
'configtest5_descriptor.json',                                          
'configtest6_descriptor.json',                                          
'simple_logging_descriptor.json'  
```

### DIFF
--- a/examples/developer-guide/adding_view_engines/app.js
+++ b/examples/developer-guide/adding_view_engines/app.js
@@ -1,0 +1,28 @@
+/*
+* Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
+* Copyrights licensed under the New BSD License.
+* See the accompanying LICENSE file for terms.
+*/
+
+
+/*jslint node:true*/
+
+'use strict';
+
+var debug = require('debug')('app'),
+    express = require('express'),
+    mojito = require('../../../'),
+    app;
+
+app = express();
+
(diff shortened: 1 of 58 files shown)


caridy:
+1

Please, add two cards to trello:
- todo to fix the skipped tests
- todo to remove the old server.js files

imalberto:
@caridy Already added the Trello cards. Thanks !

imalberto:
TravisCI says build passes both for unit and functional tests !!

https://travis-ci.org/yahoo/mojito/builds/7481963

PR #1150, opened 2013-05-30
cleanup callback handling logic, add as err listener

isao:
errors on app.listen do not invoke the callback passed as a parameter, but emit an 'error' event. So if a callback is provided, add it as an error event listener.

### DIFF
--- a/lib/mojito.js
+++ b/lib/mojito.js
@@ -497,8 +497,7 @@ MojitoServer.prototype.getHttpServer = function() {
  */
 MojitoServer.prototype.listen = function(port, host, callback) {
 
-    var logger, // <- not used in this function, cruft?
-        app = this._app,
+    var app = this._app,
         p = port || this._options.port,
         h = host || this._options.host,
         listenArgs = [p];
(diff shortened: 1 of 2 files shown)


lib/mojito.js
```diff
@@ -497,15 +497,10 @@ MojitoServer.prototype.getHttpServer = function() {
  */
 MojitoServer.prototype.listen = function(port, host, callback) {
 
-    var logger,
+    var logger, // <- not used in this function, cruft?
```
drewfish: If not used, go ahead and chop it out.

drewfish:
other than cleanup, +1

lib/mojito.js
isao: yup, I removed it a few minutes after this initial push, after running the tests and some apps.

isao: I was kind of confused how this file didn't get a jslint error for that. in the old mojito jslint command, there is an exclusion rule for /mojito/, so I was paranoid there was something intentionally funky. we'll see.

After removing it, jshint still says:
- mojito.js: line 22, col 19, 'requestCounter' is defined but never used.
- mojito.js: line 262, col 18, 'logConfig' is defined but never used.
- mojito.js: line 266, col 20, 'debugConfig' is defined but never used.
- mojito.js: line 382, col 14, 'm' is defined but never used.

Issue #1151, opened 2013-06-03
No Sanity Check in route-maker.common.js causing error on client side in IE-7 IE-8


chetanankola:
https://github.com/yahoo/mojito/blob/develop/lib/app/autoload/route-maker.common.js#L124

route.requires[matches[i].substr(1)] = '[^&]+';

fails when matches[i] is a number! 

This is urgent for us since its breaking our app on ie 6 and greater!

isao:
@chetanankola can you provide a repro, or gist of the routes.json?

caridy:
@isao can we close this?

jenny:
Will be avail in 0.6.1 this week!

PR #1152, opened 2013-06-03
fixes issue #1151 client-side route maker error IE6-8

isao:
@chetanankola reports http://git.io/Si_T_g

```
route.requires[matches[i].substr(1)] = '[^&]+';
fails when matches[i] is a number!
```

Since String#match returns null or an array, replacing the
for/in with a plain old for-loop behind an if(matches) is the better
thing to do anyway. @chetanankola reports this works in IE now. 
This error appears to already be covered by the test cases in 
tests/unit/lib/app/autoload/test-route-maker.common.js

### DIFF
--- a/lib/app/autoload/route-maker.common.js
+++ b/lib/app/autoload/route-maker.common.js
@@ -119,8 +119,8 @@ YUI.add('mojito-route-maker', function(Y, NAME) {
              */
             route.requires = {};
             matches = route.path.match(/:([^\/]+)/g);
-            for (i in matches) {
-                if (matches.hasOwnProperty(i)) {
+            if (matches) {
+                for (i = 0; i < matches.length; i++) {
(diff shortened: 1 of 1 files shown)


isao:
just to add, the `routes.json` in question looks like:

```
       "routes": {
           "_default_path": {
               "verbs": ["GET"],
               "path": "/:module/:action",
               "call": "*.*"
           }
       },
```

test cases for this type of route exist already http://git.io/6W_avQ

unit tests pass on travis.

drewfish:
+1

isao:
fwiw travis errored on node 0.10 - func only, unit + func passed on the others. error was:

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

PR #1153, opened 2013-06-03
fix lint: i++ -> i += 1

isao:


lzhan:
+1

isao:
see yahoo/mojito#1152

PR #1154, opened 2013-06-04
changes for v0.6.1

isao:
please chime in with edits, omissions, errors.


isao:
also here https://github.com/yahoo/mojito/wiki/Releasenotes0-6-1

PR #1155, opened 2013-06-05
Merging #develop into #next after 0.6.1 release

imalberto:
- Resolved conflicts
- Disabled the command jslint.js unit test which is causing the unit test suite to fail (command tests are no longer used in #next, and should be removed at some point - added a card to Trello)

Travis CI build results [here](https://travis-ci.org/yahoo/mojito/builds/7816388).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,8 +4,6 @@ version @VERSION@
 Notes
 -----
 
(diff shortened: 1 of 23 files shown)


isao:
+1

PR #1156, opened 2013-06-06
ignore /docs & save a 2M in mojito package tarballs

isao:


drewfish:
+1.  This also makes the mojito-0.6.1.tgz package go from 1.2M to 177k.

PR #1160, opened 2013-06-10
do not install/symlink "mojito" in $PATH

isao:
rm package.json "bin":{"mojito":"bin/mojito"}

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,9 +42,6 @@
         "docs",
         "lib"
     ],
-    "bin": {
-        "mojito": "bin/mojito"
-    },
     "engines": {
         "node": ">0.6",
         "npm": ">1.0"


caridy:
+1 (please ask @zhouyaoji to update docs if needed)

isao:
@caridy the [qs guide](http://developer.yahoo.com/cocktails/mojito/docs/getting_started/quickstart.html) and [cli reference](http://developer.yahoo.com/cocktails/mojito/docs/reference/mojito_cmdline.html) have been updated. I'll take another look before next release.

Folks who are using cli commands from mojito package like `node_modules/.bin/mojito {jslint,test,docs}` (for example in CI) will need to either:
1. install mojito-cli locally (i.e. as a devDep) _preferred_ and `node_modules/.bin/mojito {jslint,test,docs}` will continue to work
2. use `node_modules/mojito/bin/mojito` which is what SD does right now.

The latter case 2) will need to be updated to 1) or similar when the next phase-- cli code removal from mojito package-- is released.

caridy:
in 2) you mean `node_modules/mojito-cli/bin/mojito` right?

isao:
Ah, yes thanks Caridy. I should clarify. This PR will go in the next release, probably versioned `mojito@0.7.0`.
- `mojito` versions up to 0.6.1 have `node_modules/.bin/mojito` which points to `node_modules/mojito/bin/mojito`.
  - Some folks use either one path or the other to reach inside the mojito package for cli operations.
  - This PR removes the first (`.bin`) path, not the second.
  - The cli code will be removed from `mojito` proper soon with pull #1162. At that point the second path `node_modules/mojito/bin/mojito` will not be there either.
- `mojito-cli` (all versions) have `node_modules/.bin/mojito` which point to `node_modules/mojito-cli/bin/mojito` when installed locally with your app, as in the devDep case. 

So using `node_modules/.bin/mojito` in CI will work for _both_ old `mojito@"=<0.6.1"` by itself, or `mojito@">=0.7.0"` installed with `mojito-cli@">=0.0.4"` as a devDep.

(edited mojito@0.6.1 is last release with a symlinked "mojito" in `node_modules/.bin`)

caridy:
cool, got it!

PR #1161, opened 2013-06-10
Move carrier and device addons to ymojito

lzhan:
Need to update history file too.


caridy:
please, check the failure in travis, might be a false positive.

also, I think you have to remove the mocks from "mojito-test.js" file.

lzhan:
Good catch. Thanks. Updated. Will monitor travis result.

caridy:
+1

Issue #1159, opened 2013-06-08
not able to use mojito-data-addon if the mojit is lazy loaded via mojitProxy.invoke


ghost:
In my mojit I have two actions: "index" and "lazy", where "index" will invoke "lazy" from binder. The mojito-data-addon is included but in "lazy" action, when calling ac.data.set(), there is an error:

TypeError: Object #<Object> has no method 'set'

Apparently it is because mojito passing a `data` property in invoke command, and it make data addon failed to init itself.
https://github.com/yahoo/mojito/blob/develop/lib/app/autoload/mojit-proxy.client.js#L204
https://github.com/yahoo/mojito/blob/develop/lib/app/addons/ac/data.common.js#L25

I created a simple app the demo the issue:
https://git.corp.yahoo.com/yiwu/demo-lazyload-issue

caridy:
I was able to reproduce this. thanks for the repro case, I will try to land a fix asap.

PR #1163, opened 2013-06-10
adding rehydration process when using rpc tunnel

caridy:
Solves issue #1159: With this feature, any data defined thru `mojitProxy.data.set()` or `ac.data.set()` will travel from server to client and from client to server as part of the rehydration process of the mojit instance.
- [x] docs
- [x] tests


drewfish:
+1

PR #1165, opened 2013-06-11
Specify package depth level of node_modules dirs to load tests from

gomezd:
Specify package depth level of node_modules dirs to load tests from
Use the same depth config to get yui modules config.

When doing a `mojito test`, mojito will run all the tests in the app (or npm bundle package) as well as all the tests it finds recursively in the node_modules.

In some occasions it is desirable to run only the first level of tests (without going into node_modules), or specify the depth level (0=app, 1, 2, etc).

This change add the capability in the yui rs addon to specify the depth, as well as the -d option in the `test` command to specify it in the command line.

The default behavior of including all tests is kept for backwards compatibility when the option is not specified.

### DIFF
--- a/lib/app/addons/rs/yui.js
+++ b/lib/app/addons/rs/yui.js
@@ -209,7 +209,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
          *      those found in mojito (false).
          * @return {object} datastructure for configuring YUI
          */
-        getModulesConfig: function(env, justApp) {
+        getModulesConfig: function(env, justApp, depth) {
             var store = this.get('host'),
                 m,
                 mojit,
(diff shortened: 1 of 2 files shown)


caridy:
We plan to remove the recursive tests in favor of testing the current pkg only.

/ping @isao

gomezd:
You may still want to include some 'common tests' for example, that come from 'devDependencies' pkg in the first level of dependencies, for reusability.
Actually It would be really useful to filter by pkg, being able to decide which ones you include or not, but seems to be a larger change.

caridy:
`mojito test` will test unit tests, why should I test my dependencies? each of those dependencies are npm pkgs and they should be tested before releasing them. In other words, there is no point on testing dependencies when testing the units in the app.

gomezd:
Ageed. I didn't mean testing your dependencies but testing different apps with reusable testes that come from a dev dependency.
In particular i'm thinking of tests that can validate specific pieces of your app, say configs, file naming conventions for selectors, or things like that.
I also agree that this is not unit testing and i know you guys are thinking about this kind of tools So I'll let you decide :)

isao:
@gomezd thanks for the PR. @caridy is correct, running tests on stuff in node_modules was considered a bug, and I don't think a configurable depth is a safe workaround. Ideally these common components have their own tests and CI.

OTOH `mojito test mojit path/to/dep/tests` might work (you may need to symlink path). If you feel strongly about this please file an issue at [yahoo/mojito-cli-test](https://github.com/yahoo/mojito-cli-test/issues).

Thanks again, closing here.

PR #1170, opened 2013-06-17
rm lazyloadrpc test from flatfile list

lzhan:
The mojit has server affinity and will not work with flatfile. Remove the test from html5app test descriptor


caridy:
+1

PR #1173, opened 2013-06-18
Refactored router and dispatcher middleware to be more like Express.

imalberto:
- updated version to 0.8.x  to keep separate from #develop branch
- `routes.json` configuration is no longer contextualized - always default to "master"
- refactored `mojito-router` middleware 
- refactored `mojito-handler-dispatcher` middleware
- updated `ResourceStore.getRoutes()` to be not context aware
- refactored module `route-maker` to expand the route config
# todo
- [x] fix unit tests
- [x] fix functional tests
- [x] fix examples


lib/app/autoload/route-maker.common.js
```diff
@@ -131,6 +142,7 @@ YUI.add('mojito-route-maker', function(Y, NAME) {
             }
 
             if (typeof route.params !== 'object') {
+                Y.log('** Deprecated ** params: "name=Tom&age=23" is no longer supported', 'info');
```
caridy: should be `warn` instead of `info`, and make sure you also add `NAME` as the 3rd argument for better filtering.

lib/app/autoload/store.server.js
```diff
@@ -772,14 +772,17 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
-         * @param {object} ctx the context
-         * @return {object} routes
+         * @return {Object} routes
          */
-        getRoutes: function(ctx) {
+        getRoutes: function () {
             var appConfig,
                 cachedValue,
-                key = JSON.stringify(ctx || {}),
+                ctx = {}, // static context
```
caridy: not needed anymore

lib/app/autoload/store.server.js
```diff
@@ -772,14 +772,17 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+         *
          * @method getRoutes
-         * @param {object} ctx the context
-         * @return {object} routes
+         * @return {Object} routes
          */
-        getRoutes: function(ctx) {
+        getRoutes: function () {
             var appConfig,
                 cachedValue,
```
caridy, line 787: is this needed?

lib/dispatcher.js
```diff
@@ -0,0 +1,79 @@
...
+    The `command` should be attached to the request.
+
+    @public
+    @method dispatch
+    @param {http.ServerRequest} req
+        @param {Object} req.command the mojito command to be executed
+    @param {http.ServerResponse} res
+    @param {Function} next
+    **/
+    dispatch: function (req, res, next) {
```
caridy: This is not a middleware, it should not have a middleware signature. lets discuss the details offline.

lib/dispatcher.js
```diff
@@ -0,0 +1,79 @@
...
+
+        outputHandler.setLogger({ log: Y.log });
+
+        // storing the static app config as well as contextualized 
+        // app config per request
+        outputHandler.page.staticAppConfig = appConfig;
+        outputHandler.page.appConfig = store.getAppConfig(context);
+
+        // compute routes once per request
+        outputHandler.page.routes = store.getRoutes();
```
caridy, line 118: why do we need the routes attached to page anymore?

lib/mojito.js
```diff
@@ -113,6 +115,7 @@ Mojito.prototype._init = function (app) {
     options.context = { runtime: 'server' }; // TODO: use correct static context here
     options.mojitoRoot = __dirname;
     options.root = root;
+    console.log(process.cwd());
```
caridy: can we use `debug()` instead?

lib/router.js
```diff
@@ -0,0 +1,177 @@
...
+
+Example usage:
+
+    var express = require('express'),
+        mojito = require('mojito'),
+        app;
+
+    app = express();
+    ...
+    app.use(mojito.attachRoutes());
```
caridy: `attachRoutes` should NOT be a middleware because it will force to register the routes once the first request comes in.

lib/router.js
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        for (name in routes) {
+            if (routes.hasOwnProperty(name)) {
+                route = routes[name];
+                for (verb in route.verbs) {
+                    if (route.verbs.hasOwnProperty(verb)) {
+                        // debug('instaling handler: [' + verb + '] ' + name);
+                        verb = verb.toLowerCase();
+                        app[verb](new RegExp(route.ext_match), handler(name, route));
```
caridy: why are all routes regex?

lib/router.js
```diff
@@ -0,0 +1,177 @@
...
+                mojito.dispatch(req, res, next);
+            };
+        }
+
+        for (name in routes) {
+            if (routes.hasOwnProperty(name)) {
+                route = routes[name];
+                for (verb in route.verbs) {
+                    if (route.verbs.hasOwnProperty(verb)) {
+                        // debug('instaling handler: [' + verb + '] ' + name);
```
caridy: why are all `debug` statements commented out?

lib/router.js
```diff
@@ -0,0 +1,177 @@
...
+        routes = routeMaker.getComputedRoutes();
+
+        /**
+         * Install a request handler for a given route
+         *
+         * @param {String} name name of the route based on `routes.json` config
+         * @param {Object} route route configuration
+         * @return {Function} express.middleware
+         */
+        function handler(name, route) {
```
caridy: This should be exposed, so people can do this in their app.js:

```
app.get('/foo', mojito.dispatch('foo.index'));
```

which is going to be equivalent to do this in routes.json:

```
{
    "foo.index": {
         "path": "/foo",
         "verb": ["get"],
         "call": "foo.index"
    }
}
```

which also means, you will be able to do this in your templates: `{{#linkTo "foo.index"}}`, although this is not part of this PR, but the next one.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "mojito",
-    "version": "0.6.1",
+    "version": "0.10.1",
```
caridy: why version 10?

caridy:
finished the first pass, let's talk offline for more details.

lib/router.js
```diff
@@ -0,0 +1,220 @@
...
+                        debug('installing handler: [' + verb + ' ' + route.path + '] ' + name);
+                        verb = verb.toLowerCase();
+
+                        args = [route.path];
+
+                        if (route.params) {
+                            args.push(mid(route.params));
+                        }
+
+                        args.push(dispatcher.dispatch(route));
```
caridy: `route.call`

lib/router.js
```diff
@@ -0,0 +1,220 @@
...
+            if (routes.hasOwnProperty(name)) {
+                route = routes[name];
+                for (verb in route.verbs) {
+                    if (route.verbs.hasOwnProperty(verb)) {
+                        debug('installing handler: [' + verb + ' ' + route.path + '] ' + name);
+                        verb = verb.toLowerCase();
+
+                        args = [route.path];
+
+                        if (route.params) {
```
caridy: nice! remember to add a BIG warn for deprecation.

lib/router.js
```diff
@@ -0,0 +1,220 @@
...
+            var o = params,
+                copy;
+
+            return function (req, res, next) {
+                var cloned,
+                    key,
+                    query;
+
+                if (!copy) {
+                    // easier way ?
```
caridy: `var util = require('./path/to/lib/util');`

that file should hold any utility method needed for the server side (hence anything needed in a node module), feel free to add `copy` method in there.

lib/router.js
```diff
@@ -0,0 +1,220 @@
...
+                    copy = req.app.mojito.Y.mojito.util.copy;
+                }
+
+                // TODO: find a more appropriate namespace for this
+                req.routeMatch = req.routeMatch || {};
+                req.routeMatch.query = req.routeMatch.query || {};
+                query = req.routeMatch.query;
+
+                cloned = copy(o);
+                debug(inspect(cloned));
```
caridy: remove!

lib/router.js
```diff
@@ -0,0 +1,220 @@
...
+                // add the fixed params to the query object if they are not
+                // there
+                for (key in cloned) {
+                    if (cloned.hasOwnProperty(key) && !query[key]) {
+                        debug('copying params from routes configuration: %s', key);
+                        query[key] = cloned[key];
```
caridy: I thought those params defined per route in routes.json were intended to be just attached as req.params under the hood. In which case all these cloning and stuff are not very useful.

lib/app/autoload/route-maker.common.js
```diff
@@ -156,33 +170,33 @@ YUI.add('mojito-route-maker', function(Y, NAME) {
...
-                    route.call = route.call.replace(regex, template[key]);
-                } else {
-                    route.params[key] = val;
-                }
-            }
-            cnt += 1;
-        });
-        return route;
-    };
+    // doCallReplacement = function (route, uri) {
```
caridy: if we don't need it, remove it :)

caridy:
A lot of comments, let's do another pass tomorrow

lib/dispatcher.js
```diff
@@ -0,0 +1,132 @@
...
+        outputHandler = new OutputHandler(req, res, next);
+
+        outputHandler.setLogger({ log: Y.log });
+
+        // storing the static app config as well as contextualized 
+        // app config per request
+        outputHandler.page.staticAppConfig = appConfig;
+        outputHandler.page.appConfig = store.getAppConfig(context);
+        // routes are not contextualized anymore
+        outputHandler.page.routes = store.getRoutes();
```
imalberto, line 118: For now, this is required by `ac.config` and `ac.url`

lib/dispatcher.js
```diff
@@ -0,0 +1,132 @@
...
+        var my = this,
+            matchCall = [];
+
+        matchCall[0] = call.split('.');
+        matchCall[1] = matchCall[0].pop();
+        matchCall[0] = matchCall[0].join('.');
+
+        return function (req, res, next) {
+            var command = { instance: { } },
+                mojito = (req.app && req.app.mojito) || null,
```
caridy: not need to use `|| null` here.

lib/router.js
```diff
@@ -0,0 +1,237 @@
...
+            var o = params;
+
+            return function (req, res, next) {
+                var cloned,
+                    k,
+                    query;
+
+                // TODO: find a more appropriate namespace for this
+                req.routeMatch = req.routeMatch || {};
+                req.routeMatch.query = req.routeMatch.query || {};
```
caridy: how is this used across mojito? who uses those query params added manually from routes.json?

caridy:
we are getting close :)

lib/dispatcher.js
```diff
@@ -0,0 +1,132 @@
...
+
+    @param {String} call the name of the mojit and action concatenated
+    by a period (e.g. "admin.help" will execute mojit instance `admin` with
+    action `help`.
+    **/
+    dispatch: function (call) {
+        var my = this,
+            matchCall = [];
+
+        matchCall[0] = call.split('.');
```
caridy: this is a little confusing code. Overloading `matchCall[0]` is confusing. Either you add a good note, or change the code with more variables to avoid the overloading.

lib/dispatcher.js
```diff
@@ -0,0 +1,132 @@
...
+    @protected
+    @method handleRequest 
+    @param {http.ServerRequest} req
+        @param {Object} req.command the mojito command to be executed
+    @param {http.ServerResponse} res
+    @param {Function} next
+    **/
+    handleRequest: function (req, res, next) {
+
+        var command = req.command || null,
```
caridy: `|| null` is really not needed.

lib/router.js
```diff
@@ -0,0 +1,216 @@
...
+        function registerRoutes(routes) {
+            var name,
+                route,
+                verb;
+            for (name in routes) {
+                if (routes.hasOwnProperty(name)) {
+                    route = routes[name];
+                    for (verb in route.verbs) {
+                        if (route.verbs.hasOwnProperty(verb)) {
+                            debug('[%s %s] installing handler', name, route.path);
```
caridy: I don't think the name will be relevant! the `call` and `path` are!

lib/router.js
```diff
@@ -0,0 +1,216 @@
...
+
+
+        /**
+         * Adds a route for handlng tunnel requests from clients
+         *
+         * For now, this is combined as a single call to `attachRoutes`.
+         *
+         * TODO: should this be a separate call that `app.js` should do ?
+         */
+        function registerTunnelRpc() {
```
caridy: I don't think this belongs to here! Tunnel should be either: 

a) set in app.js as a middleware with path: e.g.: `app.post('/tunnel', tunnelMiddleware);` or
b) mojito.js can do it automatically which is not good because they devs will not have a way to change it or modify it. 

but it is definitely not related to router.js

lib/router.js
```diff
@@ -0,0 +1,216 @@
...
+                            verb = verb.toLowerCase();
+
+                            app[verb](route.path, dispatcher.dispatch(route.call));
+                        }
+                    }
+                }
+            }
+        }
+
+        registerTunnelRpc();
```
caridy: same as above!

caridy:
done with the pass!

caridy:
+1

caridy:
+1

lib/dispatcher.js
```diff
@@ -0,0 +1,136 @@
...
+            outputHandler.hook = req.hook || {};
+            Y.mojito.hooks.enableHookGroup(outputHandler.hook, 'mojito-perf');
+        }
+        // HookSystem::EndBlock
+
+        // HookSystem::StartBlock
+        // Y.mojito.hooks.hook('AppDispatch', outputHandler.hook, req, res);
+        // HookSystem::EndBlock
+
+        Y.mojito.Dispatcher.init(store).dispatch(command, outputHandler);
```
drewfish, line 133: Do we need to initialized the dispatcher for every request?

caridy, line 133: hopefully not! lets double check that to see how expensive is this.

aljimenez:
Is there a reason why routes are no longer contextualized? Currently all routes in routes.json are getting exposed to the client.

caridy:
Yes, the reason is perf. For the routes to be contextualize, we need to process them per request, this is an expensive operation, and will now allow us to rely on the internal cache implemented by express. Also, the fact that routes are normally not changed per context.

On top of that, this doesn't necessarily means you cannot limit what you expose to the client side. In fact, you have full control of what is exposed, by using express-map's filter mechanism to expose the list of routes you want, in case you don't want them all to be exposed.

PR #1182, opened 2013-06-24
keep using arrow 0.0.77 and find free port for phantomjs at runtime

lzhan:


isao:
suggested fixes:
1. portfinder.getPort's callback doesn't check for errors
1. will there always be only one instance of phantomjs? if not you should add the port number to line 238

minor:
1. needs de-linting
1. maybe it's just me, but I find the '/../' in the paths confusing and messy. can we start referring to paths from `MOJITOLIB` moving forward? i.e. path.join(MOJITOLIB, '/node_modules/phantomjs')

isao:
I'd throw rather than process.exit() for 2757556d91d9feaa364be6fcb6605423a06c6407 but it's fine. thanks for fixing all that lint!

+1

PR #1186, opened 2013-06-27
[doc] Removed note on globally installing Mojito, which is now impossible.

zhouyaoji:


isao:
minor note, it's still possible, there's just no cli unless you install the mojito-cli package

PR #1187, opened 2013-07-01
Added loadConfigs method and event to the store and rs addons.

gomezd:
Some RS addons get the host (store) configuration when they are initialized.
However the same addons can potentially modify the app/host configurations while discovering resources in the filesystem (i.e. dynamically adding dimensions or configurations from resource bundles), leaving them with incomplete or inconsistent data.
For this we have added a public method "loadConfigs" to the resource store and the addons, and we've made the store to fire a "loadConfigs" event when it's configs are loaded/changed, so addons can subscribe to and update their configs if they need to.


isao:
Thanks for this PR. I like having the loadConfig() function rather than inlined in the store init, and the new event.

Can you give some use cases of addons that legitimately need to modify the configs? I'd like to understand this a bit more. My initial reaction is that should not be allowed or accommodated... If this is needed/valuable for your apps, but not more generally, perhaps some of it can be moved to an external package.

I'm also concerned about the performance impact. The ycb stuff, for now, is expensive and this PR doubles the cost.  Also, does this result in fs or syscalls at request time?

Lastly, a test case would be nice, and please remove the travis branch config...

Thanks.

gomezd:
We have implemented the changes you requested. Please review.
Thanks

isao:
lgtm, thank you @gomezd and @jarnoux

lib/app/autoload/store.server.js
```diff
@@ -338,6 +335,15 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
         },
 
 
+        loadConfigs: function () {
```
drewfish, line 338: Missing documentation.

PR #1189, opened 2013-07-10
Use any glob pattern to run test subsets.

gomezd:
Currently there is a problem trying to run partial tests like this:
 ./run.js test -f --path func --descriptor examples/quickstartguide/*_descriptor.json

globs different than '**/8_descriptor.json' are ignored and only full descripto paths are taken into account.

Whit this change we can use either the full path or any glob to match the descriptors, as the command help says:
"--descriptor <value>    which descriptor to run. filename (or glob) relative to --path", which makes easier and less time consuming to run/validate individual tests.

### DIFF
--- a/tests/run.js
+++ b/tests/run.js
@@ -400,14 +400,9 @@ function runMojitoApp(app, cliOptions, basePath, port, params, callback) {
 
 function runFuncAppTests(cmd, callback) {
     var descriptor = cmd.descriptor || '**/*_descriptor.json',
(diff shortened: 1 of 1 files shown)


isao:
+1

lzhan:
It should work fine for both npm and ynpm env. It tries npm install first, if it is not successful, then tries ynpm.

isao:
thanks for the clarification @lzhan. my +1 stands :)

PR #1191, opened 2013-07-12
First pass at feature parity for url add-on based on app.routes.

imalberto:
CHANGELOG:
- dispatcher has been modified to annotate the middleware with mojito specific attributes
- `route-maker.find()` and `route-maker.make()` methods have been refactored to work with `app.routes`
- `route-maker.getComputedRoutes()` provides a temporary BC for the mojito-client runtime to work, until the new client router lands
- applications can now register route like: `app.get('/foo', mojito.dispatch('bar.foo'));`,
- `Y.mojito.util.copy()` now support regex and function types

TODO:
- [x] fix unit tests
- [x] fix functional tests
- [x] fix examples


lib/dispatcher.js
```diff
@@ -56,6 +63,29 @@ module.exports = {
...
+            if (mojito.routes && Object.keys(mojito.routes).length === 0) {
+                debug('building up route cache for first time');
+                methods.forEach(function (method) {
+                    var routes,
+                        cache = [];
+
+                    routes = req.app.routes[method];
+                    if (routes && routes.length > 0) {
+                        routes.forEach(function (route) {
+                            cache.push({
```
caridy: this should only happens if annotations exists, `extractAnnotations(route)`, other routes are not relevant for mojito.

lib/dispatcher.js
```diff
@@ -47,7 +48,13 @@ module.exports = {
         matchCall[1] = splits.pop();
         matchCall[0] = splits.join('.');
 
-        return function (req, res, next) {
+        function extractAnnotations(route) {
+            var fn = route && route.callbacks && route.callbacks.length &&
```
caridy, line 54: in the future we might want to support a more generic annotation mechanism, walking all `callbacks` and producing an array of annotations, for now looking at the last one is just fine.

caridy:
need more time to review the addon, too complex.

lib/app/addons/ac/deploy.server.js
```diff
@@ -116,7 +117,13 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
                 clientConfig.pathToRoot = pathToRoot;
             }
 
-            clientConfig.routes = store.getRoutes(contextClient);
+            // NOTE:
```
caridy: what is temporary here?

lib/app/addons/ac/deploy.server.js
```diff
@@ -116,7 +117,13 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
...
+            // This is temporary until the new client router lands.
+            // clientConfig.routes = store.getRoutes(contextClient);
+            //
+            // TODO: pageRoutes probably should be sanitized before shipping to
```
caridy: sanitized how? whatever we need to do, we should do it upfront as a one time operation, so when it comes to serialize and send, we have everything ready.

lib/dispatcher.js
```diff
@@ -56,6 +66,33 @@ module.exports = {
                 return next(new Error('app.mojito does not exist.'));
             }
 
+            if (mojito.routes && Object.keys(mojito.routes).length === 0) {
```
caridy: this condition could be always true if there are no routes. I propose to have mojito.routes as undefined, then once we build the cache, we set it to an object, and that's the condition.

lib/dispatcher.js
```diff
@@ -56,6 +66,33 @@ module.exports = {
...
+                debug('building up route cache for first time');
```
caridy: remove " for first time".

lib/dispatcher.js
```diff
@@ -115,7 +160,8 @@ module.exports = {
         outputHandler.page.staticAppConfig = appConfig;
         outputHandler.page.appConfig = store.getAppConfig(context);
         // routes are not contextualized anymore
-        outputHandler.page.routes = store.getRoutes();
+        // outputHandler.page.routes = store.getRoutes();
```
caridy: just remove it

lib/app/autoload/route-maker.common.js
```diff
@@ -6,253 +6,268 @@
...
 
             uri = route.path;
 
+            keys = [];
+            Y.Array.each(route.keys, function (key) {
+                keys.push(key.name);
+            });
+
+            // handle left over params by appending them to query string
             for (k in params) {
```
caridy: there is a bug here, probably an existing bug. If I have two keys: `foo` and `foobar`, and you try to replace `:foo` first, then it will invalidate `:foobar` as well. The solution is to sort the lookup array so you try to replace the longest keys first.

lib/app/autoload/route-maker.common.js
```diff
@@ -6,253 +6,268 @@
...
+                               params. `query` might have querystring portion
+                               portion, in which case they have priority.
+        @return {Object} route object matching the query, or null
+        **/
         make: function(query, verb, params) {
             // Y.log('make(' + query + ', ' + verb + ')', 'debug', NAME);
-
-            var parts = query.split('?'),
+            var my = this,
+                parts = query.split('?'),
```
caridy, line 261: are the existing querystrings irrelevant? I don't see the second part of `parts` being used.

lib/app/autoload/route-maker.common.js
```diff
@@ -267,173 +282,51 @@ YUI.add('mojito-route-maker', function(Y, NAME) {
...
-
-            // Add the fixed params to a query obj if they are not there
-            for (i in match.params) {
-                if (match.params.hasOwnProperty(i) && !match.query[i]) {
-                    match.query[i] = match.params[i];
-                }
-            }
-
-            return match;
+            return route;
```
caridy: This route is, I presume, the new route object from cache, which is probably different from what they get today. Can we transform it?

lib/app/autoload/route-maker.common.js
```diff
@@ -6,253 +6,268 @@
...
+    Convert the routes from express format to the "old" mojito format
+    for BC. This is a temporary solution until the new client router 
+    lands.
+
+    @method convertRoutes
+    @protected
+    @param {Object} appRoutes routes object from express
+    @return {Object} 
+    **/
+    function convertRoutes(appRoutes) {
```
caridy: Can you sync up with Julien to see if what are they doing with the routes? If they are not using anything fancy, we can remove this.

lib/app/autoload/route-maker.common.js
imalberto, line 261: Thanks, added the check here https://github.com/yahoo/mojito/pull/1191/files#L1R269 . Querystring as part of the query will take priority.

caridy:
Ok, let's take this for a spin.

+1

lib/app/autoload/route-maker.common.js
```diff
@@ -6,254 +6,299 @@
...
+    **/
+    function convertAppRoute(route) {
+        var template = {},
+            verb = route.method.toUpperCase(),
+            verbs = {},
+            name;
+
+        verbs[verb] = true;
+
+        name = route.path + '.' + (route.dispatch && route.dispatch.call);
```
imalberto, line 93: Prepending the "path" to the "call", so that this can be supported without having name conflict:

```
// In this case, "/foo" and "/bar" are sharing the "admin.foo" controller
app.get('/foo', mojito.dispatch('admin.foo'));
app.get('/bar', mojito.dispatch('admin.foo'));
```

lib/app/autoload/util.common.js
```diff
@@ -169,13 +169,20 @@ YUI.add('mojito-util', function(Y, NAME) {
             var newObj,
                 key,
                 len,
+                type,
                 copy = Y.mojito.util.copy;
 
             if (!oldObj || typeof oldObj !== 'object') {
```
imalberto, line 175: Avoiding an invocation of `Y.Lang.type()` if it is not required.

examples/developer-guide/configure_routing/mojits/RoutingMojit/controller.server.js
```diff
@@ -14,21 +14,34 @@ YUI.add('RoutingMojit', function(Y, NAME) {
...
             routes = ac.config.getRoutes();
+
+		// Mojito no longer support more than one VERB per route entry
+		// The route.name is now the route.path + '.' + route.call
         if (action === "index" && path === "/") {
             name = "root_route";
-            routes.root_route.verbs.forEach(function(n) {
-                methods += n + ", ";
-            });
+			// routes.root_route.verbs.forEach(function(n) {
```
caridy: remove any old code!

lib/app/addons/ac/config.common.js
```diff
@@ -95,7 +95,7 @@ YUI.add('mojito-config-addon', function(Y, NAME) {
          * @return {object} the routes
          */
         getRoutes: function() {
-            return this._page.routes;
+            return (new Y.mojito.RouteMaker(this._page.routes)).getComputedRoutes();
```
caridy: can we cache this? one per request? probably attaching it to `_page` object the first time it is used?

lib/app/autoload/route-maker.common.js
```diff
@@ -6,254 +6,299 @@
...
+    var CACHE = { routes: { }, computedRoutes: { } };
 
-    var doCallReplacement,
-        CACHE = { routes: {} }; // cache for _routes
+    /**
+    Given a route with parametrized variable, will attempt to resolve
+    it based on a set of parameters.
 
+    @method resolveParams
+    @protected
```
caridy: this is really private I think

lib/app/autoload/route-maker.common.js
```diff
@@ -6,254 +6,299 @@
...
+    **/
+    function doCallReplacement(route, uri) {
+        var uriParts = uri.split('/'),
+            pathParts = route.path.split('/'),
+            index = 0,
+            // route.dispatch.params is assumed to be already set by
+            // the cache builder
+            params = route.dispatch.params;
+
+        pathParts.forEach(function (pathPart) {
```
caridy: this will not work in old browsers, this is a common affinity file. Use `Y.Object.each` instead.

lib/app/autoload/route-maker.common.js
caridy, line 93: Is the name used to hash this into a collection? at this method, name is just another member, I don't see why it has to be unique.

caridy, line 93: I see `convertAppRoutes` is hashing it using name. I don't think this is good because internally you will never reference that by a similar name. It should be an array, just like it is in express.

lib/app/autoload/route-maker.common.js
```diff
@@ -6,254 +6,299 @@
...
+
+        template = {
+            path: route.path,
+            call: route.dispatch && route.dispatch.call,
+            // for the name to be unique per route entry,
+            // concatenate path + '.' + call
+            name: name,
+            verbs: verbs,
+            params: route.dispatch.params,
+            regex: {},
```
caridy, line 103: why not using regex as regex? who is using this structure aside from us?

lib/app/autoload/route-maker.common.js
imalberto, line 93: I think the original thought of setting the "name" to just be "call" is good enough.

Making the name unique is more of a "maintain backward compatibility" as previously, the "name" was provided by developer in `routes.json`, and by convention, they were unique across all routes.

Turns out that mojito-shaker uses the name property to group rollups per route entry. We can revert it to being set to "call" only, and document that in order to ensure uniqueness, the caller should generate and associate a named identifier to it.

lib/app/autoload/route-maker.common.js
imalberto, line 103: In Mojito today, the semantic and syntax for `regex` is different from `ext_match`.

I think once we remove the transformation later in a not-so-distant future, this won't be needed anymore. For now, might be better to keep the route configuration format the same as what it is today, for BC sake.

imalberto:
@caridy Added some comments.

caridy:
Yeah, I still need more time to finish the review.

lib/app/autoload/route-maker.common.js
imalberto: Fixed.

lib/app/autoload/route-maker.common.js
imalberto: Fixed.

lib/app/addons/ac/config.common.js
imalberto: Fixed. Adding it to `this._page.computedRoutes`.

examples/developer-guide/configure_routing/mojits/RoutingMojit/controller.server.js
imalberto: Gone !

lib/dispatcher.js
```diff
@@ -18,6 +18,7 @@ Given a `command`, dispatches it to the Mojito engine.
 
 var liburl = require('url'),
     debug = require('debug')('mojito:dispatcher'),
+    methods = require('methods'),
```
isao: this module isn't listed as a dependency in package.json. Did you mean this?

```
methods = require('express/node_modules/methods'),
```

imalberto: Its added here: https://github.com/yahoo/mojito/pull/1191/files#L10R24

isao: thx, my mistake.. I had started rebasing onto develop, and.. :/

lib/dispatcher.js
```diff
@@ -56,6 +66,34 @@ module.exports = {
                 return next(new Error('app.mojito does not exist.'));
             }
 
+            if (!mojito.routes) {
+                debug('building up route cache');
+                mojito.routes = {};
+                methods.forEach(function (method) {
```
isao: The package "methods" exports an array of HTTP methods, including a bunch of esoteric ones. But, unless I'm missing something, it doesn't seem to be needed-- this line iterates over the methods array, but is only actionable if it matches a key in `req.app.routes`.

Would it be better to just iterate over `Object.keys(req.app.routes)`? That'd generally be just `['get', 'post']` for web apps

caridy: Yes, @isao is right, you should write this routine by iterating over the methods defined by `Object.keys(req.app.routes)`, not need to use `methods`.

caridy:
ok, we can take this for a spin, and we can re-iterate on top of this.

+1 after fixing the stuff that @isao mentioned.

lib/dispatcher.js
imalberto: Thanks ! Adding diff here for reference: https://github.com/imalberto/mojito/commit/74fe35abf88af0155871a7c0d6eac37c8fec8276

Issue #1107, opened 2013-05-10
collision detection for resource versions is naive -- wrong choice for YUI modules


drewfish:
The resource store detects when a resource version is provided by multiple packages, and tries to use the one in the "lowest" package (one closest to app, possibly the app itself).  That part works fine.

However, `rs.yui.getModulesConfig()` ignores this resolution, and ends up using the code in the highest/last package, contrary to the intentions of the store.

Reproduction case: https://github.com/drewfish/mojito-issue1107

PR #1195, opened 2013-07-15
Lazy resource store

jarnoux:
This pull request implements an optional lazy resolution of resources that potentially makes the resource store creation independent of the number of dimensions and linear with the number of resources.
It also introduces the ability for plugins to alter configurations


gomezd:
Just a note from the description of this PR: "It also introduces the ability for plugins to alter configurations". This is part of a previous PR:
https://github.com/yahoo/mojito/pull/1187

isao:
This looks good to me. I think some of these changes should be the new default behavior (not a separate code path + config), but drewfish is back tomorrow. We'll get his input before proceeding.

drewfish:
I tried this approach and found that it caused mojito to be slower. The interpretation is that `_mojitRVs` is so big that it causes a large GC walk which slows things down. (Or, having the large `_mojitRVs` causes other memory pressure, trigger more GC walks or something like that.)

I'll look at the details later.

jarnoux:
I see, did you try stringifying the mojitRVs like you do for the cache?

jarnoux:
also how much slower?

drewfish:
OK, I found the experiment I did earlier. It's from March 21st this year.

Here's what I did:
- Ran a search-app-news search results page and captured the calls to `store.expandInstanceForEnv()`.
- Wrote a script that loads those calls and replays them, looking for GC events.
- Tweaked the RS to do two types of changes:  calculate `_mojitoDetails` at startup or on demand (AKA lazily), store the `_mojitoDetails` JSON.stringify()ed or as a raw datastructure.
- Ran the script with the four experiments (2 x 2).

Alas, it only captured the GC events of the scripts, not the total runtime. Here are the results:
- lazy/stringified - 9 GC events
- lazy/raw - 9 GC events
- startup/stringified - 5 GC events
- startup/raw - 6 GC events

From this I surmised that keeping the source data (`_appRVs` and `_mojitRVs`) is a lot of data which causes enough memory pressure to cause more GC runs (and more data to walk on each run).

(I had written all these results into a trello card but alas I can't find it.)

caridy:
Trello Card 432:

https://trello.com/c/Cz5hCgj6/432-search-perf-0-gc-profiling

drewfish:
Just moved the card to the right board, so it's now located at:
https://trello.com/c/Cz5hCgj6/256-search-perf-0-gc-profiling

isao:
this test would be interesting to repeat in node 0.10.x. Before then node would trigger GC "whenever the event loop was idle." But the behavior was removed because in practice, V8 was better at deciding when to GC. http://blog.nodejs.org/2013/03/11/node-v0-10-0-stable/ (search for "idle garbage collection")

I think the _count_ of GC events would not mean GC was required more often, for pre 0.10 versions..

drewfish:
The `.chosen` for resource versions was added as part of pull #1108 (to fix issue #1107) so we should be careful to preserve that.

lib/app/autoload/store.server.js
```diff
@@ -191,6 +192,10 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._appPkg = null;    // metadata about the applicaions's NPM package
             this._specPaths = {};   // spec name: full path
 
+            // Let's try to be more efficient
+            this._smarterDetails = {}; // MojitType: selector: environment: non-stringified details
```
drewfish: This structure is useful for the new algorithm, but should be called something else. It contains resource versions and is used by `_resolveVersions()` so perhaps it should be called `_mojitDetailRVs`.

lib/app/autoload/store.server.js
```diff
@@ -88,7 +88,8 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
         },
         // which addon subtypes are app-level
         ADDON_SUBTYPES_APPLEVEL = {
-            'rs': true
+            'rs': true,
+            'shaker': true
```
drewfish: I'm not sure why this is here. I can't think why shaker-specific things need to be in the mojito library.

lib/app/addons/rs/yui.js
```diff
@@ -229,7 +229,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
             mojits.push('shared');
             for (m = 0; m < mojits.length; m += 1) {
                 mojit = mojits[m];
-                ress = store.getResourceVersions({ mojit: mojit, chosen: true });
+                ress = store.getResourceVersions({ mojit: mojit });
```
drewfish, line 236: Dropping the `chosen:true` filter is fine as long as this PR doesn't regress issue #1107 .

lib/app/autoload/store.server.js
```diff
@@ -426,8 +429,8 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          */
         optimizeForEnvironment: function() {
             this._packagesVisited = {};
-            this._appRVs = [];
-            this._mojitRVs = {};
+            // this._appRVs = [];
```
drewfish: This needs to be guarded against staticAppConfig.resourceStore.lazyResolve. If we're resolving at startup we _do_ want to clear these members.

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1496,137 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
-        resolveResourceVersions: function() {
-            var p, poslKey, posl, posls = {},
-                e, env, envs = [ 'client', 'server' ],
-                affinities, selectors, sourceBase,
-                type,
-                ress,
-                s,
-                mojitRes;
+        resolveResourceVersions: function () {
+            var crtType,
```
drewfish, line 1515: `crtType` is probably better spelled as `mojitType` since that what it holds.

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1496,137 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+                resources,
+                r,
+                crtRes,
```
drewfish, line 1518: Similarly, why not just `res`? (Why is "crt" important? What info does it convey.)

jarnoux, line 1518: I use the "crt" prefix to denote it's an iterator in a loop and points to the "current" resource

lib/app/autoload/store.server.js
```diff
@@ -191,6 +192,9 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._appPkg = null;    // metadata about the applicaions's NPM package
             this._specPaths = {};   // spec name: full path
 
+            this._mojitDetailRVs = {}; // MojitType: selector: environment: non-stringified details
+            this._mojitDetailRVsCache = {};
```
drewfish: I think this is serving the same purpose as `_mojitDetails` so we should use that instead.  (The current `_mojitDetailsRVsCache` is holding the _resolved_ mojit details, not just the individual RVs, so perhaps it should be called `_mojitDetailsCache` (or perhaps just `_mojitDetails`).)

lib/app/autoload/store.server.js
```diff
@@ -191,6 +192,9 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._appPkg = null;    // metadata about the applicaions's NPM package
             this._specPaths = {};   // spec name: full path
 
+            this._mojitDetailRVs = {}; // MojitType: selector: environment: non-stringified details
```
drewfish: To be consistent with other nomenclature, the comment should probably be `MojitType: selector: affinity: non-stringified details` instead.

lib/app/autoload/store.server.js
jarnoux: that's correct

lib/app/autoload/store.server.js
jarnoux: we made shaker a plugin.base and we want to be able to have addons for it

lib/app/addons/rs/yui.js
jarnoux, line 236: fixed with sorting, see: https://github.com/gomezd/mojito/commit/145ebaa8178bbfed3120411162e1aa58322544a6

lib/app/autoload/store.server.js
drewfish: That's fine, it's just that there's nothing in the mojito code base today that has anything to do with shaker, so it's a little strange to add this. Having resources with type:addon,subtype:shaker can be done without adding this line to mojito.

lib/app/autoload/store.server.js
```diff
@@ -187,10 +188,12 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
             this._packagesVisited = {}; // package@version: path
             this._appRVs    = [];   // array of resource versions
             this._mojitRVs  = {};   // mojitType: array of resource versions
-            this._mojitDetails = {};  // mojitType: env-POSL: JSONified details
             this._appPkg = null;    // metadata about the applicaions's NPM package
             this._specPaths = {};   // spec name: full path
 
+            this._mojitDetails = {}; // mojitType: selector: environment: non-stringified details
+            this._mojitDetailsCache = {}; // mojitType+poslString+env : resolved resources
```
drewfish: For both of these "environment" is really "affinity" no?  If so it'd be clearer to say "affinity" instead.

lib/app/autoload/store.server.js
```diff
@@ -228,6 +231,15 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
         //====================================================================
         // PUBLIC METHODS
 
+        loadConfigs: function () {
```
drewfish, line 241: BTW, `loadConfigs()` is missing documentation. (Yeah, I know this PR just moved it).

lib/app/autoload/store.server.js
```diff
@@ -426,11 +428,12 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
          */
         optimizeForEnvironment: function() {
             this._packagesVisited = {};
-            this._appRVs = [];
-            this._mojitRVs = {};
+            if (!this._appConfigStatic.resourceStore || !this._appConfigStatic.resourceStore.lazyResolve) {
```
drewfish: Hmmm... I wonder if we want to stash this condition into `this.lazyResolve` or something like that.

lib/app/autoload/store.server.js
```diff
@@ -605,14 +608,17 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 // captured in the POSL.
                 cacheKey = JSON.stringify([env, posl, ctx.lang, mojitType]),
                 cacheValue = this._getMojitTypeDetailsCache[cacheKey],
-                details;
+                details,
+                Details;
```
drewfish: Uggg... having two variables that only differ by case seems to be a recipe for confusion/bugs. What does each variable hold? Is there a better way of describing each of them?

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1496,137 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
-         *      selector,
-         *      affinity (env or "common").
-         *
-         * @method resolveResourceVersions
-         * @return {nothing}
+         * For each resource, stuff it in a map mojitType->selector->affinity
+         * that will make it easy to retrieve it later. Then if the lazyResolve config
+         * is not set, go through all the possible combinations and resolve all
+         * resource versions.
+         * @return {[type]} [description]
```
drewfish: I think yuidoc will need a `@method resolveResourceVersions` in order to generate documentation for this method.  (Yuidoc doesn't parse javascript -- it only looks at comments.)

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1496,137 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+
+                    for (e = environments.length - 1; e >= 0; e--) { // 2
+                        currentResources = (this._mojitDetails[currentType][currentSelector] &&
+                            this._mojitDetails[currentType][currentSelector][environments[e]]) || [];
+
+                        // remember each resource we find for this posl, type and environment
+                        for (r = 0; r < currentResources.length; r++) { // ~3
+                            // prioritize: we prefer shallower resources
+                            if (!resolvedResources[currentResources[r].id] ||
+                                    resolvedResources[currentResources[r].id].source.pkg.depth >= currentResources[r].source.pkg.depth) {
```
drewfish: Can we use the same "sort resources by depth" trick here too?

drewfish:
Looks like travis build failed due to environmental issues (bad network) and not anything to do with code quality.

lib/app/autoload/store.server.js
jarnoux: yup

lib/app/autoload/store.server.js
jarnoux: dead code

lib/app/autoload/store.server.js
jarnoux: yup: https://github.com/gomezd/mojito/commit/448fcd2e52a0a0403fd8e6b2bc593592228c46b2

lib/app/autoload/store.server.js
jarnoux: semantically, it is environment: a request has an environment and goes there to look for resources which have affinities, but are indexed by environment.

drewfish: "Environment" is traditionally (in mojito code and docs) as things like "development", "production", "staging", etc. "Runtime" is used to describe the client vs. server. "Affinity" is used to describe which runtime a particular piece of code is intended for.

Unless there's a strongly compelling reason to use _different_ semantics here, it is strongly recommended to stay with the existing semantics. (If we want to change it, we should change it _everywhere_: all code and documentation.) That work (and discussion with the whole mojito team about the value of that work) is beyond the scope of this pull request, so please stick to existing semantics. Consistency aids clarity.

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1505,135 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
-         *      selector,
-         *      affinity (env or "common").
-         *
+         * For each resource, stuff it in a map mojitType->selector->affinity
+         * that will make it easy to retrieve it later. Then if the lazyResolve config
+         * is not set, go through all the possible combinations and resolve all
+         * resource versions.
          * @method resolveResourceVersions
-         * @return {nothing}
+         * @return {[type]} [description]
```
drewfish: Hmmm... this `@return` is confusing. What is the intended return value of this `resolveResourceVersions()` method?

lib/app/autoload/store.server.js
drewfish: Oh hmm... excepts a lot of places in the resource store "env"/"environment" is used for runtime (that is "client" or "server"). However, each resource has an "affinity" ("client", "server", or "common") for a particular "environment". If a field contains the affinity of a resource it should be labelled as "affinity".

lib/app/autoload/store.server.js
drewfish, line 1518: I've never seen that convention before, but I can see the value of it.

lib/app/autoload/store.server.js
jarnoux: nothing, dead code

lib/app/addons/rs/yui.js
drewfish, line 236: Yep, sorting works well.

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1505,134 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
                     }
                 }
             }
+
         },
 
+        /**
+         * Find a mojit details in the resource store at runtime
+         * @param  {String} type the mojit type
+         * @param  {String} env  the environment: {'client', 'server', 'common'}
```
drewfish: I think "environment" here is just "client" or "server".

lib/app/autoload/store.server.js
```diff
@@ -1490,83 +1505,134 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+                currentResources,
+                currentResourcesLen,
+                t,
+                currentType,
+                types = [type, 'shared'],
+                e,
+                affinities = [env, 'common'],
+                poslString = posl.toString(),
+                resolvedResources = {},
+                mojitRes = this.getResourceVersions({type: 'mojit', name: type, selector: '*'})[0],
```
drewfish: Might want to move this below the cache lookup, so that we don't pay the cost for the `getResourceVersions()` call if there is a cache hit.

drewfish:
I've confirmed that this (as of 8e6fc76) doesn't regress issue #1107.

drewfish:
+1

drewfish:
Looks like travis build failed due to environmental issues (bad network) and not anything to do with code quality.

drewfish:
Oh, still need test(s) to make sure that app.json resourceStore.lazyLoad (a) actually causes a lazy load, and (b) an app still works with lazy load.

drewfish:
One last task. Please add a note in the HISTORY.md file mentioning the new feature. You could add something like this in the "Features" section:

```
* There is a new `resourceStore.lazyResolve:true` option in `application.json`. 
Normally, the resource store at server start pre-calculates the details for all mojits for all contexts/selectors.
(It's optimized to only do this for dimensions/selectors that are actually used.)
By setting this new option the resource store will skip this calculation at srever start and instead do it at runtime as needed.
This might be a good option to use if you have a large number of dimensions defined in `dimensions.json` and a lot of selectors defined in `application.json`,
yet your app will only serve traffic from a subset of those.                    

```

drewfish:
Travis is having issues, but the union of builds for 3cf8896 and 91c6bf2 give full tests coverage and passing tests.

drewfish:
+1 :shipit:

imalberto:
lgtm +1

PR #1196, opened 2013-07-16
set runtime:server as default when booting app

lzhan:
functional test is added.


tests/func/applications/frameworkapp/configapp/routes.json
```diff
@@ -11,5 +11,13 @@
         "path": "/:mojit-id/:mojit-action",
         "call": "{mojit-id}.{mojit-action}"
     }
+  },
```
caridy: we are removing routes.json contextualization, better to use a different file for the test.

lib/store.js
```diff
@@ -56,6 +56,7 @@ Store.createStore = function(options) {
     if (!options.preload) {
         options.preload = 'full';
     }
+    options.context = {runtime: 'server'};
```
caridy: context might be set, instead, you can do this:

```
if (!options.context.runtime) {
    options.context.runtime = 'server';
}
```

that will work in most cases, even when you want to force runtime to be another thing from the cli, etc.

lzhan:
travis failed, but it is not related to the pr.

caridy:
+1 please, verify that travis passes before merging.

TODO
- [x] History entry is still missing

lzhan:
travis is fixed.

PR #1198, opened 2013-07-17
fix #1017 support for README.md at mojit level

### LINKED ISSUE #1017: Putting  a README.md in the mojits/ directory causes mojito to not load
```
$ touch mojits/README.md
$ ./node_modules/mojito/bin/mojito start --context environment:development
info: (mojito-resource-store): Store initialized
info: (mojito-resource-store): Store initialized

/tmp/ui/node_modules/mojito/lib/app/addons/rs/config.js:149
                            throw new Error("Error parsing file: " + fullPath 
                                  ^
Error: Error parsing file: /tmp/ui/mojits/README.md/definition.json
```

Not a big deal but a bug all the same.

lzhan:


lib/app/autoload/store.server.js
```diff
@@ -1824,6 +1824,10 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 res;
             dir = this._libs.path.resolve(this._config.root, dir);
 
+            if (!this._libs.fs.lstatSync(dir).isDirectory()) {
```
caridy, line 1837: how does this solves the issue? can you elaborate more?

lzhan, line 1837: The original implementation treats every thing under mojits dir as a mojit(dir), it tries to read files under the dir. When README.md file is put under mojits dir, parsing err is thrown. The fix check if it is dir, if it is not, just return, skip the loading process for mojit.

caridy, line 1837: Ok

caridy:
+1

TODO:
- [x] history entry
- [x] unit test

PR #1200, opened 2013-07-19
Hotswap

jarnoux:
This pull request enables mojito to "hotswap" resources without having to restart node every time developers make a change on the disk. It will monitor YUI modules and view templates
The associated configuration in the app config is resourceStore.hotswap: true.
By default, it is not activated.
NOTE: hotswapping views disables template and renderers caches (which is fine at dev time, just don't hotswap in prod.)

### DIFF
--- a/lib/app/addons/rs/hotswap-yui.js
+++ b/lib/app/addons/rs/hotswap-yui.js
@@ -18,7 +18,7 @@ YUI.add('addon-rs-hotswap-yui', function (Y, NAME) {
     Y.extend(RSAddonHotswapYUI, Y.Plugin.Base, {
         initializer: function (config) {
             var appConfigStatic = config.host.getStaticAppConfig();
-            if (appConfigStatic.resourceStore && appConfigStatic.yui.hotswap) {
+            if (appConfigStatic.resourceStore && appConfigStatic.resourceStore.hotswap) {
                 this.config = config;
                 // put watchers on the resources once resolved
                 this.afterHostMethod('parseResourceVersion', this.parseResourceVersion, this);
--- a/lib/app/addons/rs/hotswap-yui.js
+++ b/lib/app/addons/rs/hotswap-yui.js
@@ -17,8 +17,8 @@ YUI.add('addon-rs-hotswap-yui', function (Y, NAME) {
 
     Y.extend(RSAddonHotswapYUI, Y.Plugin.Base, {
         initializer: function (config) {
-            this._appConfigStatic = config.host.getStaticAppConfig();
-            if (this._appConfigStatic.resourceStore && this._appConfigStatic.yui.hotswap) {
+            var appConfigStatic = config.host.getStaticAppConfig();
+            if (appConfigStatic.resourceStore && appConfigStatic.yui.hotswap) {
                 this.config = config;
                 // put watchers on the resources once resolved
                 this.afterHostMethod('parseResourceVersion', this.parseResourceVersion, this);
@@ -35,16 +35,17 @@ YUI.add('addon-rs-hotswap-yui', function (Y, NAME) {
                 fs.watch(source.fs.fullPath, { persistent: false }, function (event) {
                     try {
                         if (fs.readFileSync(source.fs.fullPath, 'utf8')) {
-                            // load
                             host.runtimeYUI.applyConfig({ useSync: true });
-                            host.runtimeYUI.Get.js(source.fs.fullPath, {});
-                            host.runtimeYUI.applyConfig({ useSync: false });
 
+                            // load
+                            host.runtimeYUI.Get.js(source.fs.fullPath, {});
                             // use
                             host.runtimeYUI.Env._attached[res.yui.name] = false;
                             host.runtimeYUI.use(res.yui.name, function () {
                                 host.runtimeYUI.log('Reloaded yui module at: ' + source.fs.fullPath, 'info', NAME);
                             });
+
+                            host.runtimeYUI.applyConfig({ useSync: false });
                         }
                     } catch (e) {
                         host.runtimeYUI.log('Failed to reload module ' + (res.yui && res.yui.name) + ' at ' + source.fs.fullPath + '\n' + e.message, 'error', NAME);
--- a/lib/app/addons/rs/hotswap-yui.js
+++ b/lib/app/addons/rs/hotswap-yui.js
@@ -6,24 +6,22 @@
 
 /*jslint nomen: true, stupid: true */
 
-YUI.add('addon-rs-hotswap', function (Y, NAME) {
+YUI.add('addon-rs-hotswap-yui', function (Y, NAME) {
     'use strict';
     var fs = require('fs');
 
-    function RSAddonHotswap() {
-        RSAddonHotswap.superclass.constructor.apply(this, arguments);
+    function RSAddonHotswapYUI() {
+        RSAddonHotswapYUI.superclass.constructor.apply(this, arguments);
     }
-    RSAddonHotswap.NS = 'hotswap';
+    RSAddonHotswapYUI.NS = 'hotswap-yui';
 
-    Y.extend(RSAddonHotswap, Y.Plugin.Base, {
+    Y.extend(RSAddonHotswapYUI, Y.Plugin.Base, {
         initializer: function (config) {
             this._appConfigStatic = config.host.getStaticAppConfig();
-            if (this._appConfigStatic.resourceStore && this._appConfigStatic.resourceStore.hotswap) {
+            if (this._appConfigStatic.resourceStore && this._appConfigStatic.yui.hotswap) {
                 this.config = config;
                 // put watchers on the resources once resolved
                 this.afterHostMethod('parseResourceVersion', this.parseResourceVersion, this);
-                this.afterHostMethod('getStaticAppConfig', this.getStaticAppConfig, this);
-                this._cachedTemplates = true;
             }
         },
         parseResourceVersion: function (source, type, subtype, mojitType) {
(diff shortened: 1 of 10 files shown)


caridy:
Thanks @jarnoux, I will look into it today.

lib/app/autoload/view-renderer.server.js
```diff
@@ -24,10 +24,11 @@ YUI.add('mojito-view-renderer', function (Y) {
      */
     function Renderer(type, options) {
         this._type = type || 'hb';
-        this._options = options;
-        if (!cache[this._type]) {
-             // caching renderer instance
-            cache[this._type] = new (Y.mojito.addons.viewEngines[this._type])(options);
+        this._options = options || {};
+
+        if (!cache[this._type] || this._options.cacheTemplates === false) {
```
caridy: this cache is not the template cache, but the engine cache. I don't think you need this.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2011-2013, Yahoo! Inc.  All rights reserved.
+ * Copyrights licensed under the New BSD License.
+ * See the accompanying LICENSE file for terms.
+ */
+
+/*jslint nomen: true, stupid: true */
+
+YUI.add('addon-rs-hotswap', function (Y, NAME) {
```
caridy: please, adjust the 4-space-tabs to match mojito code.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+			}
+		},
+		parseResourceVersion: function (source, type, subtype, mojitType) {
+			var self = this,
+				host = self.config.host,
+				res = Y.Do.currentRetVal;
+
+			// if the modified resource is a script, reload it from the filesystem
+			// in the runtime YUI instance
+			if ('.js' === source.fs.ext) {
```
caridy: I think this only applies to yui modules, not all js files.

lib/app/autoload/view-renderer.server.js
jarnoux: I do because the cached engine contains the template caching configuration and since I can't programmatically change that config inside the renderer, it's not going to rerender the view with the new content

J.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+                fs.watch(source.fs.fullPath, { persistent: false }, function (event) {
+					try {
+						if (fs.readFileSync(source.fs.fullPath, 'utf8')) {
+							// load
+							host.runtimeYUI.applyConfig({ useSync: true });
```
caridy: `useSync` does not apply to `get` if I remember correctly.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+							host.runtimeYUI.Get.js(source.fs.fullPath, {});
+							host.runtimeYUI.applyConfig({ useSync: false });
+
+							// use
+							host.runtimeYUI.Env._attached[res.yui.name] = false;
+							host.runtimeYUI.use(res.yui.name, function () {
```
caridy: here is where you need the `useSync`, so the `use` statement can be sync.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+								host.runtimeYUI.log('Reloaded yui module at: ' + source.fs.fullPath, 'info', NAME);
+							});
+						}
+					} catch (e) {
+						host.runtimeYUI.log('Failed to reload module ' + (res.yui && res.yui.name) + ' at ' + source.fs.fullPath + '\n' + e.message, 'error', NAME);
+					}
+				});
+			} else if (res.type === 'view') {
```
caridy: if you have the the proper flag, templates will not be cached, why do you need to watch them?

```
    viewEngine:
          cacheTemplates: true
```

lib/app/addons/rs/hotswap.js
jarnoux: I believe it does, there is a conditional fs.readFile(Sync) based on that option
J.

caridy:
There is a fundamental flaw with this approach, it is the fact that YUI uses a namespace, which means you can hold references around, and those references will not be updated whenever you re-attach a module. E.g:

`controller-foo` requires `bar`, if `controller-foo` is updated, it can be re-use thru loader, and it will be able to resolve the references to the `bar` namespace, and everything will work just fine. Now, if you update `bar`, and re-use it, the references that `controller-foo` is holding will be preserved, which means that `controller-foo` will be using the old `bar`, not the new one.

There is a little that we can do about this. I think it is not a blocker.

jarnoux:
yes, this is true. there are other limitations as well, like if you add a new module in a separate file, or if you add a require that is not already loaded, the require will fail, or if it's attached to the startup YUI instance (like the resource store itself). but I think the functionality still encompasses most of the useful, most frequent cases.

J.

lib/app/addons/rs/hotswap.js
jarnoux: I think Get needs it, see: https://github.com/yui/yui3/blob/master/src/get/js/get-nodejs.js#L107

lib/app/addons/rs/hotswap.js
jarnoux: hotswap is essencially a convenience for devs, so it has to be explicitely turned on for YUI modules anyways. The addon automatically sets that `viewEngine.cacheTemplates: true` option so you only have only one option: `resourceStore.hotswap` to set instread of two. Now why the watch? it allows to keep the caching on if the dev is not modifying any view.

lib/app/addons/rs/hotswap.js
jarnoux: that is correct. would enumerating the different resource types of yui modules be a better solution?

lib/app/addons/rs/hotswap.js
jarnoux: here, please let me know if there is anything else that's a blocker to merge this pull request

lib/mojito.js
```diff
@@ -446,6 +446,8 @@ MojitoServer.prototype._configureYUI = function(Y, store) {
     modules = store.yui.getModulesConfig('server', false);
     Y.applyConfig(modules);
 
+    store.runtimeYUI = Y;
```
caridy, line 449: store should NOT know about the runtime Y for mojito, but since you need it to refresh the pieces, I think it should be a global variable instead, something like `global.mojitoY`, then accessing that from within `parseResourceVersion`.

lib/app/addons/rs/hotswap.js
caridy: No so fast @jarnoux, lets make sure we do the right thing here. I don't even know if the addon should be part of mojito, it could be package called `mojito-hotswap` that adds this capabilities. I will discuss the details with the other folks on Wed. In the meantime, I will do a second pass on the review.

lib/app/addons/rs/hotswap.js
caridy: good, then don't disable it until after `host.runtimeYUI.use(res.yui.name)`, which should be sync as well.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+            // if the modified resource is a script, reload it from the filesystem
+            // in the runtime YUI instance
+            if ('.js' === source.fs.ext) {
+                fs.watch(source.fs.fullPath, { persistent: false }, function (event) {
+                    try {
+                        if (fs.readFileSync(source.fs.fullPath, 'utf8')) {
+                            // load
+                            host.runtimeYUI.applyConfig({ useSync: true });
+                            host.runtimeYUI.Get.js(source.fs.fullPath, {});
+                            host.runtimeYUI.applyConfig({ useSync: false });
```
caridy: move this line to the end of the block.

lib/app/addons/rs/hotswap.js
```diff
@@ -0,0 +1,77 @@
...
+
+                            // use
+                            host.runtimeYUI.Env._attached[res.yui.name] = false;
+                            host.runtimeYUI.use(res.yui.name, function () {
+                                host.runtimeYUI.log('Reloaded yui module at: ' + source.fs.fullPath, 'info', NAME);
```
caridy: remove the callback, it is a sync method anyways.

lib/app/autoload/view-renderer.server.js
caridy: I'm not convinced, the cache for the templates is a lazy cache, which means we always add the item to the cache, but we don't get items from cache unless `cacheTemplates` is true. In other words, having a `_cache` object attached to the view engine is not an issue, because when the accessor wants to pick up a template and render it in the engine, it will not use that cache if `cacheTemplates` if not true.

lib/mojito.js
jlecomte, line 449: Why should the store NOT know about the YUI instance used by mojito at runtime? I don't like the idea of clobbering the global node namespace.

lib/app/addons/rs/hotswap-yui.js
```diff
@@ -0,0 +1,59 @@
...
+    var fs = require('fs');
+
+    function RSAddonHotswapYUI() {
+        RSAddonHotswapYUI.superclass.constructor.apply(this, arguments);
+    }
+    RSAddonHotswapYUI.NS = 'hotswap-yui';
+
+    Y.extend(RSAddonHotswapYUI, Y.Plugin.Base, {
+        initializer: function (config) {
+            this._appConfigStatic = config.host.getStaticAppConfig();
```
caridy: Not need to store this in memory anymore, a local variable should be just fine. Remember that this is a copy of the static config, and keeping it in memory increases the number of elements to be walked by GC.

lib/app/addons/rs/hotswap-yui.js
```diff
@@ -0,0 +1,60 @@
...
+            if ('.js' === source.fs.ext) {
+                fs.watch(source.fs.fullPath, { persistent: false }, function (event) {
+                    try {
+                        if (fs.readFileSync(source.fs.fullPath, 'utf8')) {
+                            host.runtimeYUI.applyConfig({ useSync: true });
+
+                            // load
+                            host.runtimeYUI.Get.js(source.fs.fullPath, {});
+                            // use
+                            host.runtimeYUI.Env._attached[res.yui.name] = false;
```
drewfish, line 43: Uggg.. there's a bunch of yui-internal stuff in here. Can we get a "re-use()" feature added to YUI, and use that instead?

drewfish:
Unit tests are missing.  (You can run unit tests using `npm run unit`.)

lib/app/addons/rs/hotswap-yui.js
caridy, line 43: yes, I will add a card for that for @ekashida to work on.

jlecomte:
We have decided to package this hotswap plugin in a separate npm package. I will close this PR shortly.

PR #1201, opened 2013-07-23
fix #1192 apidocs for Data.common (ac.data and ac.pageData)

### LINKED ISSUE #1192: Document the API methods for the Data addon (Data.common class).
The API documentation for the Data.common class says that this class  doesn't provide any methods, properties, attributes, or events. The general documentation covers how to use the Data addon, but developers need to be able to refer to the API docs. It's simply a matter of adding the YUIDoc tags and descriptions to the source code.

isao:
This adds two items to the class nav in the mojito api docs, and references Model.Vanilla. Technically it's `Y.Model` on the server, but it is api-compatible.
- Data.common.pageData
- Data.common.data

### DIFF
--- a/lib/app/addons/ac/data.common.js
+++ b/lib/app/addons/ac/data.common.js
@@ -16,8 +16,9 @@ YUI.add('mojito-data-addon', function (Y, NAME) {
 
     /**
      * <strong>Access point:</strong> <em>ac.data.*</em> and  <em>ac.pageData.*</em>
-     * Addon that provides access to the data and pageData models
-     * @class Data.common
+     * Addon that provides access to the data and pageData models.
+     * See also http://developer.yahoo.com/cocktails/mojito/docs/topics/mojito_data.html#sharing-data
+     * @submodule Data.common
      */
     function Addon(command, adapter, ac) {
         var data = command.instance.data,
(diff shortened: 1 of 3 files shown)


lib/app/addons/ac/data.common.js
```diff
@@ -33,6 +34,28 @@ YUI.add('mojito-data-addon', function (Y, NAME) {
...
+     * A model for getting and setting data  that is shared by a single mojit
+     * controller, binder, and/or template.
+     *
+     * See also http://developer.yahoo.com/cocktails/mojito/docs/topics/mojito_data.html#sharing-data
+     * @class data
+     * @namespace Data.common
+     * @uses Model.Vanilla
+     */
+    /**
+     * <strong>Access point:</strong> <em>ac.data.*</em>
```
caridy: `<em>ac.pageData.*</em>`?

caridy:
We could link to http://yuilibrary.com/yui/docs/api/classes/Model.html or we can document at least the `get` and `set`, which are the common methods for `pageData` and `data`.

isao:
I link to the manual, and the full api docs are pulled in from Model.Vanilla. Here's a screenshot: http://cl.ly/QQvy (get & set, as well as all the other methods, get displayed when rendered by yuidoc)

lib/app/addons/ac/data.common.js
isao: ah, yes, thanks-- copy and paste error

isao:
follow up on listing get/set explicitly-- won't work with `@uses`

isao:
travis errors were unrelated. this PR has no code changes.

```
$ wget -nv http://selenium.googlecode.com/files/selenium-server-standalone-2.30.0.jar
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

lib/app/autoload/mojit-proxy.client.js
```diff
@@ -50,18 +50,62 @@ YUI.add('mojito-mojit-proxy', function(Y, NAME) {
...
         /**
-         * The mojit data model
+         * The mojit data model, for getting and setting data that is shared by
+         * a single mojit controller, binder, and/or template.
+         * Use this.data.get() and this.data.set(). See also
+         * http://yuilibrary.com/yui/docs/api/classes/Model.html
+         * and http://developer.yahoo.com/cocktails/mojito/docs/topics/mojito_data.html#sharing-data
          * @property data
-         * @type {Object}
+         * @type {Model.Vanilla}
```
isao, line 59: while it's clear from the code on the next non-comment line, that the actual type is Y.Model, using Model.Vanilla is api-compatible and is what's used on the server. The reason I use it is so yuidoc will hyperlink to it's api docs, which are the same.

lzhan:
+1

PR #1202, opened 2013-07-23
cherry-picked cli removal

isao:
note run.js has some modifications that we'll want to also cherry-pick or copy over later. this pr is the result of this cherry-pick onto `next`, with conflict cleanup

```
git cherry-pick cea8d3db5b137a3f69b2a87766b587b19e364a01 -m 1
```


package.json
```diff
@@ -16,22 +16,15 @@
         "Lichun Zhan <lzhan@yahoo-inc.com>"
     ],
     "dependencies": {
-        "colors": "*",
         "debug": "*",
+        "express": "3.x",
```
imalberto: Please move back to `devDependencies` as the application will make express its `dependencies`.

tests/func/applications/frameworkapp/commandline/package.json
```diff
@@ -8,6 +8,7 @@
               "email": "nobody@yahoo-inc.com"
```
imalberto: Should commandline/ folder be deleted ?

isao: yes I noticed `tests/func/applications/frameworkapp/commandline/*` too, and it does look like it's no longer used. will also rm from head of develop.

imalberto:
+1

PR #1209, opened 2013-07-31
add setting of staticHandling/cache, forceUpdate in test app/examples

lzhan:


isao:
+1

caridy:
My only concern about this is that people will tend to copy those examples, which means they will copy those dev-only settings. I will like to have them into the proper `environment:development` context, and make sure that our apps are running with `--context enviroment:development` when they are about to be used to func tests. This will be ideal IMO.

lzhan:
Thanks Caridy. I will update the application.json with the setting under environment:development. I am not sure how to passing --context enviroment:development with the new way starting mojito app("node app"). Is the development env the default?

caridy:
Yeah, I don't know how that part works in our functional tests.

PR #1211, opened 2013-07-31
inject express into mojito instance

isao:
pass express app instance into mojito constructor explicitly,
instead of patching express.application.defaultConfiguration()
from mojito.js, and having app.js instantiate mojito by way of
express().

### DIFF
--- a/examples/newsboxes/app.js
+++ b/examples/newsboxes/app.js
@@ -10,11 +10,8 @@
 'use strict';
 
 var debug = require('debug')('app'),
-    express = require('express'),
-    mojito = require('../../'),
-    app;
-
-app = express();
+    app = require('express')(),
+    mojito = require('../../')(app);
 
 app.use(mojito.middleware());
 app.mojito.attachRoutes();
--- a/lib/mojito.js
+++ b/lib/mojito.js
@@ -16,11 +16,8 @@ instantiate their app the `express` way.
 
 Usage:
 
-    var app = require('express'),
-        mojito = require('mojito'),
-        app;
-
-    app = express(); // Mojito instance will be created during that call
+    var app = require('express')(),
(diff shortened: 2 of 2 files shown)


caridy:
not so sure about the reasons behind this change, but the direction we are moving with modown express extensions is a little bit different. Here is what we have in `express-yui` and `express-state` so far is this:
- `require('<extension-name>')` will `require('express')` under the hood and extend it.
- `<extension-name>.extend(express);` will extend any express class.
- `<extension-name>.augment(app);` will augment any express app instance.

These three ways to plug an express extension solves all scenario as far as I can tell. Hopefully we can agree that mojito-next is going to be, primarily, an express extension, and we can follow the same principle. Here is the code:

https://github.com/yahoo/express-yui/blob/master/lib/yui.js#L208

isao:
Yes, saw that in express-yui and others, but assumed there was some reason the approach was needed there, and I thought @ericf mentioned wanting to move away from that earlier today. Also, if both mojito.js and express-yui depend on monkey-patching the same function to get instantiated, they can't be used together easily, can they?

Anyway, the patch fell out of some other stuff I was trying for simplifying app-start, and wanted to share quickly before I left on vacation (func test app haven't been updated, etc.). I thought it was a small improvement...

caridy:
having multiple extensions wrapping the `defaultConfiguration` is really not a problem because it will create a chain of functions to be executed on the same context, that will work just fine. The question that is on the table today is whether or not we want to pursue the first item from my list (plug automatically), but on the other two items, we are convinced. But since it doesn't hurt to automatically plug the extension into express and then manually try to plug it again, that give us some more time to decide about item 1.

caridy:
@ericf wrote the proposal to formalize this here:

https://gist.github.com/ericf/6133744

isao:
that looks great. closing this pr.

PR #1213, opened 2013-08-09
cleanup of contextualized routes

drewfish:


caridy:
+1

thanks for cleaning this up @drewfish

Issue #1219, opened 2013-08-14
npm link shared mojits failed on 0.7.2


ftliou:
Please refer to:
http://developer.yahoo.com/forum/Yahoo-Mojito/npm-link-shared-mojits-failed-on/1376445752796-c287e870-fd51-473e-ac9a-557335d087a3

In mojito/lib/app/autoload/store.server.js

```
       if (!this._libs.fs.lstatSync(dir).isDirectory()) {
            return;
        }
```

This block of code stops npm linked shared mojits from being loaded.

caridy:
regression introduced by fix for issue #1017 /cc @lzhan @isao

caridy:
@lzhan can you validate and close this?

PR #1220, opened 2013-08-14
fix issue #1219 npm link shared mojits failed

lzhan:


imalberto:
@lzhan Probably should add some tests to cover this for future regression.

lib/app/autoload/store.server.js
```diff
@@ -1834,7 +1834,7 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 res;
             dir = this._libs.path.resolve(this._config.root, dir);
 
-            if (!this._libs.fs.lstatSync(dir).isDirectory()) {
+            if (!this._libs.fs.lstatSync(dir).isDirectory() && !this._libs.fs.lstatSync(dir).isSymbolicLink()) {
```
caridy: I wonder, if it is not a dir, and it is not a symlink, what else could it be? I think this is equivalent to say, if it is a file, exit :)

lzhan: That will not cover symlink file.

lzhan:
@imalberto I don't think symlink works as usual on git.

lib/app/autoload/store.server.js
isao: `fs.lstat(zzz).isDirectory()` is always false if `zzz` string points to a symlink, whether or not it resolves to a directory.

isao:
symlinks work fine in git, but not on windows..

lib/app/autoload/store.server.js
isao: (if you want to follow the symlink, use `fs.stat` instead of `fs.lstat`)

isao: is this more what you're looking for?

```
// return if dir is not a directory, or a symlink to a directory
if (!this._libs.fs.statSync(dir).isDirectory()) {
    return;
}
```

(..also move it after the exists() test further down...)

note that this will throw if `dir` is a symlink to non-existent file.

drewfish: Yeah, this looks a lot cleaner. In general we should care whether something is the exact resource we want or a symlink to it. That implies the use of stat instead of lstat. (Said another way, only use lstat if you really do care whether something is a symlink or not. We don't.)

lzhan: Got it. Thanks everyone.

lib/app/autoload/store.server.js
```diff
@@ -1834,11 +1834,11 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
                 res;
             dir = this._libs.path.resolve(this._config.root, dir);
 
-            if (!this._libs.fs.lstatSync(dir).isDirectory()) {
+            if (!(this._libs.fs.existsSync || this._libs.path.existsSync)(dir)) {
```
caridy: We don't need to worry about `path.existsSync()` anymore because we required nodejs 0.8 now.

isao:
+1

PR #1223, opened 2013-08-15
Add perf graph for mojito release builds

lzhan:


builds/mojito-perf-component/release.html
```diff
@@ -0,0 +1,56 @@
...
+<div class="graph" id="graph_start"></div>
+
+
+<script>
+    YUI().use('MojitoGHGraphs', function (Y) {
+        var count = 20;
+        var matches = window.location.search.match(/count=(\d+)/);
+        if (matches) {
+            count = parseInt(matches[1], 10);
+        }
```
drewfish: Doesn't look like this is using the `?count=` parameter, so this code can be deleted.

graphs/lib.js
```diff
@@ -187,10 +198,19 @@ YUI.add('MojitoGHGraphs', function(Y, NAME) {
 
 
         // returns promise
-        // ... fufilled with same format as getLast()
```
drewfish, line 190: I think we can keep this comment ("fulfilled with the same format as getLast()") since it -is- true :)

drewfish:
+1

PR #1224, opened 2013-08-16
fix test with 0 passed 0 failed result

lzhan:
Using YUI.add instead of YUI.use to fix the 0 passed, 0 failed, 0 skipped error. 
More unit tests fix coming.


drewfish:
:doughnut:

Issue #1215, opened 2013-08-12
[regression] lang collection in metas is missing for controller 


caridy:
Reported by @ft thru the forum: http://developer.yahoo.com/forum/Yahoo-Mojito/Internationalization-on-client-side/1376040501301-15e8a4da-ac8c-4391-985f-093001c8b477/1376293135175

It seems like the list of supported langs for each controlled is not added correctly to the meta, so, for mojits running on the client side it will never load the corresponding lang bundle. Here is an example of the meta from the repro case (https://github.com/ftliou/mojito-test):

```
YUI.add("loader-app", function(Y) {
    Y.applyConfig({
        groups: {
            app: Y.merge(((Y.config.groups && Y.config.groups.app) || {}), {
                modules: {
                    "Master": {
                        "path": "Master.js",
                        "requires": ["mojito", "dump", "mojito-composite-addon", "mojito-intl-addon", "mojito-assets-addon", "intl"]
                    },
                    "lang/Master_es": {
                        "path": "Master_es.js",
                        "requires": ["intl"]
                    }
                }
            })
        }
    });
}, "", {
    requires: ["loader-base"]
});
```

In this case, `Master` should have the collection of langs right after `requires`.

caridy:
+mojito@0.7.5

PR #1232, opened 2013-08-27
fix issue #1215 lang collection in metas is missing for controller

lzhan:


tests/fixtures/gsg5/mojits/PagedFlickr/controller.common.js
```diff
@@ -92,5 +92,6 @@ YUI.add('PagedFlickr', function(Y) {
     'mojito-params-addon',
     'mojito-url-addon',
     'mojito-util',
-    'ModelFlickr'
+    'ModelFlickr',
+    'intl'
```
caridy: Using `intl` as the indicator of lang association is not really an option because you might be getting `intl` from one of the dependencies. In this case, the module is getting lang stuff thru `mojito-intl-addon` but it does not use `Y.Lang.*` directly, so there is not need to have `intl` in the requirement list.

In the past, what I did was to rely on the name of the modules to know if there is a need to have a lang token associated with a module. We can discuss the details offline.

tests/func/applications/frameworkapp/usecase/application.json
```diff
@@ -14,6 +14,23 @@
             "i18n-2": {
                 "type": "i18n_Mojit"
             },
+            "i18n-3": {
+                "type": "HTMLFrameMojit",
+		        "config": {
```
caridy: format issues here.

tests/func/applications/frameworkapp/usecase/mojits/i18n_ChildMojit/binders/index.js
```diff
@@ -0,0 +1,39 @@
+/*jslint anon:true, sloppy:true, nomen:true*/
+YUI.add('i18n_ChildMojitBinderIndex', function(Y, NAME) {
```
caridy: If this binder is not doing much, why do we need it in the first place?

tests/func/applications/frameworkapp/usecase/mojits/i18n_ChildMojit/binders/index.js
```diff
@@ -0,0 +1,39 @@
...
+         * @param node {Node} The DOM node to which this mojit is attached.
+         */
+        bind: function(node) {
+            var me = this;
+            this.node = node;
+        }
+
+    };
+
+}, '0.0.1', {requires: ['event-mouseenter', 'mojito-client']});
```
caridy: `event-mouseenter` is not needed here.

tests/func/applications/frameworkapp/usecase/mojits/i18n_ChildMojit/controller.common.js
```diff
@@ -0,0 +1,34 @@
...
+        	ac.assets.addCss('./index.css');
+            ac.done({
+                status: 'Mojito child is working.',
+                data: { some: ac.intl.lang('some')}
+            });
+        }
+
+    };
+
+}, '0.0.1', {requires: ['mojito','dump', 'mojito-assets-addon','mojito-intl-addon','intl']});
```
caridy: - break line
- `intl` is not needed in the requires because the module is not calling `Y.Intl` or `Y.Lang` in the code.

tests/func/applications/frameworkapp/usecase/mojits/i18n_ParentMojit/binders/index.js
```diff
@@ -0,0 +1,54 @@
...
+        /**
+         * The binder method, invoked to allow the mojit to attach DOM event
+         * handlers.
+         *
+         * @param node {Node} The DOM node to which this mojit is attached.
+         */
+        bind: function(node) {
+            var me = this;
+            this.node = node;
+            /**
```
caridy: remove commented code.

tests/func/applications/frameworkapp/usecase/mojits/i18n_ParentMojit/binders/index.js
```diff
@@ -0,0 +1,54 @@
+/*jslint anon:true, sloppy:true, nomen:true*/
+YUI.add('i18n_ParentMojitBinderIndex', function(Y, NAME) {
```
caridy: same as above, why do we need this?

tests/func/applications/frameworkapp/usecase/mojits/i18n_ParentMojit/controller.common.js
```diff
@@ -0,0 +1,35 @@
...
+    Y.namespace('mojito.controllers')[NAME] = {
+
+        /**
+         * Method corresponding to the 'index' action.
+         *
+         * @param ac {Object} The ActionContext that provides access
+         *        to the Mojito API.
+         */
+        index: function(ac) {
+        	Y.log("lang="+Y.dump(Y.Intl.get(NAME)),"warn",NAME);
```
caridy: format issue.

tests/func/applications/frameworkapp/usecase/mojits/i18n_ParentMojit/controller.common.js
```diff
@@ -0,0 +1,35 @@
...
+            ac.assets.addCss('./index.css');
+            ac.composite.done({
+                status: 'Mojito is working.',
+                data: { some: ac.intl.lang('some') }
+            });
+        }
+
+    };
+
+}, '0.0.1', {requires: ['mojito','dump', 'mojito-composite-addon','mojito-intl-addon','mojito-assets-addon','intl']});
```
caridy: `intl` not needed here.

lib/app/addons/rs/yui.js
```diff
@@ -992,11 +998,27 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                         // do its thing
                     }
 
+
+                    for (i = 0; i < module.requires.length; i += 1) {
+                        if (module.requires[i] === 'intl' && !module.intl && module.name.substring(0, 7) !== ('mojito-')) {
```
caridy: this is going to be problematic, let's talk offline about the intl detection.

caridy:
done with the first pass.

lib/app/addons/rs/yui.js
```diff
@@ -946,20 +950,33 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
...
+
                     bundle = name.indexOf('lang/') === 0;
                     lang = bundle && REGEX_LOCALE.exec(name);
-
                     if (lang) {
                         mod = mod.slice(0, lang.index); // eg. lang/foo_en-US -> lang/foo
                         lang = lang[1];
+                        mod1 = mod.split("/");
+                        add = true;
+                        for (i = 0; i < intlmodules.length; i += 1) {
```
caridy: this is still problematic because the order of the modules is not strict, so you might process `lang/foo` before `foo`.

caridy: btw, `myArray.indexOf('foo') > -1` will be good enough to locate a string.

lib/app/addons/rs/yui.js
```diff
@@ -1009,6 +1031,32 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                     }
                 }
             }
+
+            for (m in resolvedMods) {
+                if (resolvedMods.hasOwnProperty(m)) {
+                    module = resolvedMods[m];
+                    for (i = 0; i < intlmodules.length; i += 1) {
```
caridy: Same here, `indexOf()` will be just enough.

lib/app/addons/rs/yui.js
lzhan: @caridy I don't see there is an issue here if foo is processed after lang. The logic is 
1. the first "for (m in resolvedMods)" will go though all the modules and if a module starts with lang/, then the name of the mojit is saved in an array.
2. the second "for (m in resolvedMods)" will go though all the module and if the mojit name is found in the array add lang info to the module.
So I don't think the order of processing the modules matters.
I don't like the second for loop, but looks like there is no good way to get those mojits directly just by name.

caridy: got it! makes sense. Let's just use `indexOf()` which is faster than the array since it is native.

lib/app/addons/rs/yui.js
```diff
@@ -954,20 +958,31 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
...
+
                     bundle = name.indexOf('lang/') === 0;
                     lang = bundle && REGEX_LOCALE.exec(name);
-
                     if (lang) {
                         mod = mod.slice(0, lang.index); // eg. lang/foo_en-US -> lang/foo
                         lang = lang[1];
+                        mod1 = mod.split("/");
+                        add = true;
+                        if (intlmodules.indexOf(mod1[1]) > -1) {
```
caridy: these 2 if statements are confusing. The thing here is just to check if the mod is not in intlmodule, add it. There is not need to use a variable `add` and two `if` anymore. Let's clear that up and get ready to merge.

lib/app/addons/rs/yui.js
```diff
@@ -1017,6 +1037,30 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
                     }
                 }
             }
+
+            for (m in resolvedMods) {
+                if (resolvedMods.hasOwnProperty(m)) {
```
caridy, line 1036: can you add some notes/comments to describe the flow of these whole block? complex blocks should be described so people can get the whole picture.

lib/app/addons/rs/yui.js
```diff
@@ -1017,6 +1032,33 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
...
             }
+
+            //scan modules in resolvedMods, if a module is also listed in intlmodules,
+            //lang holder will be added to the module's meta data.
+            //lang info will be plugged in later when _precalcLoaderMeta is called
+            for (m in resolvedMods) {
+                if (resolvedMods.hasOwnProperty(m)) {
+                    module = resolvedMods[m];
+                    if (intlmodules.indexOf(module.name) > -1) {
+                        modules[module.name] = {};
```
caridy: normally, modules and expanded_modules should be set already when reaching this point. Isn't this equivalent to say just:

```
modules[module.name].lang = ['languageholder'];
```

and the same for extended_modules? I don't think we need all this duplication of code for something that already happens before.

caridy:
+1

PR #1233, opened 2013-08-27
More refactoring of routing components

imalberto:
CHANGELOG:
- Now using `express-annotations` to augment route configuration with meta data
- Now using `express-map` to generate route config format used throughout mojito
- Now using `express-state` `serialize` method to replace `JSON.stringify` for serializing objects to client runtime
- Refactored set of components to deal with revised route configuration format generated by `express-map`
- Updated `dispatcher.js`, `router.js` to take advantage of `express-map` and `express-annotations`
- Updated `ac.url` addon and `route-maker` to leverage `express-map`

TODO:
- [x] fix unit tests
- [x] fix func tests
- [x] update examples
- [x] publish most current version of `express-annotations` to npm
- [x] publish most current version of `express-map` to npm
- [x] merge with upstream (next branch)


lib/mojito.js
```diff
@@ -69,6 +71,11 @@ Mojito extension for Express.
 **/
 function Mojito(app) {
 
+    // FIXME: This might not be required since app.js would have already
```
caridy: I don't think we should worry about this. If they try to extend it manually, let the lib to handle that :)

caridy:
This is looking good!

lib/router.js
```diff
@@ -161,26 +161,17 @@ module.exports = {
...
-                    }
-                }
-            }
+            Object.keys(routes).forEach(function (name) {
+                var route = routes[name];
+                Object.keys(route.verbs).forEach(function (verb) {
+                    debug('[%s %s] installing handler', route.call, route.path);
+                    verb = verb.toLowerCase();
+
+                    app[verb](route.path, dispatcher.dispatch(route.call));
```
ericf: I think right after the route is created, we could add the `dispatch` annotation. `dispatcher.dispatch()` returns a route handler function with a `dispatch` property, we could capture this return value and add the annotation here. I think this will be much clearer than adding the annotation it inside the `dispatcher.js`.

``` javascript
var handler = dispatcher.dispatch(route.call);

app[verb](route.path, handler);
app.annotate(route.path, { dispatch: handler.dispatch });
```

Maybe there's some timing thing I'm missing, but this means you could remove the iteration over all the routes, the `init` state flag, etc from `dispatcher.js`.

lib/dispatcher.js
```diff
@@ -65,32 +69,18 @@ module.exports = {
...
-                                    keys: route.keys,
-                                    params: route.params,
-                                    dispatch: dispatch
-                                });
-                            }
-                        });
-                        mojito.routes[method] = cache;
-                    }
+            // Do this only once
+            if (!init) {
```
ericf: I think all of this can be removed but using the the approach [here in `router.js`](https://github.com/yahoo/mojito/pull/1233/files#r6080885).

lib/dispatcher.js
```diff
@@ -160,12 +150,12 @@ module.exports = {
...
-        // - page.routes references the Express-like routes object
-        //   for backward compatiblity, this should be converted back the 
-        //   pre-next format.
-        // TODO: in future, Mojito will simply return the Express-like
-        // routes object.
-        outputHandler.page.routes = mojito.routes;
+        if (!CACHE.routes) {
+            // TODO: this could be problematic if the app adds additional
+            // routes at runtime.
+            CACHE.routes = app.getRouteMap();
```
ericf, line 148: I'd hope no one would run into this issue. But if we're worried about it, we could always override `app._router.route()` to blow alway the cache, this way when the next request comes in, the app's route map will be re-generated and cached.

ericf:
@imalberto this is looking good!

Looking at `route-maker.common.js` I was hoping that more of the APIs from `express-annotations` and `express-map` could be used, then I noticed the `.common.js`. How much of Router Maker is used client-side?

I think it would be good to talk about the annotation strategy, i.e. using simple annotation values instead of objects, like `dispatch`, instead there could be separate annotations for `call`, `params`, `options`. Also, I was thinking that `express-map`'s `pathTo()` factory could be used to by the app to create a URL generator function that could be passed to Route Maker.

lib/router.js
imalberto: There are 2 ways to register routes: using `app.VERB` and via `routes.json`. In this case, the annotations is Mojito specific, and there is no need to ask applications to attach some "annotations" that they are not going to use but is needed by the fw. By doing so, registering a route is same as any Express app:

`app.get('/foo', dispatcher('foo.index'));`

Whatever data the  `dispatcher` needs is kept private from the app.

lib/dispatcher.js
imalberto: Synced up with @ericf on this. Please see [above](https://github.com/yahoo/mojito/pull/1233/files#r6080885) in `router.js`.

lib/dispatcher.js
imalberto, line 148: @ericf Thanks for the suggestions. For this use case, it will be required for the app to do something like `dispatcher.resetRoutes()` as a prerequisite before modifying the routes configuration.

imalberto:
@ericf Re: your [previous](https://github.com/yahoo/mojito/pull/1233#issuecomment-23546468) comment, will look into how Mojito can leverage more of `express-map` and `express-annotations` in upcoming PR to address how the `pathTo` functionality can be exposed to handlebars helper (both client and server runtime).

lib/mojito.js
```diff
@@ -234,8 +241,28 @@ Mojito.prototype._configureAppInstance = function (app, store, options) {
         options: options,
         _app: app, // reference to `app` from within `app.mojito`
         attachRoutes: librouter.attachRoutes
-        // `routes` property will be lazy initialized on the first request
-        // @see lib/dispatcher.js
+    });
+
+    // Sugar example usage:
```
caridy: Can you mark this as experimental?

lib/mojito.js
```diff
@@ -234,8 +241,28 @@ Mojito.prototype._configureAppInstance = function (app, store, options) {
...
+    //
+    //     app.mojito.get('foo', '/foo', 'foo.index', params);
```
caridy: I was thinking on something less sweet! something like:

```
app.mojito.get('foo', '/foo', middleware1, middleware2, dispatch('foo.index', { params }))
```

lib/router.js
```diff
@@ -161,26 +163,18 @@ module.exports = {
...
-                    }
-                }
-            }
+            Object.keys(routes).forEach(function (name) {
+                var route = routes[name];
+                Object.keys(route.verbs).forEach(function (verb) {
+                    debug('[%s %s] installing handler', route.call, route.path);
+                    verb = verb.toLowerCase();
+
+                    app[verb](route.path, dispatcher.dispatch(route.call, route.params));
```
caridy, line 173: if this block is not going to use the sugar method, better to remove the sugar then until people start asking for it :)

lib/mojito.js
caridy: or just remove it for now.

lib/app/autoload/util.common.js
```diff
@@ -608,6 +612,45 @@ YUI.add('mojito-util', function(Y, NAME) {
...
+         *         verb: 'get',
+         *         call: 'foo.index'
+         *     }
+         *  
+         * @method decodeRouteName
+         * @public
+         * @param {String} key
+         * @return {Object}
+         */
+        decodeRouteName: function(key) {
```
caridy: why do we need this? this is a one way operation to generate a token, but the token will never be use to generate the original entries used to generate the token. remove.

lib/app/autoload/util.common.js
```diff
@@ -125,6 +125,10 @@ YUI.add('mojito-util', function(Y, NAME) {
             // is found via different indexing models. If we got anything else
             // we can just return it.
 
+            if (Object.prototype.toString.apply(obj) === "[object RegExp]") {
```
caridy, line 128: what is this?

lib/app/autoload/route-maker.common.js
```diff
@@ -313,29 +191,63 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
+        Finds a route for a given uri and verb.
 
         @method find
+        @deprecated
         @param {string} url the URL to find a route for.
         @param {string} verb the HTTP method.
-        @return {Object} the route configuration object
+        @return {Object} the route object or null
         **/
         find: function(uri, verb) {
```
caridy: please remove!

lib/app/autoload/route-maker.common.js
```diff
@@ -73,213 +73,90 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
+        @return {String} the generated path uri for the query, or null
         **/
         make: function(query, verb, params) {
-            // Y.log('make(' + query + ', ' + verb + ')', 'debug', NAME);
-            var my = this,
-                parts = query.split('?'),
+            var parts = query.split('?'),
                 call = parts[0],
-                residual = {},
+                residual = { },
```
caridy: {}

lib/app/autoload/route-maker.common.js
```diff
@@ -73,213 +73,90 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
         @method make
         @param {String} query string to convert to a URL
         @param {String} verb http method
         @param {Object} params object representing extra querystring
                                params. `query` might have querystring portion
                                portion, in which case they have priority.
-        @return {Object} route object matching the query, or null
+        @return {String} the generated path uri for the query, or null
         **/
         make: function(query, verb, params) {
```
caridy, line 126: is this api really changing or is this just a mismatch in the old doc?

lib/app/autoload/util.common.js
imalberto, line 128: This is to allow the `ac.deploy` addon to allow `regex` to be serialized to the client using `express-state` as part of the cleanse process.

lib/router.js
imalberto, line 173: Will remove the sugar method.

imalberto:
Also, after discussion with @caridy , will also be removing annotations and the `find()` method as part of `route-maker` and `ac.url` addon.

lib/app/autoload/route-maker.common.js
```diff
@@ -42,21 +41,18 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
+    Given a uri, extract the route parametrized paths and store in `route.params`.
 
-    @method doCallReplacement
+    @method resolveRouteParams
     @private
     @param {Object} route
     @param {String} uri
     **/
-    function doCallReplacement(route, uri) {
+    function resolveRouteParams(route, uri) {
```
caridy, line 51: do we really need this? I thought express-map will do all this heavy lifting.

lib/app/autoload/route-maker.common.js
```diff
@@ -158,128 +92,68 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
-                }
-
-                // skip to next one
-                return false;
-            });
-
-            return match;
+        linkTo: function(name) {
+            // TODO How can we leverage express-map#pathTo ?
+            // return (CACHE.routes[name] && CACHE.routes[name].path) || null;
```
caridy, line 108: `require('express-map/lib/pathto')`?

lib/app/autoload/route-maker.common.js
caridy, line 126: can u comment on this question?

imalberto, line 126: `make()` returns a URI, so was probably a mismatch.

lib/app/autoload/util.common.js
caridy, line 128: ok, I never fully understood why the whole escaping of the serialized token. But now that we rely on `express-state`, I will argue that `express-state` is responsible for any escaping process needed to expose data into the client side as part of the html response. /cc @ericf

lib/dispatcher.js
```diff
@@ -16,9 +16,12 @@ Given a `command`, dispatches it to the Mojito engine.
...
 "use strict";
 
-var liburl = require('url'),
+var libqs = require('querystring'),
+    liburl = require('url'),
+    libutil = require('./util'),
     debug = require('debug')('mojito:dispatcher'),
-    OutputHandler = require('./output-handler.server');
+    OutputHandler = require('./output-handler.server'),
+    CACHE = { };
```
caridy: u keep using space in between empty `{}` lol.

lib/app/autoload/route-maker.common.js
imalberto, line 51: Not on the client unfortunately. I mentioned that to @ericf and in the end, we have might a solution that works on both run time, but for now this is needed and is abstracted by `route-maker` with common affinity.

lib/app/autoload/route-maker.common.js
imalberto, line 108: Won't work on the client using this. Was going to implement this method as part of next PR to provide helpers for the built in templates. And it has to run on both run time.

lib/dispatcher.js
```diff
@@ -27,6 +30,13 @@ var liburl = require('url'),
 module.exports = {
 
     /**
+    Provides a way to reset the internal routes configurations
+    **/
+    resetRoutesConfig: function () {
```
caridy, line 35: nice, I bet this is really useful for testing at least!

lib/dispatcher.js
```diff
@@ -37,19 +47,24 @@ module.exports = {
...
         splits = call.split('.');
         matchCall[1] = splits.pop();
         matchCall[0] = splits.join('.');
 
-        // in the future we might want to support a more generic annotation
-        // mechanism, walking all callbacks and producing an array of
-        // annotations, for now looking at the last one is just fine.
+        if (params && typeof params === 'string') {
+            routeParams = libqs.parse(params);
+        }
```
caridy, line 65: if it is not a string, but an object, what do u do with it?

lib/router.js
```diff
@@ -147,12 +146,15 @@ module.exports = {
...
+    @method attachRoutes
+    @public
     @param {String|Array} routeFiles optional absolute paths to `routes.*` 
     **/
     attachRoutes: function (routesFiles) {
         var app = this._app,     // express app
             mojito = app.mojito, // the mojito instance
-            appRoot = mojito.options.root;
+            appRoot = mojito.options.root,
+            key = mojito.Y.mojito.util.encodeRouteName;
```
caridy: confusing, better to just call it encodeRouteName as a local variable, if that's the main objective here.

lib/app/autoload/util.common.js
```diff
@@ -608,6 +612,21 @@ YUI.add('mojito-util', function(Y, NAME) {
...
+         * Abstract the logic on creating a route name based on 
+         * the `verb` and `call` used by the Mojito router.
+         *
+         * @method createRouteName
+         * @public
+         * @param {String} verb 
+         * @param {String} call
+         * @return {String} key
+         */
+        encodeRouteName : function(verb, call) {
```
caridy: linting is needed here, space between func name and :

lib/dispatcher.js
imalberto, line 65: Good catch. Should set `routeParams = params || {};` above.

lib/app/autoload/util.common.js
imalberto: Agreed. Problem is most of those files have `sloppy: true`; At some point those should be removed in favor of lest lint config. The lack of space in this case is to be consistent in that file. Otherwise some function will have space, and others won't . They should all be cleaned up in one go, in a separate PR maybe.

lib/router.js
imalberto: Will rename as suggested.

lib/app/autoload/util.common.js
imalberto, line 128: ok. if @ericf confirms, we can remove it.

caridy, line 128: @drewfish @isao do you guys remember what was the problem with the whole encoding thing?

lib/dispatcher.js
caridy, line 65: I literally made you to get in a trap! Now in line `93` you're augmenting the original `params` object per request, but `params` is actually an app level value, leaking data between requests :), solution will be described in lined `93`. jejeje

lib/dispatcher.js
```diff
@@ -101,7 +90,7 @@ module.exports = {
             command.action = matchCall[1];
             command.context = req.context;
             command.params = {
-                route: req.params || {},
+                route: libutil.extend(routeParams, req.params) || {},
```
caridy:
```
route: libutil.extend({}, routeParams, req.params),
```

will work better :), avoiding the leak.

lib/app/autoload/util.common.js
isao, line 128: If memory serves cleanse/uncleanse was for protecting against XSS in an app's configs that were sent to client. IMO, it shouldn't apply to things the developer has defined. But perhaps there was a concern developers would put user input in the config object destined for the client at runtime.

caridy, line 128: oh, right, I remember now. Well, at this point there is only few things that might be runtime data specific, so we can constrain that to those elements to avoid doing the full sanitizing process on all data we sent. Let's create a card for that @imalberto. Also, how expensive if this operation today?

caridy:
+1

imalberto:
# Unit Test Report Summary

Total Number of Executed Tests: 487
Total Number of Passed Tests: 475
Total Number of Failed Tests: 0
Total Number of Skipped Tests: 12

PR #1236, opened 2013-09-04
expose the runtime YUI instance to RS addons

drewfish:
This is made of the following changes:
- added `getAppConfig()` to `lib/store.js`
- rearranged the server startup to:
  - call new `getAppConfig()` to get YUI configuration
  - create YUI instance
  - create store, passing along the YUI instance
  - rest of server initialization as before
- update `Store.createStore()` to attach the passed YUI instances to the returned store object as `store.runtimeYUI`

This supersedes the non-addon changes in PR #1200 .

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,10 @@ Deprecations, Removals
 
 Features
 --------
+* The `lib/store.js` has a new `getAppConfig()` function. This is a better choice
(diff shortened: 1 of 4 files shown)


caridy:
+1

caridy:
Todo:
- [ ] merge to next (which might be a little more complicated)
- [x] update history entry

PR #1240, opened 2013-09-10
version of PR1236 (hotswap support) for "next" branch

drewfish:


drewfish:
Travis node-0.10 failure looks like a transient failure (failed to get "nopt" npm module).

drewfish:
The failed travis "func" jobs appear to be broken since commit 3e8f18f.

lib/mojito.js
```diff
@@ -84,19 +84,17 @@ Creates a new instance of mojito and attach to `app`.
 
 @method _init
 @protected
-@param {express.application} app optional 
+@param {express.application} app optional
 
 **/
 Mojito.prototype._init = function (app) {
```
imalberto, line 90: The merge for the `_init` function seems to have moved some of the "next" code around. Can you compare with what we have [here](https://github.com/yahoo/mojito/blob/next/lib/mojito.js) ?

imalberto:
LGTM

lib/mojito.js
drewfish, line 90: Yes, this was intentional. The new flow requires that we create the runtime YUI instance before creating the resource store. (The resource store used to be created first.)

Issue #1251, opened 2013-09-25
easy way to get contextualized models


drewfish:
The current `ac.models.get()` takes the YUI module name as the argument. This means that I (as an app developer) need to know ahead of time which YUI module to use for a model.

The feature request is to augment `ac.models.get()` to take the model name (for example, `foo` for `models/foo.server.js`) and to use the correct model for the current context.

drewfish:
As it turns out, the mojit resolution process already has this exact mapping. (It can be found at `ac.command.instance.models`.)

For backwards compatibility we should probably first look in `Y.mojito.models[name]` then fallback to `Y.mojito.models[instance.models[name]]` which implements this feature. (Plus good error checking of course.)

caridy:
Sounds good to me!

PR #1256, opened 2013-09-30
fix issue #1251 easy way to get contextualized models

lzhan:


lib/app/addons/ac/models.common.js
```diff
@@ -38,22 +38,23 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
         get: function (modelName) {
 
             var cache = this._page.models || {},
-                model = this._models[modelName] || cache[modelName];
+                model = this._models[modelName] || cache[modelName],
+                mymodel = Y.mojito.models[modelName] ? modelName : this._instance.models[modelName];
```
drewfish: The variable name "mymodel" isn't very descriptive. I believe that this is intended to hold the YUI module name of the model, so perhaps we can call it "yuiName" instead.

lib/app/addons/ac/models.common.js
```diff
@@ -38,22 +38,23 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
...
-                model = Y.mojito.util.heir(Y.mojito.models[modelName]);
+                model = Y.mojito.util.heir(Y.mojito.models[mymodel]);
 
                 if (Y.Lang.isFunction(model.init)) {
                     // NOTE that we use the same config here that we use to
                     // config the controller
                     model.init(this._instance.config);
                 }
-                this._models[modelName] = model;
+                this._models[mymodel] = model;
```
drewfish: I think that the model should still be registered using the passed in name, that is "modelName".

drewfish:
Still needs a unit test.

drewfish:
Looks good.

tests/func/applications/frameworkapp/usecase/mojits/contextModelMojit/models/mymodel.common.js
```diff
@@ -0,0 +1,34 @@
...
+         * @param callback {function(err,data)} The callback function to call when the
+         *        data has been retrieved.
+         */
+         getData: function(callback) {
+             callback(null, {message: 'Hello Mojito'});
+         }
+
+    };
+
+}, '0.0.1', {requires: []});
```
caridy: break line please

tests/func/applications/frameworkapp/usecase/mojits/contextModelMojit/models/mymodel.common.iphone.js
```diff
@@ -0,0 +1,34 @@
...
+         * @param callback {function(err,data)} The callback function to call when the
+         *        data has been retrieved.
+         */
+         getData: function(callback) {
+             callback(null, {message: 'Hello Iphone Data'});
+         }
+
+    };
+
+}, '0.0.1', {requires: []});
```
caridy: break line please

tests/func/applications/frameworkapp/usecase/mojits/contextModelMojit/controller.common.js
```diff
@@ -0,0 +1,45 @@
...
+
+    };
+
+}, '0.0.1', {requires: [
+    'mojito',
+    'mojito-assets-addon',
+    'mojito-models-addon',
+    'contextModelMojitModel',
+    'IphonecontextModelMojitModel'
+]});
```
caridy: break line please

lib/app/addons/ac/models.common.js
```diff
@@ -38,15 +38,16 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
         get: function (modelName) {
 
             var cache = this._page.models || {},
-                model = this._models[modelName] || cache[modelName];
+                model = this._models[modelName] || cache[modelName],
+                yuiName = Y.mojito.models[modelName] ? modelName : this._instance.models[modelName];
```
caridy: Does this means that if we have two models: `foo.js` and `foo.mobile.js`, which the module names `foo` and `foo-iphone`, and in my controlled I ask for `foo`, the contextualized version will never be used. Is that a possibility? I wonder why the contextualized version of the model does not have the priority over the regular model name?

lib/app/addons/ac/models.common.js
```diff
@@ -38,15 +38,16 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
         get: function (modelName) {
```
caridy, line 38: Update method docs to reflect the workflow and the priority in the lookup process.

lib/app/addons/ac/models.common.js
drewfish: Yes, that's possible. I was thinking we should keep the YUI module name has priority to maintain backwards compatibility. (If BC wasn't a concerns I'd just say to use the model name, and not allow the YUI module name.)

caridy: Ok, let's change it them. The thing is, if you're using the yui module name (which is usually the same as the model name, especially if you rely on NAME argument), then you're not contextualizing this, and by making the change with this PR will NOT break that. It will still work. The only case where BC is not going to be contained, is if you have a different name for your model, in which case the contextualization will resolve it anyways, since it uses the internal registry for the mapping. What do you think?

caridy: Also, I don't like the `yuiName` name for that variable. It is not really the `yui name` but the real model name, the other `modelName` is the synthetic one.

lib/app/addons/ac/models.common.js
```diff
@@ -32,21 +32,23 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
         /**
          * Gets model instance
          * @method get
-         * @param {string} modelName The name of the model.
+         * @param {string} modelName The name of the model or the contextualized model file name,
```
drewfish: This should perhaps say

```
The YUI module name of the model, or the model's filename (minus the affinity and context parts).
If the model's filename is give, a contextualized version of the model will be returned.
```

lib/app/addons/ac/models.common.js
drewfish: I think there is some confusion here.  `NAME` is the YUI module name.  We _do_ expect models to continue registering themselves as `Y.mojito.models[NAME] = ...`.

The "model name" comes from the model's **filename**, so `models/foo.server.js` has a model name of `foo` (but could have a YUI module name of `FooModel` or anything really).

So, the `yuiName` variable really _does_ contain the YUI module name of the model.  This can (and often is) different than the "module name".

caridy: Ok, you're right. All the contextualized references are driven by the filename (e.g. what view to use, or what binder to use, etc). In that case, and based on our discussion about models the other day, I will say that `ac.models.get()` should only be driven by filenames, and not attempt to resolve a model by its internal name (yuiName). So we keep the resolution process consistent across all the `ac.*` members. This is also going to help in mojito-next, and yes, this might be a non-BC change.

drewfish: Cool, if we agree on the non-BC change then yeah we can make it simpler.  (Regardless of what mojito-current does, mojito-next should _only_ have the by-filename lookup.)

HISTORY.md
```diff
@@ -4,15 +4,36 @@ version @VERSION@
...
 
+* **!Backwards-Incompatible Change!** Getting model instance by passing model 
+  YUI module name to `ac.models.get` has been removed.
+
 Features
 --------
 
+* Contextualized model support is added:
+  ac.models.get(modelName) will take the model's logical name (filename minus 
+  the affinity and context parts)and return contextualized model instance.
```
drewfish: space after closing parenthesis

HISTORY.md
```diff
@@ -4,15 +4,36 @@ version @VERSION@
...
+  ac.models.get(modelName) will take the model's logical name (filename minus 
+  the affinity and context parts)and return contextualized model instance. 
+
+  For example: 
+  Different model files can be provided for different context settings under models directory:
+  .../models/
+             mymodel.common.iphone.js
+             mymodel.common.js
+
+  Under certain context, calling ac.models.get('mymodel') from controller will
```
drewfish: This sentence is a little clunky.  Perhaps:

```
Calling `ac.models.get('mymodel')` wil return the model which is appropriate for the current context.
```

examples/developer-guide/binding_events/mojits/PagerMojit/binders/index.js
```diff
@@ -47,36 +47,40 @@ YUI.add('PagerMojitBinder', function (Y, NAME) {
...
-                            for (key in props) {
-                                if (props.hasOwnProperty(key)) {
-                                    if (typeof (props[key]) === 'object') {
-                                        continue;
+                        self.mojitProxy.data.set('imageId', imageId);
+                        self.mojitProxy.invoke('contentModel',{},
+                            function(error, raw){
+                                if (error) {
+                                    Y.log("Error found:"+ error);
+                                }
```
drewfish, line 67: If there is an error I'm not sure we can make any assumptions about the contents of the `raw` variable. So, we might need a `return` after the Y.log.

examples/simple/part8/mojits/myMojit/controller.server.js
```diff
@@ -11,7 +11,7 @@ YUI.add('myMojit', function(Y, NAME) {
         index: function(ac) {
 
 // DOING
```
drewfish: Hmm... this extra comment should probably be removed.  (Alas, looks like something I might have left... :disappointed: .)

lib/app/addons/ac/models.common.js
```diff
@@ -32,21 +32,22 @@ YUI.add('mojito-models-addon', function (Y, NAME) {
         /**
          * Gets model instance
          * @method get
-         * @param {string} modelName The name of the model.
-         * @return {object} model instance, or null.
+         * @param {string} modelName The model's logical name (filename minus the affinity and context parts),
```
drewfish: There's something not quite right about the docs for `modelName`, perhaps just that the comma needs to change to a period.

tests/func/applications/frameworkapp/usecase/mojits/PagedFlickr/views/index.iphone.mu.html
```diff
@@ -1,5 +1,5 @@
 <div id="{{mojit_view_id}}">
-    <h1>{{ greeting }}</h1>
+    <h2>{{ greeting }}</h3>
```
drewfish: Tags are unbalanced:  h2 opens but h3 closes.

tests/func/usecases/usecasestest_descriptor.json
```diff
@@ -13,88 +13,105 @@
...
-                "group" : "smoke,usecases",
+            "hello" : {
+                "group" : "smoke,usecases,hello",
                 "params" : {
                     "test" : "testhello.js",
                     "page" : "$$config.baseUrl$$/hello-1/index"
                 }
             },
+            "contextmodel" : {
+                "group" : "smoke,usecases,lichun",
```
drewfish: Hmmm... I curious why you are particularly interested in this test... :smile:

drewfish:
+1

PR #1264, opened 2013-10-16
better output handler response header check

jarnoux:
The output handler was maintaining its own guard to prevent writing headers when they are already sent. This is not an efficient solution as the headers can potentially have been written by virtually anyone that has access to the response. This was causing our app to crash and stop when we had two errors.

### DIFF
--- a/lib/output-handler.server.js
+++ b/lib/output-handler.server.js
@@ -123,7 +123,7 @@ OutputHandler.prototype = {
 
 
     _writeHeaders: function() {
-        if (!this.headersSent) {
+        if (!this.res.headersSent) {
 
             // passing a falsy reason phrase would break, because node uses every non-string arguments[1]
             // as header object/array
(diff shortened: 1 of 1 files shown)


caridy:
LGTM, although the functional tests are failing, but might not be related to this PR. /cc @lzhan can you take care of this?

lzhan:
@caridy  LGTM too. Failed functional tests are due to existing YQL/flickr issue. I have run this pr with the branch with fixed functional app/tests, everything is fine. 
@jarnoux would you please provide a unit or functional test for this pr?

caridy:
@lzhan, a unit test for this is going to be difficult because @jarnoux removed the piece of logic that we controlled in favor of something controlled by express directly. What we might need to do is to review if we were doing any unit test around that manual verification, and remove them.

In the other hand, if we want to really test this, we might want to use `supertest` which allow us to fake an express request, but I think for now we are good.

lzhan:
@caridy Got it. Thanks. I will update HISTORY file and merge this pr.

PR #1268, opened 2013-10-24
fix arrow version to 0.0.82

lzhan:
Fix arrow version until this bug is resolved:
http://bug.corp.yahoo.com/show_bug.cgi?id=6593515


drewfish:
+1

PR #1269, opened 2013-10-24
Updated the archetypes so that the models are correctly accessed from the controller

zhouyaoji:


lib/app/archetypes/mojit/full/controller.server.js.hb
```diff
@@ -34,7 +34,5 @@ YUI.add('{{name}}', function(Y, NAME) {
                 });
             });
         }
-
     };
-
-}, '0.0.1', {requires: ['mojito', 'mojito-assets-addon', 'mojito-models-addon', '{{name}}ModelFoo']});
+}, '0.0.1', {requires: ['mojito', 'mojito-assets-addon', 'mojito-models-addon']});
```
drewfish: Have you tried this?  I think we might still need the `requires` to have `{{name}}ModelFoo`.

zhouyaoji: Yes, I did. I talked to Isao, and he said that `mojito create` uses the archetypes in the `mojito-cli-create` package, so the archetypes in the `mojito` source are obsolete. I'm closing this PR and opening a new one that has removed the `archetypes` directory to avoid future confusion.

PR #1270, opened 2013-10-24
Removed archetypes, fixed an error in the binding example, updated docs to point to `archetypes` directory in `mojito-cli-create`.

zhouyaoji:


drewfish:
Travis has 4 failing tests, not sure if they're related though.

lzhan:
+1

PR #1271, opened 2013-10-28
[docs] Added a section to the Mojito overview about the differences between Mojito and other Web app architectures.

zhouyaoji:


docs/dev_guide/intro/mojito_overview.rst
```diff
@@ -111,8 +111,69 @@ Localization and Internationalization
...
+Frameworks such as Rails and Django allow you to connect directly to a database through
+object-relational mapping (ORM) with no intermediate service layer. With Mojito, you
+do not directly connect to databases, but instead make HTTP requests to an existing
+service, thereby, the separating the application code completely from database queries.
+
+Mojito applications can use `YQL <http://developer.yahoo.com/yql>`_ to make requests to 
+other Web services to get data as well, which gives the application code more control over 
+the requested data, but this again is not a direct database request as YQL is a RESTful 
+proxy that will make HTTP requests to other Web services based on rules defined in XML tables. 
+
```
clarle, line 173: If you're not using YQL, we should recommend using the Mojito REST Library to make calls to RESTful APIs:

http://developer.yahoo.com/cocktails/mojito/docs/api_overview/mojito_rest_lib.html

This would also work on both client and server, so it lets you do the same thing as YQL if you'd rather work with more standardized RESTful API routes.

zhouyaoji, line 173: Thanks, @clarle, I'll add that to the documentation.

PR #1272, opened 2013-10-28
Pin express-state and express-map version to specific version for now

imalberto:
`express-map` is introducing some non-BC changes in their next release and mojito-next will have to be updated.


imalberto:
Reverting this PR as it is not needed.

PR #1275, opened 2013-10-30
[example] Added mocked models, updated app config to define selector, and used middleware to get request context.

zhouyaoji:


zhouyaoji:
@lzhan Updated trib app to use live data by default but mocked data when the query string parameter `environment=development` is passed to the app.

Issue #1276, opened 2013-10-31
Funtional tests failed on node8 after pr #1264 is merged


lzhan:
4 tests failed. All of them has:
ac.flush("Hello, world!--from flush,");
ac.done("Hello, world!--from done");

error: (mojito-dispatcher): Error from dispatch on instance 'ACMojit':
error: (mojito-dispatcher): Can't render headers after they are sent to the client.
............
error: (outputhandler.server): { [Error: Can't render headers after they are sent to the client.] code: 500 }

PR #1277, opened 2013-10-31
fix issue #1276

lzhan:
resolve functional tests failures on node8 after pr1264 is merged.

### DIFF
--- a/lib/output-handler.server.js
+++ b/lib/output-handler.server.js
@@ -123,7 +123,7 @@ OutputHandler.prototype = {
 
 
     _writeHeaders: function() {
-        //support both node8 and node10
+        //support both node8 and node10, res.headersSent was introduced in node10.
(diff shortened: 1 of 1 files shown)


lib/output-handler.server.js
```diff
@@ -123,8 +123,10 @@ OutputHandler.prototype = {
 
 
     _writeHeaders: function() {
-        if (!this.res.headersSent) {
+        //support both node8 and node10
```
drewfish: I'd maybe say something addition, such as "res.headersSent was introduced in node10".

drewfish:
+1

caridy:
Let's make sure we add a trello card, or an issue in github to remove the unnecessary code once we drop support for 0.8

lzhan:
@caridy Thanks for the reminder. I am adding a card now.

PR #1278, opened 2013-11-01
[examples] Added middleware and updated app configs so that apps use mocked models by default.

zhouyaoji:


lzhan:
I like the idea adding customized contextualizer. 

I think the default context: environment:development, should be used for non-mocked data, so that there is no change for people who run these examples and there is no doc change needed for this update/pr.

zhouyaoji:
@lzhan I updated the middleware to only use the `environment:development` context when given the query string `environment=development`, so the apps will run in the default context and get live data. For the tests, I updated the descriptors so that the applications are pinged with the `environment=development` query string parameters and then return the mocked.

lzhan:
@zhouyaoji I don't see middleware/mojito-contextualizer.js files. Maybe you forgot to add them?

lzhan:
+1

PR #1286, opened 2013-11-13
Handle invalid context exceptions with output handler

gomezd:
The `validateContext` method of the resource store throws exceptions for invalid contexts, however the dispatcher doesn't handle these exceptions and the app crashes.

```
Error: INVALID dimension key "foo"
    at RSAddonSuperBundle.Y.extend.validateContext (/git-data/common/utils/mojito-dimensions-base/addons/rs/super-bundle.server.js:198:33)
    at ResourceStore.<anonymous> (/git-data/apps/search-apps-news/node_modules/mojito/node_modules/yui/oop/oop-min.js:1:2230)
    at i.Method.exec (/git-data/apps/search-apps-news/node_modules/mojito/node_modules/yui/event-custom-base/event-custom-base-min.js:1:1161)
    at ResourceStore.o.(anonymous function).r.(anonymous function) [as validateContext] (/git-data/apps/search-apps-news/node_modules/mojito/node_modules/yui/event-custom-base/event-custom-base-min.js:1:616)
    at Object.Y.namespace.Dispatcher.dispatch (/git-data/apps/search-apps-news/node_modules/mojito/lib/app/autoload/dispatch.server.js:125:19)
    at Object.dispatcher [as handle] (/git-data/apps/search-apps-news/node_modules/mojito/lib/mojito.js:363:41)
    at next (/git-data/apps/search-apps-news/node_modules/mojito/node_modules/express/node_modules/connect/lib/http.js:204:15)
    at Object.handle (/git-data/apps/search-apps-news/node_modules/mojito/lib/app/middleware/mojito-router.js:101:13)
    at next (/git-data/apps/search-apps-news/node_modules/mojito/node_modules/express/node_modules/connect/lib/http.js:204:15)
    at run (/git-data/apps/search-apps-news/node_modules/mojito/lib/app/middleware/mojito-handler-tunnel.js:34:24)

//client sees this
Internal Server Error
```

With this small change the dispatcher handles gracefully invalid context errors, and gives users the ability to implement their own version of validateContext and throw custom errors, in our case some context dimensions would result in a "Bad request" 400 error instead of the default 500.

Instead of crashing we would get:

```
error: (outputhandler.server): { [Error: INVALID dimension key "foo"] code: 400 }
info: (outputhandler.server): done

// client sees this
Error: 400

Error details are not available.
```

### DIFF
--- a/tests/unit/lib/app/autoload/test-dispatch.server.js
+++ b/tests/unit/lib/app/autoload/test-dispatch.server.js
@@ -7,7 +7,7 @@ YUI.add('mojito-dispatcher-server-tests', function(Y, NAME) {
 
     var suite = new Y.Test.Suite(NAME),
         A = Y.Assert,
-        dispatcher = Y.mojito.Dispatcher,
+        dispatcher,
         store,
         command,
         adapter;
@@ -17,6 +17,7 @@ YUI.add('mojito-dispatcher-server-tests', function(Y, NAME) {
         name: 'dispatch',
 
         'setUp': function() {
+            dispatcher = Y.mojito.util.heir(Y.mojito.Dispatcher);
             store = {
                 getStaticAppConfig: function() {
                     return { yui: {} };
@@ -57,6 +58,7 @@ YUI.add('mojito-dispatcher-server-tests', function(Y, NAME) {
         },
 
         'tearDown': function() {
+            dispatcher = null;
             store = null;
             command = null;
             adapter = null;
@@ -119,8 +121,7 @@ YUI.add('mojito-dispatcher-server-tests', function(Y, NAME) {
 
         'test dispatch with valid controller': function () {
             var tunnel,
-                acCommand,
-                _createActionContext = dispatcher._createActionContext;
+                acCommand;
 
             errorTriggered = false;
             dispatcher.init(store, tunnel);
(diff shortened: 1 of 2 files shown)


caridy:
LGTM, we just need the unit test for this, and we can merge it.

gomezd:
Done.

tests/unit/lib/app/autoload/test-dispatch.server.js
```diff
@@ -173,6 +173,36 @@ YUI.add('mojito-dispatcher-server-tests', function(Y, NAME) {
             dispatcher._createActionContext = _createActionContext;
         },
 
+        'test dispatch with invalid context': function () {
+            var tunnel,
+                adapterError,
+                adapterErrorCalled,
+                _createActionContext = dispatcher._createActionContext;
```
caridy: I will argue that this is a back practice when it comes to mock members of an object. the problem here is that if there is an exception within the body of the function that is doing the mocking, the original method will not be restored, and successive tests might break as a consequence of it.

normally, what I recommend here is to use the `setUp` and `tearDown` to store and restore the methods that might get mock in the individual tests, guaranteeing that each test will always get the right original method.

drewfish: You might also want to create a new dispatcher for every test (in the `setUp`). To do this you might need to use `Y.mojito.util.heir(Y.mojito.Dispatcher)` so that you're using a separate object for each test.

gomezd:
Done suggested changes to unit tests.

caridy:
+1

@lzhan can you merge this?

@gomezd is the test in travis related to this test? or something else?

gomezd:
@caridy Yes, what is running in travis is for this PR

lzhan:
@gomezd @caridy I am checking why travis build failed on node8. There are three tests on dashboard example failed.

PR #1288, opened 2013-11-15
[examples/doc] Added the statichandling object to the app config and updated the configure_routing example (and docs) to work with `next`. 

zhouyaoji:


lzhan:
+1

Issue #1254, opened 2013-09-26
Allow the same context to be used multiple times in configuration


jlecomte:
Our application configuration is spread across multiple files (we use the "applicationConfigFiles" directive to include additional configuration files into our main application.yaml...) Unfortunately, when the same context is specified multiple times, e.g. "runtime:server,lang:fr-FR", mojito throws an error:

```
error: (addon-rs-config): 
settings "runtime:server,lang:fr-FR" exists in both 
/blah/blah/blih/base.yaml and
/blah/blah/blah/application.yaml
```

Because the order matters in the "applicationConfigFiles" directive, we should be allowed to use the same context multiple times. If the same configuration setting is set multiple times, the last one should take precedence...

drewfish:
Should the last one take precedence, or merge over an earlier one?  The merge allows a particular context to be "built up" from multiple files.  (Alas, merging arrays is problematic.)

jlecomte:
merge over = take precedence. I don't see the difference...

drewfish:
I guess I was trying to make a distinction between "replace" and "merge".  You seem to have been thinking of "merge" too, so I think we have the same idea :smile:.

jlecomte:
Can we prioritize this request? We would like to have it by the end of November/early December. Thanks!

caridy:
bumping this into trello board: https://trello.com/c/XYLOloRT/768-search-gh-1254-allow-the-same-context-to-be-used-multiple-times-in-configuration

drewfish:
Available in mojito 0.8.1.

PR #1290, opened 2013-11-18
fix issue #1254 Allow the same context to be used multiple times in configuration

lzhan:
The unit test will not pass until the https://github.com/yahoo/ycb/pull/21 is merged and ycb is updated.


lib/app/addons/rs/config.js
```diff
@@ -222,11 +222,6 @@ YUI.add('addon-rs-config', function(Y, NAME) {
                         Y.log('missing "settings" in YCB file: ' + path, 'error', NAME);
                         return;
                     }
-                    if (settings[section.settings]) {
-                        Y.log('settings "' + section.settings + '" exists in both \n' + path +
-                                ' and\n' + settings[section.settings], 'error', NAME);
-                        return;
-                    }
                     settings[section.settings] = path;
```
drewfish: If we're no longer doing `if (settings[section.settings])` then I think we can drop this assignment (`settings[section.settings] = path`) as well.

caridy: yeah, this assignment sounds a little bit weird for me as well.

drewfish:
We might want to hold off on this until a new version of ycb is available, and then update package.json to depend on that.  (Otherwise PR is looking good.)

drewfish:
Hmmm... I wonder if there's some way we can get travis to re-test this (now that ycb@1.1.0 has been released).

lzhan:
The last travis run is with ycb 1.1.0. The unit test is fine. Func tests failed on node10 with dashboard app. I don't think it is related to this pr.

drewfish:
+1

PR #1296, opened 2013-11-21
update arrow to v0.0.86

lzhan:
Mojito integration job passed with arrow 0.0.86. 
Update mojito to use this verison of arrow so that we can continue setup Mojito job on node10.


drewfish:
travis passed, +1

Issue #1280, opened 2013-11-01
Composite.execute cannot be called multiple times in the same mojit


aljimenez:
The following use of composite.execute results in a crash since the Y.Parallel queue that is used internally is not cleared after execution is done.

``` javascript
ac.composite.execute(config, function (data, meta) {
    ac.composite.execute(newConfig, function (data2, meta2) {
        ...
    });
});
```

There are many valid use cases for this. In my case I would like to execute a mojit which then accumulates some data, which I then need to execute other mojits. One easy fix in composite.common.js would be to reset the Y.Parallel queue before calling the callback towards the end of the execute method. That way subsequent executions would use a new queue:

``` javascript
self.queue = new Y.Parallel({
    context: self
});

cb(content, meta);
```

drewfish:
It seems reasonable to me to want to use ac.composite.execute() multiple times like you suggest.

clarle:
Issue resolved in #1300, and released in 0.8.2. 

Thanks for the report @aljimenez!

PR #1300, opened 2013-11-27
fix issue #1280 Composite.execute cannot be called multiple times in the...

lzhan:
... same mojit


tests/unit/lib/app/addons/ac/test-composite.common.js
```diff
@@ -112,6 +112,16 @@ YUI().use('mojito-composite-addon', 'test', function(Y) {
...
                 A.areSame('kid_a__meta', meta.kid_a, "wrong kid_a meta");
                 A.areSame('kid_b__meta', meta.kid_b, "wrong kid_b meta");
+
+                config1 = {
+                    children: {
+                        kid_c: { id: 'kid_c', type: 'kidc' }
+                    }
+                }
+                c.execute(config1, function(data1, meta1) {
+                    A.isString(data1.kid_c, "missing kid_c data");
```
drewfish: I think we also need to make sure that neither `data1` nor `meta1` has information from `kid_a` or `kid_b`.  (Said another way, the results of the first call to `execute()` shouldn't be mixed into the results of the second call.)

lib/app/addons/ac/composite.common.js
```diff
@@ -226,6 +226,10 @@ callback({
                     ac.assets.mixAssets(meta.assets, cfg.assets);
                 }
 
+                this.queue = new Y.Parallel({
```
caridy, line 228: yes, as @drewfish mentioned before, this is going to be problematic because you might call `execute` mutliple times even when the previous `execute()` is not done. So, what I will propose is to create the new `this.queue` object right after calling `this.queue.done()` in line 199. That means everytime you call `execute()` the current queue will get in motion, and will become inaccessible automatically, a new queue object will be create. there is not need to keep a hold on the old queue, what matters is the callback when it is done or when an error occur, and that we have it anyway.

tests/unit/lib/app/addons/ac/test-composite.common.js
```diff
@@ -100,8 +100,29 @@ YUI().use('mojito-composite-addon', 'test', function(Y) {
                         kid_b: { id: 'kid_b', type: 'kidb' }
                     }
                 },
+                config1 = {
```
caridy, line 104: I suspect that some of the children data will be mixed, can you make sure that in all tests you are getting what you're executing in the callback, without leaks between execute commands?

caridy:
I'm still not convinced that this is right. I think it works, and the tests passes because the mock for `ac.done()` on every child is synchronous, but if you add a `setTimeout()` there, it will fail. Can we try it?

lzhan:
Sure. Let me try it.

caridy:
@lzhan I still have my reservations about this. I think the problem is there somewhere, we are just not seeing it. The good news is that this is not going to make it worst, in fact it fixes the reported issue. So, I will give you a +1 so we can move on, but make sure that you open an issue and a trello card to solve the remaining issue, at least to investigate how to deal with multiple calls to execute while the callbacks will be executed in random order.

caridy:
+1

lzhan:
Card is created: https://trello.com/c/dxxyVWTZ/806-investigate-how-to-deal-with-multiple-calls-to-execute-while-the-callbacks-will-be-executed-in-random-order

Issue #1306, opened 2013-12-13
ac.partial.render() throws error for handlebar custom helper


jithine:
I have a handlebar helper which is added via

```
 ac.helpers.set('generateSelect', generateSelectHelper);
```

When running 

```
148             ac.partial.render(newData, 'new-tasks', function (err, html) {
149                 Y.log(err);
150                 Y.log(html);
151                 ac.done(html);
152             });
```

I get this error  

> Uncaught Error: Could not find property 'generateSelect' 

Content of new-tasks.hb.html

```
    1 <div id="{{mojit_view_id}}">
    2     {{#each data}}
    3         {{> task }}
    4     {{/each}}
    5 </div>
```

Contents of ./partials/task.hb.html

```
...
  4             <div class="content-row task-data pure-g">
  5                 <div class="drop-down-container pure-u-1-6">
  6                     {{{generateSelect tasks "task"}}}
  7                 </div>
  8            </div>
...
```

On an additional note ac.partial.render() does not accept a handle bar view partial located inside partials/ folder which restricts it's use. Eg:

```
 ac.partial.render(newData, './partials/task', ...
```

Gives an error saying cannot find ./partials/task'

caridy:
Hey, this is a limitation of handlebars and not an issue in mojito. Handlebars will not propagate context from parent to a partial, that's why all functions and data available at the regular template will not be available into the partial unless you use `../something` to reference to the parent context. Here is the discussion about the actual fix in the next version of handlebars, which includes some of the workaround:

https://github.com/wycats/handlebars.js/issues/182

jithine:
@caridy But it does work when I use ac.done() and the associated view uses the partial. ie instead of ac.partial.render () this works

```
150             ac.done({data : newData }, {
151                 view :  {
152                     name: "new-tasks"
153                 }
154             });
```

Then what is the purpose of ac.partial ? From the naming it suggests that it's for manipulating partials but it cannot load html inside partials directory. Documentation - http://developer.yahoo.com/cocktails/mojito/api/classes/Partial.common.html#method_render - says for ac.partial.render() view has to be views folder, which makes me conclude that ac.partial is not for handlebar partials.

My use-case if to instantiate this partial with new data, like in YUI template - http://yuilibrary.com/yui/docs/template/

caridy:
[...]

PR #1308, opened 2013-12-20
Fix issue #1306: ac.partial.render() throws error for handlebar custom helper

lzhan:


caridy:
@lzhan are u still working on this one?

lzhan:
Not really. But I just noticed that I need to fix the travis failure. About the issue with "partials" as dir name, if we don't want to make any change, we might need to document it.

caridy:
+1

caridy:
this solves issue #1306

PR #1310, opened 2014-01-07
rm time checking as no control on testing machine time

lzhan:
test failed on grid because of the time difference.


drewfish:
+1

Issue #1312, opened 2014-01-08
request for MojitProxy.refreshView() not to reload css-assets


ItsAsbreuk:
I have composite mojits that are refreshedViewed() multiple time with different childmojit configuration.

the css of this mastermojit is reloaded a huge amount of time now. It would be great if the css-assets wre not reloaded. Either by default, or as an option, like mojitProxy.refreshView({fetchCSS: false})

Best regards,
Marco.

ItsAsbreuk:
where fetchCSS=false not applies to the childmojits that haven't been loaded yet - of coarse..

drewfish:
I think mojito should be smart enough to not reload a CSS asset that has already been loaded on the page.

ItsAsbreuk:
i agree. just keep in mind that we can do a refreshView with param-options where the param-options make the controller to serve different assets. so the logic should track css-files instead of mojits-actions.

caridy:
@ItsAsbreuk, are you using `shaker`?

ItsAsbreuk:
@caridy no not for these current projects
(i will activate shaker in a couple of week though)

ItsAsbreuk:
Here is a screenshot of my system: 

http://projects.itsasbreuk.nl/screenshot.jpg

In the jpg-image you see in the inspector that index.css is loaded twice.
(which i can confirm when inspecting the actual html)

caridy:
alright, let's get this fixed. @ItsAsbreuk if you can provide a very basic repro case, that will be awesome.

/cc @lzhan can you investigate more about this?

ItsAsbreuk:
Yeah sure
I can build a repo this weekend.
Let you know.

lzhan:
@caridy @ItsAsbreuk I will work on it.

lzhan:
@ItsAsbreuk @caridy I can reproduce this with our existing test app.

caridy:
@lzhan was this fixed?

lzhan:
The pull request with solution is closed. Long term solution needs to be discussed.

caridy:
Ok, I will close this one, and you can open a new issue for the long term solution.

PR #1313, opened 2014-01-22
Fix issue #1312: request for MojitProxy.refreshView() not to reload css-assets

lzhan:


lib/app/autoload/output-handler.client.js
```diff
@@ -27,16 +27,27 @@ YUI.add('mojito-output-handler', function(Y, NAME) {
         },
             done = {},
             blobNode = '',
-            doneChecker;
+            doneChecker,
+            loadedcssele = document.getElementsByTagName("link");
```
imalberto, line 31: Might be preferable to rely on YUI instead: `loadedcssele = Y.all('link');`

caridy, line 31: I don't think so, because `Y.all` returns a fixed array while `getElementsByTagName()` returns a dynamic collection that changes overtime with new `link` tags.

caridy:
In general, this part of the code is very fragile, and we should be able to improve it drastically using modules, but that will requires more time. Let's discuss it offline.

caridy:
@lzhan ping me tomorrow to discuss this

lzhan:
Ok. Will do. Thanks.

PR #1321, opened 2014-01-30
Required changes to support Search App

imalberto:
CHANGELOG
- added ac.url.find() API to support mojito-\* related dependencies
- added ac.url.linkTo() helper to support url generation based on route name
- lib/mojito.js now integrates with expressjs via "Branding"
- lib/middleware : app specific middleware are now passed a `midConfig` object during registration
- lib/router : uses `getStaticAppConfig().routesFiles` as the source to load routes configuration

TODO
- [x] fix unit tests
- [x] fix functional tests
- [x] update examples
- [ ] update `next` docs


lib/dispatcher.js
```diff
@@ -64,13 +64,6 @@ module.exports = {
             routeParams = libqs.parse(params);
         }
 
-        // now looking at the last callback with the appropriate "annotations"
-        function extractAnnotations(route) {
```
imalberto, line 68: Unused function from previous revisions. Removed in this PR.

lib/app/autoload/route-maker.common.js
```diff
@@ -97,22 +97,126 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
+
+            // Extract the url paramterized paths and store in `params`
+            copy.params = copy.params || {};
+            len = copy.keys.length;
+            for (i = 0, j = 1; i < len; i += 1, j += 1) {
+                key = copy.keys[i];
+                copy.params[key.name] = matches[j];
+            }
+
+            // Lookup the `call` by aliases
```
imalberto, line 111: Not too happy about this block about how the `call` is decoded.

lib/mojito.js
```diff
@@ -58,6 +48,15 @@ var debug = require('debug')('mojito'),
 //  Mojito Global
 //  ----------------------------------------------------------------------------
 
+function extend(app, options) {
```
imalberto, line 49: With this change, now `app.js` can do something like this:

```
var express = require('express'),
      libmojito = require('mojito'),
      app = express();

libmojito.extend(app, {
    context: {
        runtime: 'server',
        environment: 'staging',
        colo: 'sp1'
    }
});

// proceed as usual
... ....
```

caridy, line 49: why is the `port` needed?

imalberto, line 49: `port` is not needed, since the actual `listen` is done at the `app.js` level. Good point. Will be removing references to `port` in next commit.

lib/app/autoload/route-maker.common.js
```diff
@@ -97,22 +97,124 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
+
+            // Return clone to caller
+            copy = Y.mojito.util.copy(route);
+
+            // Extract the url paramterized paths and store in `params`
+            copy.params = copy.params || {};
+            len = copy.keys.length;
+            for (i = 0, j = 1; i < len; i += 1, j += 1) {
+                key = copy.keys[i];
+                copy.params[key.name] = matches[j];
```
drewfish, line 108: If `j` is just one more than `i` it might make more sense to do `= matches[i + 1]` instead.

imalberto:
@caridy PR is ready to merge. The unit and func tests are OK when running individually , but failing as a whole when running of the entire suite, related to the issues @lzhan is investigating using the Locator feature from Arrow.

Once this PR is merged, @zhouyaoji can then merge his PRs related to examples and docs (since they've changed). See #1324 #1327

examples/developer-guide/adding_view_engines/app.js
```diff
@@ -15,10 +15,12 @@ var debug = require('debug')('app'),
     app;
 
 app = express();
+app.set('port', process.env.PORT || 8666);
+mojito.extend(app);
 
-app.use(mojito.middleware());
+app.use(app.mojito.middleware());
```
caridy: middleware are monadic functions that has nothing to do with the state of the app. I wonder why you decided to move this from `mojito.middlware` to `app.mojito.middleware()`?

examples/developer-guide/adding_view_engines/app.js
```diff
@@ -15,10 +15,12 @@ var debug = require('debug')('app'),
...
 app.mojito.attachRoutes();
-app.post('/tunnel', mojito.tunnelMiddleware());
+app.post('/tunnel', app.mojito.tunnelMiddleware());
```
caridy: same as above.

examples/developer-guide/dashboard/07_binders/app.js
```diff
@@ -11,10 +11,14 @@
...
-    mojito = require('../../../../'),
-    app;
+    libmojito = require('../../../../'),
+    app,
+    mojito;
 
 app = express();
+app.set('port', process.env.PORT || 8666);
+libmojito.extend(app);
+mojito = app.mojito;
```
caridy: I don't like this.

lib/app/addons/ac/url.common.js
```diff
@@ -30,6 +30,26 @@ YUI.add('mojito-url-addon', function(Y, NAME) {
     UrlAcAddon.prototype = {
 
         namespace: 'url',
+        /**
+        Finds the first matching route from the given URL
+        @method find
+        @param {String} url the URL to find a route for.
+        @param {String} the HTTP method. Default value is "get".
+        **/
+        find: function(url, verb) {
```
caridy, line 39: I thought we were going to move this into the corresponding pkg that requires it.

lib/app/autoload/route-maker.common.js
```diff
@@ -92,27 +60,131 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
         //
 
         /**
+        Finds a route for a given method+URL
+        @method find
+        @param {String} url the URL to find a route for.
+        @param {String} verb the HTTP method.
+        @return {Object} the matched route object, or null
+        **/
+        find: function(uri, verb) {
```
caridy, line 73: same as above.

lib/app/autoload/route-maker.common.js
```diff
@@ -92,27 +60,131 @@ YUI.add('mojito-route-maker', function (Y, NAME) {
...
         @method linkTo
         @param {String} name
+        @param {Object} context
         @return {String|null} path
         **/
-        linkTo: function(name) {
-            // TODO How can we leverage express-map#pathTo ?
-            // return (CACHE.routes[name] && CACHE.routes[name].path) || null;
-            return null;
+        linkTo: function(name, context) {
```
caridy: why is the context needed here? is this the mojito context?

lib/middleware.js
```diff
@@ -100,23 +100,46 @@ function middleware() {
...
             mid,
+            midConfig,
+            midName,
             appDir,    // app root directory
             appConfig; // base application config
 
 
         store = mojito.store;
         appConfig = store.getStaticAppConfig();
+        midConfig = {
```
caridy, line 115: I will vote for `libmojito.middleware()` as a middleware function. then within that function having an internal cache routine that triggers on the first request, to be able to pick `app.mojito`, and `app.mojito.store` to collect any app level settings that is needed by the middleware to function.

lib/mojito.js
```diff
@@ -8,29 +8,21 @@
...
+        app,
+        mojito;
 
-    app.post('/tunnel', mojito.tunnelMiddleware()); // setup tunnel dispatcher
+    app = express();
+    libmojito.extend(app, {});
+    mojito = app.mojito;
 
-    app.mojito.attachRoutes(); // load routes defined in `routes.json`
+    app.use(mojito.middleware());
```
caridy: isn't this suppose to happen in the app.js?

lib/mojito.js
```diff
@@ -251,28 +237,20 @@ Mojito.prototype._configureAppInstance = function (app, store, options, appConfi
...
 //  EXPORT(S)
 //  ----------------------------------------------------------------------------
 
-/**
 
-Export Mojito extension as a function.
+module.exports = {
+    // BEGIN DO NOT USE - for testing only
+    // @private
+    Mojito: Mojito,
```
caridy: why? normally, if you do the proper extend (following the principles described by @ericf) you should be able to get a handle on the original class by looking into the defined property.

caridy:
it is looking good. small details here and there, but in general, we are pretty close to get this one merged.

lib/app/autoload/route-maker.common.js
imalberto: I think `params` is perhaps better suited here, since it will be used to resolved any parametrized urls.

lib/middleware.js
imalberto, line 115: @caridy That's the approach Mojito took for all its built-in middleware.

This specific implementation is to provide a level of BC for all other middleware that relies on the internal store, logger, etc. If not, that will require all of them to be updated.

In addition, we can document that going forward, the recommended way is to do as you suggested, and that this is mark a deprecated feature and will be removed in the future. /cc @zhouyaoji

lib/mojito.js
imalberto: You are correct. I think that specific line though is part of the header file documentation showing basic usage for now to extend `app` with `mojito`.

lib/mojito.js
imalberto: Good point!

lib/middleware.js
drewfish, line 115: What about just having mojito (or the user's app.js) just `app.set('store', mojito.store)` instead? Same approach for other midConfig fields as well.  Then each middleware can simply `req.app.get('store')`.

imalberto, line 115: @drewfish I like this idea overall instead of just tagging onto `app.mojito` namespace! The only thing is that `"store"` is very generic, and I am thinking perhaps a more mojito specific NS should be used. e.g. `app.set('mojito store', mojito.store');`.

For this specific issue here, search team have quite a few mojito-\* extensions out there with middleware that relies on `midConfig` being passed to for initialization, and having to delay the initialization until that first request will basically break those. We can provide a level of BC but make sure we document those as _deprecated_ and _not recommented_ for future use, which I think is the least resistant path.

tests/func/serveronly/serveronlytest_descriptor.json
```diff
@@ -550,6 +586,12 @@
                             }
                         },
                         {
+                            "controller": "locator",
+                            "params": {
+                                "value": "#output"
```
lzhan: The element id for this test should be #my_header.
This will fix the travis failure on the serveronly app test.

lib/app/autoload/route-maker.common.js
caridy: so, it is not the context of mojito, just a context object. cool, lets rename it to params.

lib/middleware.js
drewfish, line 115: Yeah, I agree about the change where initialization occurs on the first request.

I agree that app.set('mojito store') is probably better.

As far as search middleware boot-up, remember that they are in control app.js which means that they can load there middleware however they want. (They don't necessarily need to let mojito do it.) Hmmm... the issue with what I'm describing is that we do want sugar for mojito to register the middleware that it knows about (the middleware that the RS found). If they wanted to instantiate (with stuff like midConfig) middleware by hand (not using mojito's sugar) then they still need to find out the list of middleware that mojito knows about, but this is just a RS.getResourceVersions() call so shouldn't be too hard.

lib/mojito.js
```diff
@@ -87,46 +91,33 @@ Creates a new instance of mojito and attach to `app`.
...
         appRoot = process.cwd(),
-        options = {},   //
+        appConfig,      // application.json
         store;          // reference to resource store
 
     // debug('applying mojito.defaultConfiguration()');
 
     if (!app.mojito) {
-        throw new Error('`app.mojito` was not initialized correctly.');
+        throw new Error('`app.mojito` was not initialized propertly.');
```
drewfish: "propertly" ?

lib/app/addons/ac/url.common.js
imalberto, line 39: sync'ed with @caridy offline. For now, the `find()` api will remain in the `ac.url` addon.

lib/middleware.js
caridy, line 115: I like `mojito store` as a config. we are already doing something similar with express-yui, by using `app.set('locator', locatorInstance);`

imalberto, line 115: I've updated `lib/mojito` so that now `app.js` can access the store if needed:

```
// app.js
...
store = app.get('mojito store'); // access to the resource store
versions = store.getResourceVersions(...);
...
```

We can update the built in middleware to use that new interface in a future PR.

tests/func/serveronly/serveronlytest_descriptor.json
imalberto: Fixed.

caridy:
+1

PR #1322, opened 2014-01-31
Remove expanded metadata and all related computations for yui

caridy:
After some investigation about expanded meta data, we figure that in newer versions of YUI it doesn't really provide any extra perf improvement. On top of that, it seems that no one was using this feature. With this PR we should see a little bit of an improvement in the booting time.

TODO:
- [x] unit tests
- [x] func tests
- [x] docs


caridy:
Tests to fix:

```
Failed mojito-addon-rs-yui-tests on nodejs
13 Passed, 1 Failed , 1 skipped 
 in yui rs addon tests
 yui meta for loader-app-base-en-US:Value should be undefined.
 Expected: undefined (undefined)
 Actual: [object Object] (object)
```

Func:

```
*************************************************************************************
                           List of failed descriptors                                  
Failed Descriptor Path : /home/travis/build/yahoo/mojito/tests/func/common/common_descriptor.json
Total Number of Failed Tests : 2
-----------------------------------------------------------------------
```

drewfish:
looks good

caridy:
There is something weird going on with the unit test that is failing, I see:

```
A.isUndefined(config.groups.app.modules['lang/PagedFlickr_de'], '`lang/PagedFlickr_de` should be undefined');
```

failing due to the fact that the module is in fact defined, but it seems that it should not be defined, although we can clearly see that module in fixture: `tests/fixtures/gsg5/mojits/PagedFlickr/PagedFlickr_de.js`. 

I don't remember anything about making lang bundles undefined modules.

/cc @drewfish

drewfish:
Should that particular lang bundle be undefined because it shouldn't apply to that particular environment/context ?

caridy:
yeah, that's absolutely right, there is something going on here, I will investigate more.

imalberto:
@caridy Once you have this merged, we should plan to release a `0.10.0a` alpha in NPM so that we can start baking the version in the other dependent packages.

caridy:
All systems are green! we can go ahead and merge this. /cc @imalberto

PR #1326, opened 2014-02-04
update tests to make time/date related tests run fine on both grid and local machines

lzhan:
Func tests failed when using firefox on grid because grid machine is using UTC time/date. This pr updates tests related time/date so that func tests can be run on both local machine and on Grid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
         "mojito-cli": "~0.1",
         "node-static": ">0.6.8",
         "wrench": "~1.3.9",
-        "yahoo-arrow": "0.0.89",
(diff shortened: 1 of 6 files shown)


drewfish:
```
0 Passed, 1 Failed , 0 skipped 
 in testCase_yui_3_10_3_2_1391551010905_38
 test mojitproxybroadcastdynamic:Values should be equal.
 Expected: bunny (string)
 Actual: null (object)
```

Fails on node 0.10, passes on node 0.8.

lzhan:
The test passed locally with node10. The failure is not related to the changes.

PR #1328, opened 2014-02-05
update Date related tests since grid uses different timezone

lzhan:


caridy:
still very fragile, and we will probably replace all this with the new Intl stuff that we are working on it. I will give you a +1 to get this out of the way until the refactor, but this test seems to be very fragile.

+1

PR #1329, opened 2014-02-07
More fixes for Search App

imalberto:
More fixes for Search App
- [x] Fix: load `routes.json` relative to `appDir` if not absolute path
- [x] New: added support for parametrized `call` by using `libmojito.dispatch()` - see tests for examples
- [x] Added / updated tests
- [x] Update docs

### DIFF
--- a/lib/dispatcher.js
+++ b/lib/dispatcher.js
@@ -74,11 +74,7 @@ module.exports = {
             }
 
             // do the `call` replacement
-            if (call.indexOf('{') > -1) {
-                recall = sub(call, req.params || {});
-            } else {
-                recall = call;
-            }
+            recall = sub(call, req.params || {});
             splits = recall.split('.');
             matchCall[1] = splits.pop();
             matchCall[0] = splits.join('.');
--- a/lib/dispatcher.js
+++ b/lib/dispatcher.js
@@ -49,16 +49,11 @@ module.exports = {
     action `help`.
     @param {Object} params optional routing parameters to attach to the
     request to mojito
+    @return {Function} middleware fn
     **/
     dispatch: function (call, params) {
         var my = module.exports,
-            matchCall = [],
-            routeParams = params || {},
-            splits;
-
-        splits = call.split('.');
-        matchCall[1] = splits.pop();
-        matchCall[0] = splits.join('.');
+            routeParams = params || {};
 
         if (params && typeof params === 'string') {
             routeParams = libqs.parse(params);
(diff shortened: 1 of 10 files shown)


lib/dispatcher.js
```diff
@@ -66,15 +61,28 @@ module.exports = {
...
+            // do the `call` replacement
+            if (call.indexOf('{') > -1) {
+                recall = sub(call, req.params || {});
+            } else {
+                recall = call;
+            }
+            splits = recall.split('.');
+            matchCall[1] = splits.pop();
+            matchCall[0] = splits.join('.');
+
```
imalberto, line 81: Not a big fan that the `call` is split on every request to be able to handle dynamic substitution. But with this change, one can do this:

```
// app.js
app.get('/:mojitId/:action', libmojito.dispatch('{mojitId}.{action}'));
```

lib/dispatcher.js
```diff
@@ -66,15 +61,28 @@ module.exports = {
...
+                recall,
+                splits;
 
             if (!mojito) {
                 return next(new Error('app.mojito does not exist.'));
             }
 
+            // do the `call` replacement
+            if (call.indexOf('{') > -1) {
+                recall = sub(call, req.params || {});
```
caridy: this is not an expensive operation, specially because it happens once per request. why the `if` then?

tests/func/applications/frameworkapp/configapp/app.js
```diff
@@ -22,14 +23,12 @@ app.use(libmojito.middleware());
...
 
-function rt(req, res, next) {
-    return libmojito.dispatch(req.params.type + '.' + req.params.action)(req, res, next);
-}
-app.get('/:type/:action', rt);
-app.post('/:type/:action', rt);
-app.put('/:type/:action', rt);
-app.head('/:type/:action', rt);
-app['delete']('/:type/:action', rt);
+mid = libmojito.dispatch('{type}.{action}');
```
caridy: probably we should call this `dispatcherMiddleware` instead of `mid` for better docs.

lib/dispatcher.js
caridy, line 81: that's ok, it is not an expensive and reiterative operation.

lib/dispatcher.js
imalberto: just checking if it is worth doing the `sub` on the value. otherwise `sub` will happen for every `call`, even if it is not parametrized. if that's not that expensive, I would like to remove it too.

caridy: this is not about the split, but the sub.

imalberto: yep, typo. i meant `sub`, not `split` which has to happen since it is dynamic in this case. updated my original comments.

tests/func/applications/frameworkapp/configapp/app.js
imalberto: fixed a bunch of examples too.

lib/dispatcher.js
caridy: yes, sub is very performant (http://yuilibrary.com/yui/docs/api/files/yui_js_yui-lang.js.html#l251), it is not going to take more than the `indexOf('{')` operation that you have anyways. lets change it.

imalberto: thanks. updated.

caridy:
+1 remember to update the docs

imalberto:
yep. working with @zhouyaoji on that.

PR #1336, opened 2014-02-14
improve serialization of meta modules

caridy:
using `express-state`'s serialization method to stringify meta modules that can have conditional and test functions.

this PR has not effect in mojito from the feature point of view.


imalberto:
+1

PR #1337, opened 2014-02-14
removing all hacks that were in place to prevent injections

caridy:
`express-state`'s serialization routine can take care of all that now


imalberto:
+1 
yeah for `express-state` !

PR #1338, opened 2014-02-18
Introduce a backward compatibility change for mix and match middleware with Mojito. Unit tests included.

imalberto:
`libmiddleware.middleware()` now can mix and match middleware that either exports a middleware function, or exports a function that returns a middleware function, based on the name convention of the middleware with the "mojito-" prefix.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -64,10 +64,21 @@ The most common usage is:
 
     app = express();
     libmojito.extend(app);
+    // Will load the middleware configuration from `application.json`
     app.use(libmojito.middleware());
 
-    // Or to use a specific Mojito middleware
+    // NOTE: Only Mojito built-in middleware are accessible via
+    //       `libmojito.middleware` object.
+    //       See `MOJITO_MIDDLEWARE` for the complete list.
+    //
+    // Or use the following format to specify explicit order:
+    //
     // app.use(libmojito.middleware['mojito-handler-static']);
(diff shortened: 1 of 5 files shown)


lib/middleware.js
```diff
@@ -67,7 +66,7 @@ The most common usage is:
     libmojito.extend(app);
     app.use(libmojito.middleware());
 
-    // TODO to use a specific Mojito middleware
+    // Or to use a specific Mojito middleware
     // app.use(libmojito.middleware['mojito-handler-static']);
```
caridy, line 76: this one is contradictory with the PR, if it starts with `mojito-`, why it is not a middleware factory/generator? it seems that there are more exceptions here: if it is a mojito internal middleware, then we don't use the generator, is that it?

caridy, line 76: I'm not saying it is incorrect, I'm just saying that this should be documented somehow.

imalberto, line 76: Correct, you can only use `app.use(libmojito.middleware['mojito-handler-static']);` for the mojito built-in middleware, since those lazy init at the first request. I think the source of confusion is because Mojito is trying to support the lazy init way, but at the same time provide some level of BC for legacy extensions.

 I will update the source file to doc this.

/cc @zhouyaoji to make sure we capture this in the wiki and clarify this.

caridy:
+1

PR #1346, opened 2014-02-21
Search next

imalberto:
Changes in this PR:
- Now that Mojito extends Express, moving the `express` `peerDependencies` to `devDependencies`.
- Pinning the version of `express` to 3.4.1 to avoid the issues documented [https://github.com/visionmedia/express/issues/1816](https://github.com/visionmedia/express/issues/1816).

This should get us to blue builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,6 @@
         "node": ">0.6",
         "npm": ">1.0"
     },
-    "peerDependencies": {
-    },
     "devDependencies": {
         "async": "*",
         "benchmark": "1.x",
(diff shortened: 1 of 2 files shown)


caridy:
+1

make sure we add  A LOT of docs around this.

imalberto:
/cc @zhouyaoji 
Joe, here is the gist of it:

Just to clarify, any express app that uses a later version , relies on the `bodyParser` middleware to handle POST body requests will see the same issue as outlined [here](https://github.com/visionmedia/express/issues/1816), with or without Mojito.

Mojito pins down the express version to this last know "good" version for running the tests suites, but it no way restricts the application to this express version. There are other alternatives to `bodyParser`.

PR #1347, opened 2014-02-24
add checking to make sure testing element is loaded

lzhan:
testlazyloadclient test failed when running with firefox.


drewfish:
+1

PR #1348, opened 2014-02-24
add more checking before testing page

lzhan:
Also trying to see if this can fix the issue with sd firefox.


tests/func/examples/quickstartguide/quickstartguide_descriptor.json
```diff
@@ -60,6 +75,12 @@
                            }
                         },
                         {
+                           "controller": "locator",
+                           "params": {
+                                  "value": "head title"
```
imalberto: Indentation does not seem correct.

imalberto:
+1

PR #1356, opened 2014-02-27
Allow the app starting with different port. This will fix the sd firefox issue.

lzhan:


tests/func/examples/quickstartguide/quickstartguide_descriptor.json
```diff
@@ -5,7 +5,7 @@
         "name" : "quickstartguide",
 
         "config" :{
-            "baseUrl" : "http://localhost:8666",
+            "baseUrl" : "http://partialsight.corp.yahoo.com:8666",
```
imalberto: @lzhan  This does not look right.

imalberto:
+1

PR #1360, opened 2014-03-06
Fix the test module for mocking mojit models.

imalberto:
With this fix, apps can test their models, which are now accessed via the `ac.models` addon, like this:

```
var ac = new Y.mojito.MockActionContext({
    addons: [ 'http' ],
    models: [ 'foo' ]
});

ac.models.get('foo').expect({
    method: 'getData',
    args: ...
});

controller.index(ac);

// somewhere in controller
ac.models.get('foo').getData(....);
```

### DIFF
--- a/lib/app/autoload/mojito-test.common.js
+++ b/lib/app/autoload/mojito-test.common.js
@@ -35,11 +35,14 @@ YUI.add('mojito-test', function(Y, NAME) {
         source[name] = new EasyMock();
     }
 
-
     function createMockModel(source, name) {
+        source._models.push(name);
         source.models[name] = new EasyMock();
-    }
 
+        source.models.get = function () {
+            return source.models[name];
+        };
+    }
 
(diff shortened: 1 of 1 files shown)


lib/app/autoload/mojito-test.common.js
```diff
@@ -35,11 +35,14 @@ YUI.add('mojito-test', function(Y, NAME) {
         source[name] = new EasyMock();
     }
 
-
     function createMockModel(source, name) {
+        source._models.push(name);
         source.models[name] = new EasyMock();
-    }
 
+        source.models.get = function () {
```
caridy, line 42: this method should probably be defined where `source.models` is defined. here it will be set everytime a new mock is created, not a big deal though.

caridy:
LGTM

imalberto:
Thanks @caridy

PR #1363, opened 2014-03-12
[code_exs] Updated the routing for the examples to use app.js, not routes.json.

zhouyaoji:
The only examples that still attach routes from `routes.json` were those that used the Url addon, which requires the `routes.json`.  Some errors were also fixed in the dashboard module apps, which are not tested. 

Note: the `app.js` for each example has the line `app.mojito.attachRoutes();` commented out with a note about what it does and that `routes.json` has been deprecated.

### DIFF
--- a/examples/developer-guide/binding_events/app.js
+++ b/examples/developer-guide/binding_events/app.js
@@ -21,11 +21,18 @@ libmojito.extend(app);
 
 app.use(customContextualizerMiddleware());
 app.use(libmojito.middleware());
+
+// To use the routing configured in `routes.json`, which
+// is deprecated.
+// The controller uses the Url addon, which requires the
+// `routes.json` file.
 app.mojito.attachRoutes();
 
 app.get('/status', function (req, res) {
     res.send('200 OK');
 });
+
+app.get('/', libmojito.dispatch('frame.index'));
 
 app.listen(app.get('port'), function () {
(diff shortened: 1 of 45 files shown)


examples/developer-guide/binding_events/app.js
```diff
@@ -21,12 +21,19 @@ libmojito.extend(app);
 
 app.use(customContextualizerMiddleware());
 app.use(libmojito.middleware());
+
+// To use the routing configured in `routes.json`, which
+// is deprecated.
+// The controller uses the Url addon, which requires the
```
caridy, line 27: what is this? can u elabore more?

caridy:
What about all the `routes.json` files in all those examples that have the line commented now? did u removed the actual files? or are u keeping them around for now?

caridy:
this LGTM btw.

zhouyaoji:
@caridy I didn't delete any of the `routes.json` files in the example code.  I thought it best to wait for the time when the file is no longer supported to delete it.

caridy:
I'm just concern that they try to copy the example, changing routes.json, and nothing happens. Probably we should add a note into those routes.json, at the top, saying something like, "...to use this routes do this...", or simply remove them to avoid confusion and left an example with one that works, probably combining routes.json + some other express routes.

PR #1364, opened 2014-03-14
[docs/code_ex] Updated steps and doc around `routes.json` in exs, removed one ex, updated another.

zhouyaoji:
The steps in the example now use `app.js` for routing and sections regarding routing have been updated to discuss how to define routing paths in `app.js`. In addition, the binding example was updated so that it doesn't use the Url addon, which requires `routes.json`, and the generating_urls example was deleted as was its test.


docs/dev_guide/code_exs/binding_events.rst
```diff
@@ -562,7 +558,7 @@ If the query string created by ``Y.QueryString.stringify`` is "page=2" ,
        for (var k in params) {
          mergedParams[k] = params[k];
        }
-       return actionContext.url.make('frame', 'index', Y.QueryString.stringify(mergedParams));
+       return "/?" + Y.QueryString.stringify(mergedParams));
```
caridy, line 561: why not using `url` addon anymore?

docs/dev_guide/code_exs/binding_events.rst
```diff
@@ -791,12 +766,11 @@ To set up and run ``binding_events``:
             for (var k in params) {
               mergedParams[k] = params[k];
             }
-            return actionContext.url.make('frame', 'index', Y.QueryString.stringify(mergedParams));
+            return "/?" + Y.QueryString.stringify(mergedParams);
```
caridy, line 769: me as above.

docs/dev_guide/code_exs/generating_urls.rst
```diff
@@ -1,262 +0,0 @@
-===============
-Generating URLs
```
caridy, line 2: ok, let's talk about this, because this features should be available, at least that's my understanding.

caridy:
Let's clarify about the `url` addon offline.

docs/dev_guide/code_exs/generating_urls.rst
zhouyaoji, line 2: It's available, but does not work unless you attach the routes from `routes.json`, at least that's what I discovered when testing the Url addon. I'm not sure how we want to handle this problem (feel free to confirm), whether it's through documentation, filing an issue, or removing the addon.

caridy, line 2: this is definitely an issue/bug, because it is suppose to use the express routes directly. /cc @imalberto

zhouyaoji, line 2: @caridy Can I merge this PR and file an issue or shall we wait?

caridy, line 2: we should wait until we clarify this.

zhouyaoji, line 2: Okay, no problem. FYI, this is blocking the hand-off and the GA of v0.9. Thanks.

imalberto:
/ping @caridy @zhouyaoji 
Hey guys,
I've updated the [wiki](https://github.com/yahoo/mojito/wiki/05.-Mojito-v0.9:-Configuring-Routing#mojito-v09) to include the additional `app.map()` that are required if `app.js` is registering routes outside of `routes.json`.

Remember that `routes.json` is sugar for now to actual `app.get()` and `app.map()` api calls. If needed, the application (preferably) can even provide their own wrapper around these calls for more complex configuration.

This is also mentioned in the [HISTORY.md](https://github.com/yahoo/mojito/blob/develop/HISTORY.md#features-1).

caridy:
@imalberto that doesn't clarify on the url addon, is that still supported? if yes, how? It seems like it fails for @zhouyaoji

imalberto:
@caridy 
The `url` add-on is still supported, and is used by both `mojito-shaker` and `mojito-debug` extensions. The `url` add-on relies on the routes being properly registered. I think looking at the examples, the routes are not registered via `routes.json` but in `app.js`. The matching `app.map()` calls should setup the right mapping for the `url` add-on to work.

zhouyaoji:
@imalberto @caridy Great, I was able to use the Url addon after registering the routes with `app.get`. I'll update the documentation and add the `generating_urls` example/test again.

zhouyaoji:
@imalberto @caridy I added and an updated `generating_urls` example/documentation, the tests, updated the file `mojito_routing.rst` to include how to make URLs w/ the Url addon and registering routes w/ `app.map`. See the last commit: 154edc2d84b39eb7500298d2539c075ea0f80a11

PR #1370, opened 2014-04-30
Lazy Mojits and Langs

aljimenez:
This pull requests introduces two `resourceStore` options, `lazyMojits` and `lazyLangs`. When set to true these options significantly reduce start up times for large applications, by lazy loading resources only when they are needed.


examples/quickstartguide/mojits/Read/lang/Read_en-US.js
```diff
@@ -9,10 +9,10 @@ Internationalize the printing texts generated by the progeam.
 Visit http://developer.yahoo.com/cocktails/mojito/docs/code_exs/i18n_apps.html for details.
 **/
 
-YUI.add("lang/read_en-US", function (Y, NAME) {
+YUI.add("lang/Read_en-US", function (Y, NAME) {
```
caridy, line 12: we should avoid using camelcase on module names, I recall some issues with CDN in the past.

caridy:
Btw, the last few PRs issued against `develop` have been merged really quick, without giving us at least 24 hrs to review them, it seems that we are not having a proper review process anymore, can we bring that back? /cc @aljimenez @jlecomte

jlecomte:
Absolutely! I think Albert was not sure whether you guys were still willing to review, especially since quite a few of you have signed off of the internal mojito-team ilist. Thanks for your help!

aljimenez:
@caridy, thanks for looking into the pull request; I thought you all had moved on to other projects, but I definitely appreciate feedback, so I'll make sure to give everyone a chance to comment.

examples/quickstartguide/mojits/Read/lang/Read_en-US.js
aljimenez, line 12: I renamed it to be consistent with the mojit name, which is capitalized. When using the lazyLangs option, langs are processed through their filenames, which saves a lot of time since we don't have to read and compile each lang file at startup; but it also means that conventions need to be strictly followed. I'll keep in mind the CDN issue.

PR #1371, opened 2014-04-30
Merge develop into master

aljimenez:


lib/app/addons/ac/intl.common.js
```diff
@@ -14,6 +14,15 @@
  */
 YUI.add('mojito-intl-addon', function(Y, NAME) {
 
+    // Overwriting setLang such that yuiActiveLang is not set,
+    // this prevents independent requests from interfering with each other.
+    // Previously if one request set a valid language, a subsequent request with
+    // an invalid language would end up using the previous language.
+    var setLang = function (module) {
+        //delete Y.Intl._mod(module).yuiActiveLang;
+        return Y.Intl.setLang.apply(Y.Intl, arguments);
```
caridy, line 23: this is very misleading and very error prompt. Mojito that this known limitation that whenever you plan to use a lang bundle, you need to call `Y.Intl.setLang()`, but doing what you're doing in line 55 is going to be problematic.

I have few recommendations for this:
- you should rely on `Y.Intl.lookupBestLang()` (http://yuilibrary.com/yui/docs/api/classes/Intl.html#method_lookupBestLang), and `Y.Intl.getAvailableLangs()` to avoid calling `setLang()` with an invalid lang.
- if the issue is to avoid having bundles that do not contain the right lang information, then you could do completion of the bundles and messages in bundles during build process to guarantee that all bundles and all language are available for use. this is the same we do in flickr and media, and it reduce the complexity of the app code and the framework code drastically in favor of a more normalized build process.

lib/app/addons/ac/intl.common.js
```diff
@@ -56,12 +71,18 @@ YUI.add('mojito-intl-addon', function(Y, NAME) {
...
-            Y.Intl.setLang('datatype-date-format', lang);
-            return Y.DataType.Date.format(date, {format: '%x'});
+            var lang = this.ac.instance.closestLang || 'en',
+                module = 'datatype-date-format',
+                activeLang = Y.Intl._mod(module).yuiActiveLang,
+                formattedDate;
+
+            // Deleting the active lang in order to be independent of
+            // previous languages set.
+            delete Y.Intl._mod(module).yuiActiveLang;
```
caridy, line 81: keep in mind that `delete` is problematic due to its nature. It will mess up with the V8 internal optimization, slowing down the access to any member of the object that suffers a delete. In most cases, setting the value to `undefined` or `null` is more than enough.

lib/app/addons/ac/intl.common.js
```diff
@@ -56,12 +71,18 @@ YUI.add('mojito-intl-addon', function(Y, NAME) {
...
+            Y.Intl.setLang(module, lang);
+            formattedDate = Y.DataType.Date.format(date, {format: '%x'});
+            Y.Intl._mod(module).yuiActiveLang = activeLang;
```
caridy, line 84: By doing this, you're effectively bypassing the internals of the YUI library to set up the proper active lang (including the event system associated with it).

lib/app/addons/ac/intl.common.js
```diff
@@ -56,12 +71,18 @@ YUI.add('mojito-intl-addon', function(Y, NAME) {
          * @return {string} formatted data for language.
          */
         formatDate: function(date) {
-            var lang = Y.mojito.util.findClosestLang(this.ac.context.lang, this.ac.instance.langs) ||
```
caridy, line 59: this is what I was saying before, we were looking for the closest match, to make sure that we don't use an invalid one, but that doesn't guarantee that a particular module has that lang, that's the missing link here I think, and it should be easy to resolve without having to get into all this complexity.

lib/app/addons/rs/selector.js
```diff
@@ -87,7 +88,13 @@ YUI.add('addon-rs-selector', function(Y, NAME) {
             this._appConfigYCB = this.host.config.getAppConfigYCB();
 
             cacheKey = JSON.stringify(ctx);
-            posl = this._poslCache[cacheKey];
+
+            if (update) {
+                delete this._poslCache[cacheKey];
```
caridy, line 93: avoid the use of `delete` in static objects.

lib/app/addons/ac/intl.common.js
aljimenez, line 23: Great, I will implement a less hacky solution soon. But this does fix a very important issue, before if a requests sets the language to, for example Chinese, and then another unrelated requests tries to set it to Spanish but there is no Spanish lang, then the last language would be used, so the user would see Chinese instead of the default lang. This was a big problem because that means that requests and mojits could interfere with each other.

lib/app/addons/ac/intl.common.js
aljimenez, line 81: Good tip :)

PR #1375, opened 2014-05-28
Read YUI modules only once

aljimenez:
Currently YUI modules are read twice; once to capture YUI metadata and then again when they are actually added to the runtime YUI. This is inefficient since each module requires two synchronous reads, and potentially problematic since part of the code is executed twice. This pull request manages to capture YUI metadata and adds YUI modules at the same time by hooking into the YUI.add method. This reduces startup time by up to 20%, and only executes modules once with the actual YUI object. Instead of using VM to execute modules, it does exactly what YUI does (wrapping code in module.exports); this actually allows us to capture the line number of syntax errors and runtime errors within the module (VM does not give line numbers).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,18 @@
+version 0.9.5
+=============
+
+Notes
+-----
+
+* Reduced start up time by up to 20% by capturing YUI module details and executing them in one step instead of separetely.
+* Now YUI modules are only executed once, with the real runtime YUI object scoped.
+* Syntax errors are now reported with line numbers after failing to compile YUI modules.
+
+Bug Fixes
+---------
+
+* Fixed issue where application start up would crash due to syntax error in a YUI module.
+
 version 0.9.4
 =============
 
--- a/lib/app/addons/rs/yui.js
+++ b/lib/app/addons/rs/yui.js
@@ -27,6 +27,7 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
         libmime = require('mime'),
         liburl  = require('url'),
         libutil = require('../../../util.js'),
+        Module = require('module'),
         serialize = require('express-state/lib/serialize'),
 
         WARN_SERVER_MODULES = /\b(dom-[\w\-]+|node-[\w\-]+|io-upload-iframe)/ig,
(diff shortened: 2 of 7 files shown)


yahoocla:
CLA is valid!

caridy:
The problem here is that you really don't know if the code is a YUI module or not, that's why it was executed in a different context to identify what type of script it is, then the proper execution in the main context.

aljimenez:
Thanks for looking into this. Can you give me an example where this would cause a problem. By the time we want to capture the YUI details, we assume the file is a YUI module, if it is not, then the user made a mistake. In that case the resource would not have any yui metadata and nothing would be added since YUI.add was never called. I see more potential problems the way things were done before since the code is executed twice. What if the user wants to call YUI.applyConfig before calling YUI.add, or what if the user does some other computation outside of YUI.add? Previously it would either crash or the computation would be done twice.

aljimenez:
I am ready to merge this to the develop branch. By the way, these series of pure start up optimizations since mojito@0.9.0 reduce the startup time of the dashboard example by over 60% and the Search application by over 40% (without any lazy loading).

caridy:
@aljimenez sorry for the delay, I was on the road, few notes:
- `YUI.applyConfig()` has nothing to do with `YUI.add()`, they are orthogonal
- if you have a vendor script in your mojit, something that is not a YUI module and tries to use `window` for example (since you use it from the client side only), you need to control the error (I think you have that one covered in this PR).
- this assumption is bold: "By the time we want to capture the YUI details, we assume the file is a YUI module".

The other thing is that you could simply check, if it is a yui module in a context, and then capturing the callback function and re-execute (call) the `YUI.add()` in the current context without having to read the file again, that is probably the ideal solution.

aljimenez:
@caridy thanks for the explanation.

When I mentioned YUI.applyConfig, what I meant was that the yui module may have other statements in addition to YUI.add. Before, YUI was mocked so if the module called YUI.applyConfig, it would crash since the mocked object didn't have that function. Now the user can put any code in the module, without worrying whether it is being executed by mojito to get YUI details or by YUI to add the module, the code just runs once.

Regarding capturing YUI details, the pull request does the same as before: if YUI.add is not called or there is some syntax error, then res.yui remains undefined. Only now when there are syntax/runtime errors, we get the callstack. Also previously if a module under yui_modules had a syntax error then the entire start up would crash, I fixed this issue.

PR #1377, opened 2014-05-29
Intl Addon Fix

aljimenez:
Removed hack that ensured mojits/requests translations didn't interfere. Now the closest available language is determined, resorting to a default lang or the root lang; if no language is set, then no translation is performed, this prevent the active language from being used, thereby eliminating mojits/requests interference.


yahoocla:
CLA is valid!

PR #1378, opened 2014-05-30
Mixing in static context into req's context

aljimenez:
This pull requests makes the static context available in the req's context, giving the req's context precedence as brought up in issue #1377.

### DIFF
--- a/lib/app/middleware/mojito-contextualizer.js
+++ b/lib/app/middleware/mojito-contextualizer.js
@@ -5,7 +5,7 @@
  */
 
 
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

caridy:
I thought this was done internally whenever you try to use that context to access data from the store.

drewfish:
It is, inside the store. However, `req.context` doesn't have it, so it's not possible to write other middleware that uses `req.context` to do interesting things.

caridy:
fair enough.

PR #1380, opened 2014-06-13
Removing connect's deprecation warnings

aljimenez:
Connect's multipart has been deprecated (see [connect multipart](http://www.senchalabs.org/connect/multipart.html)), and [express.bodyParser](https://github.com/senchalabs/connect/blob/2.x/lib/middleware/bodyParser.js#L54) is still making use of it. Since mojito's [parser-body middleware](https://github.com/yahoo/mojito/blob/develop/lib/app/middleware/mojito-parser-body.js) is using the express bodyParser there are connect warning during applications startup:

```
connect: multipart: use parser (multiparty, busboy, formidable) directly
connect: limit: Restrict request size at location of read
```

This pull request simply takes the original express.bodyParser [implementation](https://github.com/senchalabs/connect/blob/2.x/lib/middleware/bodyParser.js#L54), without the multipart support. This means that users would have to use their own middleware to process file uploads.

### DIFF
--- a/lib/app/middleware/mojito-parser-body.js
+++ b/lib/app/middleware/mojito-parser-body.js
@@ -18,3 +18,27 @@ Export a function which can create a body parser.
 @return {Object} The newly constructed body parser.
 **/
 module.exports = require('express').bodyParser;
+
+var express = require('express'),
+    json = express.json,
+    urlencoded = express.urlencoded;
+
+module.exports = function bodyParser(options) {
+    var _json = json(options),
+        _urlencoded = urlencoded(options);
+
+
+    return function (req, res, next) {
+        _json(req, res, function (err) {
+            if (err) {
+                return next(err);
+            }
+            _urlencoded(req, res, function (err) {
+                if (err) {
+                    return next(err);
+                }
+                next();
+            });
+        });
+    };
+};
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

lib/app/middleware/mojito-parser-body.js
```diff
@@ -18,3 +18,27 @@ Export a function which can create a body parser.
...
+            }
+            _urlencoded(req, res, function (err) {
+                if (err) {
+                    return next(err);
+                }
+                next();
+            });
+        });
+    };
+};
```
caridy, line 43: breakline

aljimenez, line 43: Since file uploads are no longer handled by this middleware, this change can be considered non-backward compatible, and I assume we should increment the mojito version to 0.10.0

caridy, line 43: yeah, that makes sense.

aljimenez, line 43: I've been thinking about this pull request, and I now think that I should revert the merge and instead start making these sorts of changes once Mojito starts supporting express 4. It seems like with express 4 we can no longer use its middlewares directly for [mojito-parser-body](https://github.com/yahoo/mojito/blob/develop/lib/app/middleware/mojito-parser-body.js) and [mojito-parser-cookies](https://github.com/yahoo/mojito/blob/develop/lib/app/middleware/mojito-parser-cookies.js) (see [migrating from express 3 to 4](https://github.com/visionmedia/express/wiki/Migrating-from-3.x-to-4.x#overview)). So I rather make all the express 4 changes for one release, and keep the next release backward compatible.

PR #1382, opened 2014-06-13
Clearer error messages

aljimenez:
There are many different kinds of error conditions due to an invalid configured mojit; however mojito usually prints the undescriptive error below:

```
error: (outputhandler.server): { [Error: Cannot expand instance [searchhtmlframe], or instance.controller is undefined] code: 500 }
```

This makes it difficult to trace the actual problem. This commit addresses this problem by giving more useful error messages as shown below:
#### Mojit is missing:

```
error: (mojito-dispatcher): 
Error expanding instance for "searchhtmlframe": 
Error: Cannot find the "ShakerPipelineFrame" mojit. Make sure "ShakerPipelineFrame" exists in the application.
    at ResourceStore.Y.extend.resolveVersion ...
```
#### Mojit controller is missing

This means either the controller doesn't exist, it never called YUI.add, or it had a syntax error.

```
error: (mojito-dispatcher): Error expanding instance for "searchhtmlframe": no valid controller found in the "ShakerPipelineFrame" mojit
```
#### Mojit specs is missing a type:

```
error: (mojito-dispatcher): 
Error expanding instance for "searchhtmlframe": 
Error: Instance is missing a mojit type.
    at ResourceStore.Y.extend.expandInstanceForEnv ...
```
#### Controller did not set Y.mojito.controller[NAME] properly

**Before:**

```
error: (outputhandler.server): { [Error: Invalid controller name [ShakerPipelineFrameMojit] for mojit [ShakerPipelineFrame].] code: 500 }
```

**After:**

```
error: (mojito-dispatcher): Invalid controller, named "ShakerPipelineFrameMojit", in the "ShakerPipelineFrame" mojit. The controller does not define its actions under Y.namespace('mojito.controllers')['ShakerPipelineFrameMojit']
```

### DIFF
--- a/lib/app/addons/rs/yui.js
+++ b/lib/app/addons/rs/yui.js
@@ -600,6 +600,11 @@ YUI.add('addon-rs-yui', function(Y, NAME) {
             var store = this.get('host');
 
             this._captureYUIModuleDetails(res);
+
+            if (!res.yui) {
+                // Do not add YUI resources that failed while capturing details.
+                return new Y.Do.Halt();
+            }
         },
 
 
--- a/lib/app/autoload/action-context.common.js
+++ b/lib/app/autoload/action-context.common.js
@@ -322,7 +322,7 @@ YUI.add('mojito-action-context', function(Y, NAME) {
                 actionFunction = '__call';
             } else {
                 // If there is still no joy then die
-                error = new Error("No method '" + command.action + "' on controller type '" + command.instance.type + "'");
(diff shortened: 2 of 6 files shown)


yahoocla:
CLA is valid!

caridy:
LGTM, there is a failing test though.

aljimenez:
Great. I'll fix the tests, they are just expecting the previous error messages.

PR #1383, opened 2014-06-19
Catching and printing exceptions in binders' init/bind methods

aljimenez:
It is important to catch these exceptions, otherwise if a binder has an exception then the mojito-client will stop its execution. This results in other binders not getting called and also leaves the mojito-client in an erroneous state, which can result in the mojito-client always remaining paused, and calls to mp.invoke would never work.

### DIFF
--- a/lib/app/autoload/mojito-client.client.js
+++ b/lib/app/autoload/mojito-client.client.js
@@ -140,7 +140,7 @@ YUI.add('mojito-client', function(Y, NAME) {
             try {
                 binder.bind(node, element);
             } catch (e) {
-                Y.log(e.stack || e.message, 'error');
+                Y.log(e.stack || e.message, 'error', NAME);
             }
         }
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

lib/app/autoload/mojito-client.client.js
```diff
@@ -641,11 +645,14 @@ YUI.add('mojito-client', function(Y, NAME) {
                     });
 
                     if (Y.Lang.isFunction(binder.init)) {
-                        binder.init(mojitProxy);
+                        try {
+                            binder.init(mojitProxy);
+                        } catch (e) {
+                            Y.log(e.stack || e.message, 'error');
```
caridy: 3rd argument is important to filter out errors per component/library, either you use `mojito` or `mojito-client`.

caridy:
LGTM

PR #1385, opened 2014-06-27
Allow mojits to extend YUI modules in other mojits

aljimenez:
This pull request allows mojits to easily extend YUI modules of other mojits. For example:

mojits/Vehicle/controller.server.js

``` js
YUI.add('VehicleController', function (Y, NAME) {
    var Vehicle = function () {
        this.type = 'Vehicle';
        this.transmission = 'automatic';
    };

    Y.namespace('mojito.controllers')[NAME] = Vehicle;
    Y.extend(Vehicle, Y.Base, {
        index: function () {
            ac.done(ac.params.body('brand') + ' ' + this.getType());
        },
        getType: function () {
            return this.type;
        }
    });
}, '0.0.1', {
    requires: {
        'oop',
        'mojito-params-addon'
    }
});
```

mojits/Motorcycle/controller.server.js

``` js
YUI.add('MotorcycleController', function (Y, NAME) {
    var Vehicle = Y.mojito.controllers.VehicleController,
        Motorcycle = function () {
            Motorcycle.superclass.constructor.call(this);
            this.type = 'Motorcycle';
            this.transmission = 'manual';
        };

    Y.namespace('mojito.controllers')[NAME] = Motorcycle;
    Y.mojito.util.extend(Motorcycle, Vehicle);
}, '0.0.1', {
    requires: {
        'mojito-util',
        'VehicleController'
    }
});
```

In this example, the Motorcycle mojit extends the Vehicle mojit, which extends the Y.Base class (extending Y.Base is not a requirement, the example just illustrates multiple level of extensions). Notice that the controllers can be defined as a function; this allows mojits to extend using Y.extend, which ensures that the prototype chain is maintained regardless of how many levels of extensions. Also notice that the Motorcycle controller does not have to re-specify the 'mojito-params-addon' requirement; it just has to specify the Vehicle controller requirement, and its addon will automatically include any addon required by Vehicle controller.

To ensure that mojit dependencies work with resourceStore > lazyMojits on, mojits depending on other mojits should specify their dependencies in defaults.json:

mojits/Motorcycle/defaults.json

``` js
[{
    "settings": [ "master" ],
    "dependencies": [ "Vehicle" ]
}]
```

### DIFF
--- a/lib/app/addons/rs/dispatch-helper.js
+++ b/lib/app/addons/rs/dispatch-helper.js
@@ -37,7 +37,11 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
 
 
         initializer: function (config) {
-            this.modules = {}; // Controllers and ac addons.
+            this.modules = {
+                client: {},
+                server: {},
+                common: {}
+            }; // Controllers and ac addons.
             this.beforeHostMethod('addResourceVersion', this.addResourceVersion, this);
             this.onHostEvent('resolveMojitDetails', this.onResolveMojitDetails, this);
         },
@@ -57,10 +61,11 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
                 // should be derived from its addons requirements, recursively including the addons of
                 // the controllers and addons it requires.
 
+                var affinity = res.affinity.toString();
                 // Do not overwrite a previous module with the same name; this ensures that
                 // if there are duplicate modules, only the shallowest is considered.
-                if (!this.modules[res.yui.name]) {
-                    this.modules[res.yui.name] = res;
+                if (!this.modules[affinity][res.yui.name]) {
+                    this.modules[affinity][res.yui.name] = res;
                 }
             }
         },
@@ -96,22 +101,25 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
             details.acAddons = [];
             (function getRequiredAddons(requires, seen) {
                 var r,
-                    module;
+                    module,
+                    moduleName;
                 for (r = 0; r < requires.length; r++) {
-                    module = requires[r];
-                    if (seen[module] || !self.modules[module] ||
-                            (self.modules[module].affinity.affinity !== env &&
-                                self.modules[module].affinity.affinity !== 'common')) {
+                    moduleName = requires[r];
+
+                    if (seen[moduleName]) {
                         continue;
                     }
-                    seen[module] = true;
+                    seen[moduleName] = true;
 
-                    getRequiredAddons(self.modules[module].yui.meta.requires, seen);
-                    if (acAddonNames[module]) {
-                        details.acAddons.push(acAddonNames[module]);
+                    module = self.modules[env][moduleName] || self.modules.common[moduleName];
+                    if (module) {
+                        getRequiredAddons(module.yui.meta.requires, seen);
+                        if (acAddonNames[moduleName]) {
+                            details.acAddons.push(acAddonNames[moduleName]);
+                        }
                     }
                 }
-            }(self.modules[details.controller].yui.meta.requires, {}));
+            }([details.controller], {}));
         }
 
     });
--- a/lib/app/addons/rs/dispatch-helper.js
+++ b/lib/app/addons/rs/dispatch-helper.js
@@ -56,7 +56,12 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
                 // We only care about controllers and AC addons since a mojit's controller's addons
                 // should be derived from its addons requirements, recursively including the addons of
                 // the controllers and addons it requires.
-                this.modules[res.yui.name] = res;
+
+                // Do not overwrite a previous module with the same name; this ensures that
+                // if there are duplicate modules, only the shallowest is considered.
+                if (!this.modules[res.yui.name]) {
+                    this.modules[res.yui.name] = res;
+                }
             }
         },
 
--- a/lib/app/addons/rs/dispatch-helper.js
+++ b/lib/app/addons/rs/dispatch-helper.js
@@ -4,6 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
+/*jslint plusplus: true */
 /*global YUI*/
 
 
@@ -43,13 +44,18 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
 
         /**
          * Using AOP, this is called before the ResourceStore's version, in order
-         * to keep track of all the controllers and ac addons.
+         * to keep track of all the controllers and AC addons. These modules are then
+         * used by onResolveMojitDetails in order to precompute list of AC addons for
+         * a mojit's controller.
          * @method addResourceVersion
          * @param {object} res resource version metadata
          * @return {nothing}
          */
         addResourceVersion: function (res) {
-            if (res.type === 'controller' || res.type === 'addon' && res.subtype === 'ac') {
+            if (res.type === 'controller' || (res.type === 'addon' && res.subtype === 'ac')) {
+                // We only care about controllers and AC addons since a mojit's controller's addons
+                // should be derived from its addons requirements, recursively including the addons of
+                // the controllers and addons it requires.
                 this.modules[res.yui.name] = res;
             }
         },
@@ -89,7 +95,8 @@ YUI.add('addon-rs-dispatch-helper', function (Y, NAME) {
                 for (r = 0; r < requires.length; r++) {
                     module = requires[r];
                     if (seen[module] || !self.modules[module] ||
-                        (self.modules[module].affinity.affinity !== env && self.modules[module].affinity.affinity !== 'common')) {
+                            (self.modules[module].affinity.affinity !== env &&
+                                self.modules[module].affinity.affinity !== 'common')) {
                         continue;
                     }
                     seen[module] = true;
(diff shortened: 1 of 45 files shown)


yahoocla:
CLA is valid!

gomezd:
IMHO we need to do extensive performance testing to determine any possible impact of this.
And i'm not a big fan of having to declare the dependencies in another file (defaults.json), seems to me confusing and error prone.

aljimenez:
This pull request does not change the execution flow for any existing application. It only adds a new feature, which is to allow a mojit controller to be specified as a function with a prototype. This has no performance impact, since the execution flow remains the same. The only thing that changes is that instead of mojito using Y.mojito.util.heir to create the controller object (see [heir](https://github.com/yahoo/mojito/blob/develop/lib/app/autoload/util.common.js#L86)), the controller object is created directly from the controller function. This means that we don't have to create a new function every time, which obviously wouldn't degrade performance or memory usage. Specifying the dependency somehow is necessary only if using lazyMojits, there is no way around this, and shouldn't be a huge issue for users since extending the modules of another mojit should be rare. In fact without the pull request mojits extending another mojit's modules would have to specify the dependency through YUI().applyConfig (see [ShakerPipelineFrame](https://github.com/yahoo/mojito-pipeline/blob/master/mojits/ShakerPipelineFrame/controller.server.js#L12)), in case lazyMojits is on. Specifying the dependencies in default.json is probably the cleanest most intuitive way (and probably good practice anyways since in general mojits are self contained, so its a good idea to mention that it depends on another mojit in its configuration).

caridy:
This PR is misleading, it is not extending mojits, it is extending individual modules within a mojit, and that's supported today as far as I can tell, by relying on the YUI module definition, I wonder why are you trying to mix the concepts of a mojit and modules in this PR.

aljimenez:
Thats right, you can extend any YUI module of another mojit, including controllers, binders, models... This pull request enhances 3 things:
1. When you require a controller, you also get the addons of that controller. Before if you required an external controller you would also need to copy the list of addons required by that controller. (See this [example](https://git.corp.yahoo.com/searchfe/search-apps-news/blob/master/mojits/rss/controller.server.js#L121))
2. The pull request allows you to use the Y.extend directly on the controller, since now controllers can be defined as functions. This is important in order to maintain the prototype chain regardless of how deep the level of extending. For example mojit c's controller can extend mojit's b, which extends mojit's a's controller. Previously you would have to use Y.merge since Y.extend only works on functions not object literals. And if you used Y.merge you would lose the prototype chain and mojit c's controller would not properly inherit mojit's a controller methods.
3. When using lazyMojits, mojits are loaded as they are needed, however if a mojit depends on the YUI module of another mojit, then those YUI modules would not be available. This is currently solved by having a Y.applyConfig statement before doing YUI.add (see [ShakerPipelineFrame](https://github.com/yahoo/mojito-pipeline/blob/eaf414ee7d398eb5d4075c65a9a1a315fc9d76eb/mojits/ShakerPipelineFrame/controller.server.js#L12)). Defining the mojit's external dependencies in defaults.json, is a cleaner solution since it doesn't require the applyConfig statement and is also an easy way to see if this mojit depends on another mojit.

aljimenez:
In the latest commit, [Y.mojito.util.heir](https://github.com/yahoo/mojito/commit/65f285510ffde9adc64b4947ae614c91c9656788#diff-bf08fb07212a332fa82c578236db9992R102) accepts functions in addition to object literals. This allows controllers/binders/models to be defined as functions, which can be useful when extending other YUI modules. The commit also adds [Y.mojito.util.extend](https://github.com/yahoo/mojito/commit/65f285510ffde9adc64b4947ae614c91c9656788#diff-bf08fb07212a332fa82c578236db9992R85), a wrapper around Y.extend, in order to accept either functions or object literals as arguments.

lib/app/addons/rs/dispatch-helper.js
```diff
@@ -0,0 +1,120 @@
...
+         * @method addResourceVersion
+         * @param {object} res resource version metadata
+         * @return {nothing}
+         */
+        addResourceVersion: function (res) {
+            if (res.type === 'controller' || (res.type === 'addon' && res.subtype === 'ac')) {
+                // We only care about controllers and AC addons since a mojit's controller's addons
+                // should be derived from its addons requirements, recursively including the addons of
+                // the controllers and addons it requires.
+                this.modules[res.yui.name] = res;
```
aljimenez: Here we keep track of all the ac addons and controllers, which we use below in order to populate a details.acAddons which includes all ac addons specified in its requirement tree.

lib/app/autoload/store.server.js
```diff
@@ -667,11 +672,31 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+                // of the dependencies have been added so that this mojit's YUI modules can
+                // refer to any YUI module belonging to its dependencies.
+                // Note that circular dependencies are allowed and will not result in an infinite loop.
+                if (this.lazyMojits && defaults && Y.Lang.isArray(defaults.dependencies)) {
+                    for (i = 0; i < defaults.dependencies.length; i++) {
+                        dependency = defaults.dependencies[i];
+                        if (this._unloadedMojits[dependency]) {
+                            // Only get mojit details if the mojit hasn't been loaded;
+                            // this prevents an infinite loop if there are circular dependencies.
+                            this.getMojitTypeDetails(env, ctx, dependency);
```
aljimenez, line 691: Here we recursively get mojit details of the dependencies, which ensures that we have all the resources available for this mojit before we call resolveVersion. There we determine details.acAddons, so its important to know all the ac addons that might have been specified by the mojit's dependencies.

lib/app/autoload/store.server.js
```diff
@@ -2239,6 +2250,18 @@ YUI.add('mojito-resource-store', function(Y, NAME) {
...
+
+                // Determine the mojit's dependencies.
+                dependencies = details.defaults && details.defaults.dependencies;
+                dependencies = Y.Lang.isArray(dependencies) ? dependencies : [];
+                // Populate details.modules with all the modules of this mojit's dependencies.
+                for (i = 0; i < dependencies.length; i++) {
+                    dependency = dependencies[i];
+                    dependencyDetails = this.getMojitTypeDetails(env, ctx, dependency);
+                    if (dependencyDetails) {
+                        Y.mix(details.modules, dependencyDetails.modules);
```
aljimenez: Here we populate details.modules with all the YUI modules belonging to this module including the modules of its dependencies. This code is only executed when the mojit is required on the client side, in which case if lazyMojits/lazyLangs is on, the current loader may not have all the YUI modules listed so we populate details.modules so that the client side knows the location any YUI module this mojit might refer to.

lib/app/autoload/util.common.js
```diff
@@ -82,8 +82,27 @@ YUI.add('mojito-util', function(Y, NAME) {
...
+                r.prototype = o;
+            }
+            if (s && typeof s === 'object') {
+                o = s;
+                s = function () {};
+                s.prototype = o;
+            }
+
+            return Y.extend(r, s, px, sx);
+        },
```
aljimenez, line 99: The new Y.mojito.util.extend method is a wrapper around Y.extend, which allows the r/s arguments to be either functions or the usual object literal used for controllers/mojits/binders.

lib/app/autoload/util.common.js
```diff
@@ -82,8 +82,27 @@ YUI.add('mojito-util', function(Y, NAME) {
...
+
 
         heir: function(o) {
+            if (typeof o === 'function') {
+                return new o();
+            }
```
aljimenez, line 105: This allows mojits to be specified as functions. Often while extending a module it is useful to have a constructor for hooking into the methods (see [this](https://github.com/yahoo/mojito/blob/extend-mojits/tests/func/applications/frameworkapp/common/mojits/GreenChild/binders/index.js#L22) example). This is also necessary if Y.mojito.util.extend is used because it returns a function.

caridy:
need some time to go over this tomorrow.

aljimenez:
Thanks Caridy, let me know if I should clarify anything.

lib/app/mojits/TunnelProxy/controller.server.js
```diff
@@ -38,6 +38,26 @@ YUI.add('TunnelProxy', function(Y, NAME) {
...
+                    // Add the YUI loader meta to the assets.
+                    meta.assets.top = meta.assets.top || {};
+                    meta.assets.top.js = meta.assets.top.js || [];
+                    // Append the current time to ensure that the loader meta is updated on the client side.
+                    meta.assets.top.js.push(yuiConfig.seed[yuiConfig.seed.length - 1] + '?' + new Date().getTime());
+                    meta.assets.top.blob = meta.assets.top.blob || [];
+                    // Use the loader-app module, to ensure the new loader meta gets applied.
+                    meta.assets.top.blob.push('<script>YUI().use("loader-app");</script>');
+                }
+
```
aljimenez, line 60: When there is a tunnel request and mojits/langs are loaded lazily, then the loader meta data on the client becomes out of date and so here we update the loader on the server and add the loader to meta.assets such that the client loads the new loader meta.

aljimenez:
@caridy can you review this when you get a chance; we have some application code that depend on this pull request. Thanks.

caridy:
@aljimenez let's have a chat tomorrow about this PR, I have few specific questions and some suggestions, but this is a very complex PR, hard to make sense of all pieces.

aljimenez:
@caridy sounds good. The pull request did get a bit complicated due to some refactoring and some other fixes that I slipped in, somewhat unrelated to the original pull request. I'll make sure to separate the distinct features in different branches in the future, thanks.

caridy:
ok, I will not hold back on this: +0.9

aljimenez:
Thanks @caridy I will merge this soon, and I'll keep in mind your suggestions.

PR #1386, opened 2014-07-02
Route annotations in routes.json and hidding routes from the client

aljimenez:
This pull request allows users to specify annotations in routes.json. By default routes in routes.json contain an annotation.client set to true. Setting a route's annotations.client to false prevents the route from being exposed to the client.

### DIFF
--- a/lib/app/addons/ac/deploy.server.js
+++ b/lib/app/addons/ac/deploy.server.js
@@ -69,7 +69,7 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
                 initializer, // script for YUI initialization
                 pathToRoot,
                 pageData = this.adapter && this.adapter.page && this.adapter.page.data,
-                pageRoutes = this.adapter && this.adapter.page && this.adapter.page.routes,
+                pageRoutes = this.adapter && this.adapter.page && this.adapter.page.clientRoutes,
                 updateLoaderCount = store._updateLoaderCount,
                 loaderRess,
                 lang = this.ac.context.lang;
--- a/lib/dispatcher.js
+++ b/lib/dispatcher.js
@@ -107,7 +107,7 @@ module.exports = {
     The `command` should be attached to the request.
 
     @protected
-    @method handleRequest 
+    @method handleRequest
     @param {http.ServerRequest} req
         @param {Object} req.command the mojito command to be executed
     @param {http.ServerResponse} res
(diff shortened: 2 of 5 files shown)


yahoocla:
CLA is valid!

lib/app/addons/ac/deploy.server.js
```diff
@@ -69,7 +69,7 @@ YUI.add('mojito-deploy-addon', function (Y, NAME) {
                 initializer, // script for YUI initialization
                 pathToRoot,
                 pageData = this.adapter && this.adapter.page && this.adapter.page.data,
-                pageRoutes = this.adapter && this.adapter.page && this.adapter.page.routes,
+                pageRoutes = this.adapter && this.adapter.page && this.adapter.page.clientRoutes,
```
aljimenez, line 72: We only expose client specific routes to the client.

lib/dispatcher.js
```diff
@@ -143,9 +143,13 @@ module.exports = {
         // - routes are not contextualized anymore
         if (!CACHE.routes) {
             CACHE.routes = app.getRouteMap();
+            CACHE.clientRoutes = app.getRouteMap({
+                client: true
+            });
```
aljimenez, line 148: Here we filter routes by those that have the client annotation set to true (which is the default).

lib/router.js
```diff
@@ -162,6 +162,9 @@ module.exports = {
         function registerRoutes(routes) {
             Object.keys(routes).forEach(function (name) {
                 var route = routes[name];
+                route.annotations = route.annotations || {};
+                route.annotations.client = route.annotations.client !== false;
+                app.annotate(route.path, route.annotations);
```
aljimenez, line 167: Here we set a default annotations.client to true and use app.annotate to pass any all annotations, including any specified by the user in routes.json.

aljimenez:
I will merge this pull request soon, if there aren't any concerns.

lib/router.js
```diff
@@ -162,6 +162,9 @@ module.exports = {
         function registerRoutes(routes) {
             Object.keys(routes).forEach(function (name) {
                 var route = routes[name];
+                route.annotations = route.annotations || {};
+                route.annotations.client = route.annotations.client !== false;
```
caridy, line 166: this line in confusing. is this a typecast to boolean with default to true?

caridy:
LGTM

lib/router.js
aljimenez, line 166: It is a bit confusing, I'll add a comment. It just means that the client value is always considered true unless its set to specifically false.

PR #1389, opened 2014-12-05
Version 0.9.8

aljimenez:
This pull request updates YUI to the latest version (3.18.x). It also ensures that action contexts objects include the addons of any yui module (not just controllers) required directly/indirectly by a controller.


yahoocla:
CLA is valid!

PR #1391, opened 2015-03-12
Fix configuration loading on Node v0.12.x and io.js.

reid:
Use `err.code` for Node forward compatibility.

The `errno` property of errors varies between Node v0.10.x and v0.12.x.
Use the property `code` instead, which is the same across both versions
and io.js.


yahoocla:
CLA is valid!

